Issue #3, opened 2018-10-29
Create Rails generators


hopsoft:
# Requirements

The generator should support 2 primary objectives. 

1. Setup and prepare the application to use StimulusReflex
1. Support generating new Reflexes

## Setup

```ruby
bundle exec rails stimulus_reflex:install
```

This will check for the existence of Stimulus in the project and will also create the following files.

- app/javascript/controllers/application_controller.js - something similar to https://github.com/hopsoft/stimulus_reflex_todomvc/blob/master/app/javascript/controllers/application_controller.js
- app/javascript/controllers/example_controller.js
- app/reflexes/application_reflex.rb
- app/javascript/controllers/example_reflex.rb

These files will include basic boilerplate and establish some patterns and practices that developers can follow to build their own StimulusReflex features.

## Generators

_NOTE: Generator args should behave similar to other Rails generators._

```ruby
bundle exec rails generate stimulus_reflex demo
```

This above command will generate 2 files.

- app/javascript/controllers/demo_controller.js
- app/reflexes/demo_reflex.rb

The files will include basic boilerplate to help developers get started.

### Optional Generators

```ruby
bundle exec rails generate stimulus_reflex:controller demo
```

This above command will generate 1 file.

- app/javascript/controllers/demo_controller.js

```ruby
bundle exec rails generate stimulus_reflex:reflex demo
```

This above command will generate 1 file.

- app/reflexes/demo_reflex.rb

PR #5, opened 2018-10-31
Add generator for server-side Stimulus controller

brascoder:


hopsoft:
This is looking great. I have some additional thoughts now that I see how this can shape up.

The generator should help teach what the best practices are and how the JavaScript and Ruby files are related. I think running a generator command like this

```sh
rails generate stimulus:reflex UserSelect
```

Should create the following files (or something like this)

```javascript
// app/javascript/controllers/user_select_controller.js
import { Controller } from 'stimulus';

  connect() {
    StimulusReflex.register(this);
  }

  doSomething(event) {
    send('UserSelect#do_something', true, "N number of args, ...");
  }
}
```

```ruby
# app/stimulus_controllers/user_select_stimulus_controller.rb
UserSelectStimulusController < StimulusReflex::Controller
  # set this to what the ActionCable connection is identified_by
  #delegate :current_user, to: :channel

  def do_something(accpets_args, message)
    # accepts_args == true
    # message == "N·number·of·args,·..."

    # do work, save to database, etc...
    # the current page will re-render
  end
end
```

---

Side note: We may need an additional generator to ensure that the following lines are added to `cable.js`

```javascript
// app/assets/javascripts/cable.js
//= require cable_ready
//= require stimulus_reflex
```

hopsoft:
One additional requirement: The generator should support help docs. i.e. `rails generate stimulus:reflex --help`

andrewmcodes:
@hopsoft happy to update this PR or add a new one. What are you wanting to see here?

hopsoft:
Not sure yet given the features that are about to land. At the very least the ruby reflex should be generated with perhaps an optional stimulus controller... or maybe we omit generating any javascript at all since it wont be strictly required.

andrewmcodes:
I’ll wait until some of the churn settles down. 👍

hopsoft:
NOTE: I updated requirements on #3

hopsoft:
Resolved by #67

PR #9, opened 2019-07-18
Ruby function splat args return arity of -1

pjforde1978:
I wasn't sure why I could specify an argument in my Reflex function, but it turns out a Ruby function with *args has arity -1, which means that you're actually sending a splat in but the end user developer must specify a specific number of arguments to receive.

This is not my primary area of expertise, but I am working on the assumption that this was an oversight. If there's a good reason to not encourage the use of splat arguments, I'd love to learn.


hopsoft:
This is a solid improvement. Thanks.

PR #10, opened 2019-07-22
Add support for room param to facilitate per-session broadcasts

pjforde1978:
If there are no connection identifiers in use, broadcasts will be sent to all connected clients. Since the params hash returns empty strings as default values, there's no harm in adding a check for the room parameter, which is a common convention in Action Cable development.

In my proposed changes to stimulus_reflex.js, I check the document head for a meta tag called session, which I pack with the session.id in the application layout.

### DIFF
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -12,7 +12,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     ids = connection.identifiers.map { |identifier|
       send(identifier).try(:id)
     }
-    "#{channel_name}#{ids.join ":"}"
+    "#{channel_name}#{ids.join ":"}#{params[:room]}"
   end
 
   def subscribed


hopsoft:
I very much agree that this needs to support `rooms`. The challenge is that StimulusReflex controls both the client and server. We need a way to set the `room` when the client side Stimulus controller is registered. I'll be experimenting with this later today.

pjforde1978:
I’m a fan of how I did it in my copy of the library in my project. Can you try doing a fresh git clone and don’t change out the libraries?

hopsoft:
I've started this work on the client. https://github.com/hopsoft/stimulus_reflex_client/pull/1

I'll also need to make some changes in [CableReady](https://github.com/hopsoft/cable_ready) to support this.

hopsoft:
I've moved this effort to #11 

Note that I'm pulling the JavaScript back into this repo.

hopsoft:
This functionality is now merged. Thanks for recommending it and pushing for it.

PR #12, opened 2019-08-14
Passive reflex mode with multiple argument support

pjforde1978:
This is a 2nd pass at adding passive reflex support. I took pains to follow the general structure of what you had, and this functionality is now activated when the register() call is made from the Stimulus controller. Arguments are any data attribute that starts with reflex-arg and they are processed in the order they appear.

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -1,17 +1,16 @@
-import ActionCable from 'actioncable';
-import CableReady from 'cable_ready';
+import ActionCable from 'actioncable'
+import CableReady from 'cable_ready'
 
-const app = window.App || {};
-app.StimulusReflex = app.StimulusReflex || {};
+const app = window.App || {}
+app.StimulusReflex = app.StimulusReflex || {}
 app.StimulusReflex.consumer =
-  app.StimulusReflex.consumer || ActionCable.createConsumer();
-app.StimulusReflex.subscriptions = app.StimulusReflex.subscriptions || {};
+  app.StimulusReflex.consumer || ActionCable.createConsumer()
+app.StimulusReflex.subscriptions = app.StimulusReflex.subscriptions || {}
 
 const createSubscription = controller => {
-  const { channel, room } = controller.StimulusReflex;
-  const id = `${channel}${room}`;
-  const renderDelay = controller.StimulusReflex.renderDelay || 25;
-
+  const { channel, room } = controller.StimulusReflex
+  const id = `${channel}${room}`
+  const renderDelay = controller.StimulusReflex.renderDelay || 25
   const subscription =
     app.StimulusReflex.subscriptions[id] ||
     app.StimulusReflex.consumer.subscriptions.create(
@@ -19,30 +18,57 @@ const createSubscription = controller => {
       {
         received: data => {
           if (data.cableReady) {
-            clearTimeout(controller.StimulusReflex.timeout);
+            clearTimeout(controller.StimulusReflex.timeout)
             controller.StimulusReflex.timeout = setTimeout(() => {
-              CableReady.perform(data.operations);
-            }, renderDelay);
+              CableReady.perform(data.operations)
+            }, renderDelay)
           }
-        },
+        }
       }
-    );
+    )
 
-  app.StimulusReflex.subscriptions[id] = subscription;
-  controller.StimulusReflex.subscription = subscription;
-};
+  app.StimulusReflex.subscriptions[id] = subscription
+  controller.StimulusReflex.subscription = subscription
+}
 
 const extend = controller => {
   Object.assign(controller, {
-    stimulate() {
-      clearTimeout(controller.StimulusReflex.timeout);
-      const url = location.href;
-      let args = Array.prototype.slice.call(arguments);
-      let target = args.shift();
-      controller.StimulusReflex.subscription.send({ target, args, url });
+    stimulate () {
+      clearTimeout(controller.StimulusReflex.timeout)
+      const args = Array.prototype.slice.call(arguments)
+      const target = args.shift()
+      const url = location.href
+      controller.StimulusReflex.subscription.send({ target, args, url })
     },
-  });
-};
+    reflex (el) {
+      clearTimeout(controller.StimulusReflex.timeout)
+      const name = el.target.dataset.reflex.split('#')
+      const target = `${name[0].charAt(0).toUpperCase() +
+        name[0].slice(1)}Reflex#${name[1]}`
+      let args = []
+      for (const arg in el.target.dataset) {
+        if (/^reflexArg/.test(arg)) args.push(el.target.dataset[arg])
+      }
+      const url = location.href
+      controller.StimulusReflex.subscription.send({ target, args, url })
+    },
+    wire () {
+      document
+        .querySelectorAll('[data-reflex]')
+        .forEach(el => el.addEventListener('click', controller.reflex))
+    },
+    unwire () {
+      document
+        .querySelectorAll('[data-reflex]')
+        .forEach(el => el.removeEventListener('click', controller.reflex))
+    }
+  })
+
+  document.addEventListener('cable-ready:before-morph', controller.unwire)
+  document.addEventListener('cable-ready:after-morph', controller.wire)
+
+  controller.wire()
+}
 
 export default {
   //
@@ -54,10 +80,10 @@ export default {
   //   * renderDelay - amount of time to delay before mutating the DOM (adds latency but reduces jitter)
   //
   register: (controller, options = {}) => {
-    const channel = 'StimulusReflex::Channel';
-    const room = controller.element.dataset.room || '';
-    controller.StimulusReflex = { ...options, channel, room };
-    createSubscription(controller);
-    extend(controller);
-  },
-};
+    const channel = 'StimulusReflex::Channel'
+    const room = controller.element.dataset.room || ''
+    controller.StimulusReflex = { ...options, channel, room }
+    createSubscription(controller)
+    extend(controller)
+  }
+}


hopsoft:
Thanks for the ideas and the PR. I'm excited to dig into this later. Also, I'll be sure to add a prettier config to the project 😉

hopsoft:
This is very interesting. I like the idea of a Stimulus inspired declarative option that allows developers to skip the JavaScript altogether; however, I'm not a fan of limiting the HTML elements that this might be applied to.

Here's a proposal for the declarative API. _Heavily inspired by what you've done._

```html
<div data-reflex="click->ServerSideReflex#do_something" 
     data-reflex-argument1="foo" 
     data-reflex-argument2="bar">
</div>
```

```ruby
class ServerSideReflex < StimulusReflex::Reflex
  def do_something(a, b)
    # hard work
  end
end
```

---

Another option might be to lean on standard HTML for the arguments. Something like:

```html
<div data-reflex="click->ServerSideReflex#do_something">
  <input type="hidden" name="reflex[]" value="foo">
  <input type="hidden" name="reflex[]" value="bar">
</div>
```

I think I prefer the first style. What do you think?

pjforde1978:
Now it's my turn to be super-fuzzy on the details. How is proposal #1 different from my PR's API?

andrewmcodes:
@pjforde1978 forgive my ignorance and lack of JS skill but can you explain what this is doing/problem it solves/use case? 

Also from a layman’s view it does seem you are limiting the html elements this can be applied to. Looks like you are only applying to the [shorthand pairs](https://stimulusjs.org/reference/actions). Would this still work if applied to a div, icon, etc.?

hopsoft:
Ah yes. I clearly see now that you're simply matching the Stimulus shorthand syntax for `actions`. https://stimulusjs.org/reference/actions#event-shorthand _I should have followed the link in your comments._ 😉 

I suppose the only difference may be the semantics of `data-reflex-argN`

I'll try this out on a project or two and update this PR with my experiences and any proposed changes.

pjforde1978:
@andrewmcodes just to be clear, I submitted a PR that had exactly one line of commented URL as meaningful documentation, so if there's any failing it's on me and will be resolved as soon as Mr. Hopkins makes a call on API syntax. ;)

As for the semantics... you know the old parable about how today only you and God understand your code but in [three weeks, apparently] it's only God?

I submitted a parallel PR to the todomvc project that demonstrated the proposed changed working. Sadly, the PR appears to be in conflict with the current master branch. Did you guys have a chance to try it out before it broke? I'm actually going to go update that PR to make sure that there's a solid reference in place.

Anyhow, the way my proposed patch handles arguments is that it accepts any argument in the order it is declared. This is because Reflex passes an array of argument values instead of a hash of named pairs.

Additionally, I provided a second attribute check that passes the current value of the HTML element as the argument. There could well be other HTML attributes we'd want to support, such as checked. I haven't even considered how input type=file is supposed to work... but for now, that's as far as I got.

pjforde1978:
Alright, I just added checkbox support. Game on, boys!

pjforde1978:
Heads-up: I just merged + resolved conflicts in the todomvc which might be instructive to see what's possible with passive, attribute-driven Reflex.

https://github.com/hopsoft/stimulus_reflex_todomvc/pull/4

Get it before it breaks again! ;)

javascript/stimulus_reflex.js
```diff
@@ -1,48 +1,92 @@
...
+          : 'addEventListener'
+      const events = {
+        a: 'click',
+        button: 'click',
+        form: 'submit',
+        input: 'change',
+        select: 'change',
+        textarea: 'change'
+      }
+      document.querySelectorAll('[data-reflex]').forEach(el => {
```
pjforde1978: This 100% needs to be scoped to the DOM element that the controller is declared on, or else you'll create duplicate events.

pjforde1978:
One thing I just realized after re-working todomvc is that I never actually defined an HTML element to be the filter controller. I also never defined a node to be the todos controller. In other words, the connect() method of both controllers is never being called, which in turn means that StimulusReflex.register() is only being called once on the page.

If I hadn't forgotten those details, it would have been called thrice and there would be three events per passive element because my scan is scoped to the document. Whoopsie!

https://github.com/hopsoft/stimulus_reflex/blob/9c34cc58c904b404fb05bf2a861a0ca4ff99c57e/javascript/stimulus_reflex.js#L71

I'll patch this once I have figured out the best practices way to reference the controllers' parent element directly.

hopsoft:
It would be great if we provided a default StimulusReflex client side controller that auto registers all `data-reflex` elements. Then you theoretically could get away without the need to write any JavaScript at all.

pjforde1978:
Seems like a reasonable goal

That does seem to be the main thrust of Chris' pitch!

So hey, one thing we've never talked about is what I see as the third-leg of the Stimulus paradigm, which is the concept of a target. I've never fallen in love with targets but it's 100% (for me) a matter of taste and style and preference. That said, I've also never really forced myself to embrace them and so maybe I am missing out big time. I dunno! However, I am writing this because I just realized that I also haven't seen you writing anything with targets and maybe we both silently embraced each other's personal choice while completely forgetting that the "average" Stimulus user might come across this and wonder why we're completely ignoring targets. Is this even something I should be spending brain cells wondering about? Serious question.

If we should at least have a short conversation about it, maybe that conversation is "I've thought about it and it's completely irrelevant to what we're doing". That works for me. But is there some aspect of how people expect to be using Stimulus that Reflex messes with? I haven't eaten yet so I apologize if this sounds disturbingly like a fever dream.

The way I like to use Stimulus is that I typically have one instance of one controller for the page "running stuff" and there are some cases where I have custom UI elements that have their own thing going on with a separate controller in addition to the page controller. My page controllers are always scoped to the body tag. When I'm using Stimulus Reflex, I call it in my connect() and I'm good to go. Passive reflexes mean that my controllers are usually quite short because they represent only the exceptions that I can't handle using passive declarations.

I noticed that on the todomvc there's three controllers active and - at least the way you had it - there's a todo controller on every item. I'm guessing that this is a standard convention in iOS development but most of my MVC experience has been in Rails where there's always exactly one controller in play at any given time... so that's had a big influence on how I structure my Stimulus controllers.

All of this is a long way of me saying that since I always have one page-level controller that's scoped to my body tag, I can reliably call Reflex once during my page initialization step. If everyone coded like this, we could make Reflex initialize by (for example) including it in the app/javascript/controllers/index.js. However, because this is by no means a standard for everyone, perhaps the best way to handle initialization is to support an optional boolean 2nd parameter to register() that controls whether or not to call the passive wire methods and install the event handlers. Or perhaps there's other options to explore. The nice thing about being a relatively unknown project is that we still have plenty of room to sculpt the API before a larger community of people get annoyed. This is the literal perfect time to hash this stuff out.

pjforde1978:
I don't know why it keeps giving me such a hard time about some random semi-colons but the patch now scopes its searching and event management to the controller's host HTML element. While this is clearly a positive step, I'm concerned that you can still get duplicate events if you nest Stimulus controllers in a hierarchy.

One additional potential approach to preventing duplicate event handlers is that we could set a data attribute on the grandparent <html> element denoting that Reflex has already been instantiated, no need to do it again. Similarly, you could set a global variable on the window namespace. Same end-result, while offending different people. :)

It's good to have options, anyhow.

I've going to re-watch Chris McCord's presentation and take notes on how he steps through the mental landscape of getting his audience excited. He does a lot of things right, but there's also a lot that could be cut out or trimmed because it's intended for a nerdy audience... by which I mean implementors and not necessarily the pluralist rank-and-file web devs, who are not attending Elixir conferences in Prague. I really believe that there's a parallel to the "blog in 15 minutes" for Reflex that will be jaw-dropping to a significant audience.

hopsoft:
## Declarative Reflexes

First off, I want to thank you for advancing the idea of passive/declarative reflexes.

---

### This proposal aims to provide the option of creating reactive StimulusReflex applications without the need to write any JavaScript whatsoever.

---

Bear in mind that this will be an abstraction on top of the existing StimulusReflex libary and API.

If user's opt-in to the proposed syntax/markup, a client side Stimulus controller would be implicitly created for the respective DOM elements and would be registered with StimulusReflex.

#### Rejected

I'm rejecting the idea of matching [Stimulus' event shorthand](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/stimulus_reflex/channel.rb#L42-L48) style for common elements + actions for 2 primary reasons.

1. I don't want to be chasing the Stimulus implementation.
1. I think the explicitness of declaring `data-stimulus` will prove helpful. See below.

### Initial API Proposal

### Client

The DOM element's [dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) attributes would be serialized and passed as an `options` hash as the last argument to the server side reflex method.

```html
<!-- app/views/example.html.erb -->
<div data-stimulus="click"
     data-reflex="ExampleReflex#work"
     data-foo="value"
     data-bar="false">
</div>
```

### Server

All server side reflex methods would be required to accept an options hash as the last argument.
The `options` argument would be a server side representation of the invoking DOM element's [dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset).

_[We may be able to make this optional if we can dynamically detect support based on method arity...](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/stimulus_reflex/channel.rb#L42-L48) prior to invoking the reflex._

```ruby
# app/reflexes/example.rb
  def work(options = {})
    foo = options[:foo]
    bar = options[:bar]
    # do the work
  end
end
```

pjforde1978:
This is excellent stuff, Nate!

When you say existing API, I assume you mean that you chose data-stimulus so that it could sit alongside data-action and not impact existing explicit calls. The only thing I'd say is that I'm not 100% convinced data-stimulus is the ideal name but I am 90% convinced and not currently sitting on any better suggestions. data-reflex-action?

I am fine with dropping support for shorthand, as it doesn't make sense with the new API.

I am thrilled by the idea of an options hash. It is clearly superior to the current situation and will be far less bug-prone. Will you filter out some of the attributes such as data-action, data-reflex and [maybe] data-stimulus or send the whole batch?

One thing your API proposal doesn't seem to cover is arguments based on element attributes such as value and checked. I believe these are important - especially value. And once you have value, then you're committed to checked and possibly others. Do you have any thoughts on how that could fit into the picture?

Finally, just some philosophy: I don't know if I'm fully enthused by the "any Javascript whatsoever". We've mentioned it a bit in relation to Chris and his talk, but I just don't think it's plausible that the "no Javascript" promise is ever really true. There's going to be some exceptions and gotchas in just about every application of any sophistication... especially if typing characters in a box is considered sophistication. Obviously, we all have strong opinions about providing options to the developer community that don't involve spinning up React to build a CRUD app. I feel like there has to be exciting messaging that finds a middle-ground in terms of getting people excited about what's possible. It's an ongoing conversation in my home right now. :)

hopsoft:
> I'm not 100% convinced data-stimulus is the ideal name

Somewhat agree but the pairing of `data-stimulus` + `data-reflex` works quite well given the library name.

> API proposal doesn't seem to cover is arguments based on element attributes such as value and checked

Absolutely an important point. I've been thinking on how to best get these types of values passed to the server side reflex method. I may end up packing all of the DOM element's attributes into the options hash. Something like this will be needed, I'm just not sure what shape it will have yet.

> There's going to be some exceptions and gotchas in just about every application of any sophistication

Agree with this. I think the sweet spot to communicate to developers will be along the same lines of Rails ujs. No specific claims that you won't need to write JavaScript, just clear documentation regarding what the library does. Essentially, **"Look at all the things I'm not doing."**

pjforde1978:
How cool would it be if the reflex functions received two hashes as arguments; _options_ and _element_. (Although, if we're designing an ideal API this might be the time to turn _option_ into _attributes_ or _arguments_.]

You probably don't want to pack *all* of the element attributes. There's a whack of them. In fact, so far my list only contains value and checked... and checked is only for checkboxes. I am plagued by the feeling that I'm forgetting more, but you know what I mean.

PR #19, opened 2019-09-08
Wait for it (proposal)

pjforde1978:


pjforde1978:
Note that it's just lib/stimulus_reflex/reflex.rb that is relevant to this PR. Sorry, still learning the ropes!

pjforde1978:
This might get added to the library at some point, but for now, the most recent version of the function and instructions on its use are at https://app.gitbook.com/@stimulusreflex/s/stimulusreflex/patterns

Closing for now.

PR #21, opened 2019-09-08
Implicitly send DOM attributes to reflex methods

hopsoft:
## Implicit options hash passed to server side reflex methods

Foundational feature to support some of the items proposed on #12 Note that I'd like to see this PR merged before #12 and others as this is foundational. 

1. Server side reflex methods can now accept both required and optional arguments.

```ruby
  def work(required_arg, optional_arg = "value")
    # do the work
  end
end
```

```ruby
  def work(required_arg, *optional_args)
    # do the work
  end
end
```

2. Server side reflex methods can now accept an optional keyword argument of `options` which holds all DOM attributes of the Stimulus DOM element.

```ruby
  def work(required_arg, options:)
    # do the work
  end
end
```

```ruby
  def work(required_arg, options: {})
    # do the work
  end
end
```

```ruby
  def work(required_arg, optional_arg = "value", options:)
    # do the work
  end
end
```

**Note that that the `options` keyword argument value is a Hash like object but is not a Ruby Hash.** However, for most purposes it can be treated as an idiomatic options hash.

The following DOM element

```html
<checkbox label="Example" checked="true" data-controller="checkbox" data-value="123" />
```

Produces the following behavior in a reflex method

```ruby
  def work(options:)
    options[:label] # => "Example"
    options[:checked] # => "true"
    options["data-controller"] # => "checkbox"
    options["data-value"] # => "123"
    options.dataset[:controller] # => "checkbox"
    options.dataset[:value] # => "123"
  end
end
```


pjforde1978:
Coupla' questions:

1. I'm new to the options: syntax. What does this imply, in literal terms? eg. what's the difference between options: {} and options:

2. I was going to suggest HashWithIndifferentAccess, so we're on the same page. One nice thing about a Rails project is that we can safely assume ActiveSupport is in the picture. I do wish that it was _truly_ indifferent access; I'd love to be able to say options.value or options.dataAttributeX but that's probably sacrilege.

3. I agree that packing element.attributes is a solid way to handle value and checked. Nice thinking on this. However, the major gotcha here is the value of a textarea? For _reasons_, the attributes collection doesn't include value even though you can certainly access element.value

hopsoft:
> 1. I'm new to the options: syntax. What does this imply, in literal terms? eg. what's the difference between options: {} and options:

`options:` is a required keyword argument. `options: {}` is an optional keyword argument with a default value. This PR supports both, so it's really just a syntactic style preference for developers using StimulusReflex. Note that the `options` keyword argument can also be entirely omitted if it's not used by the reflex.

> 3. ... For reasons, the attributes collection doesn't include value [of textarea]

We will definitely have to cover this scenario. Are there any other DOM gotchas like this that you can think of?

hopsoft:
I think I may have a simple solution that covers most edge cases like `textarea`

SEE: https://github.com/hopsoft/stimulus_reflex/pull/21/commits/4f926c385d7357e45a427bf2af237aa2d5ebcc6c

Thoughts?

pjforde1978:
Good catch on the multiple select values. My memory is that Rails creates an array, but I rarely work with multi-selects. Serializing the values into an array seems about roughly the only viable option, no?

I've been doing some Googling to try and uncover any other gotchas, and in the process I've back-filled a lot of the things about element attributes vs DOM attributes that I've managed to stay blissfully ignorant of thanks to jQuery for so many years. I'm going to do a minor brain dump and some of this is likely old news to you.

Each DOM node has an attributes property, which is a NamedNodeMap (essentially an array with a few extra features). In particular, this means you can get element.attributes.length and loop through them. Each individual attribute is an Attr object, which has (among other things) name and value properties.

Accessing element.attributes returns the list of attributes defined in the HTML markup **at the time of access**, and could theoretically be out of sync with the DOM value. This is why the textarea doesn't have value in its attributes collection; the value is defined by innerHTML and only made available as the value accessor as part of the DOM API after the fact.

Setting a value attribute via setAttribute() updates the default value, triggering a redraw of the element and typically updating the DOM. It's generally preferential to get and set DOM values instead of changing element attributes for performance reasons. https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#Notes

As a rule of thumb, only use them for attributes that don’t have a corresponding DOM property (such as colspan), or if you really want to “persist” those changes to the HTML (e.g. to keep them when cloning an element or modifying its parent’s .innerHTML)

I attempted to find a way to extract a list of element attributes && node values that reconciles nicely, but so far I'm coming up short.

```
const node = $0 // assumed to be the textarea I'm typing into on GH for discussion

// outputs the long list of attributes, including value
for (let key in node) {
  console.log(key)
}

// generates no output, which stumps me as I was expecting value etc
for (let key in node) {
  if (node.hasOwnProperty(key)) console.log(key)
}

// returns false, even after I type some characters into the textarea
node.hasOwnProperty('value')

// returns false, which is expected as it's not in the HTML
node.hasAttribute('value')

// no output
for (let key in node.attributes) {
    if (key.specified) console.log(key)
}

// generates an array of strings based on current HTML
Array.from(node.attributes)

// adds a data-test1 attribute to the HTML, viewable in Inspector
node.setAttribute('data-test1', "pete")

// adds a data-test2 attribute to the HTML
node.dataset.test2 = "pete"

// run after the previous two statements
Array.from(node.attributes)
// returns [name, id, placeholder, aria-label, class, required, spellcheck, data-test1, data-test2]

// now this is interesting
for (let key in node) {
    if (node.getAttribute(key)) console.log(key, node.getAttribute(key))
}
```

That last command outputs this:

> name comment[body]
> placeholder Leave a comment
> spellcheck false
> id new_comment_field

Note that I'm just assuming all contemporary evergreen browsers play by the same rules. I know that some versions of IE 11 and below start doing weird things eg. element.attributes returns all possible attributes even if not defined in HTML.

Ultimately and in conclusion for now, I'm more confused than ever but I now feel like I have a solid working understanding of the very interesting line between the document markup - which can remain fluid and trigger DOM redraws - and the DOM memory structure itself, which can get out-of-sync with the document markup. **I worry about the potential for race conditions** and strange regressions. Do we have a minimum browser spec on this project.

hopsoft:
> Do we have a minimum browser spec on this project?

Not yet, but we should consider setting a target for this.

PR #23, opened 2019-09-09
Access controller instance variables

pjforde1978:
This is a super rough preview of a proposal to support accessing the instance variables from the controller inside of the reflex.

The implementation probably isn't efficient but the basic idea is that in most index or show actions, you end up setting a bunch of important instance variables that would likely be highly useful to have access to in the reflex. After all, in many cases the reflex takes the place of the update action, so it's good to know what you're updating. Otherwise, I suspect people will reimplement copies of their index/show actions from their controllers inside of the reflexes, and that's a code smell.

You can see the rabbit hole I went down; first, I tried to override initialize but quickly realized that I have no opportunity to set an initial value for something on the initial page load from the reflex. So I borrowed the render_page function and turned it into a controller action triggering machine. I set a variable called @stimulus_reflex_no_render so that the controller action can optionally render(body: nil) which should mean ERB is never instantiated and no time is wasted.

The instance variables have their @ stripped off and are then injected into a HashWithIndifferentAccess that is accessible to the reflexes as @action_variables, with a better name suggestion welcome.

If we keep this around, I recommend consolidating render_page and access_controller_instance_variables and possibly keeping the setup in scope so that we can call controller.process multiple times without tearing it down in the middle.

Note that this PR is not very useful unless you're also running:

https://github.com/pjforde1978/reflex/tree/access_controller_instance_variables


hopsoft:
I may fully not understand what you're trying to accomplish, but I do question whether or not this is necessary given the following order of operations.

1. Invoke a server side Reflex method
1. Instantiate a Rails controller and invoke the appropriate action method

The Reflex is called first. Also any instance variables set in the Reflex method are made available to the Rails controller action. This means you can do something akin to this.

```ruby
class UserReflex < StimulusReflex::Reflex
  def example(user_id)
    @user = User.find(user_id)
  end
end
```

```ruby
class UsersController < ApplicationController
  def show
    @user ||= User.find(params[:id])
  end
end
```

If `users#show` is called from the Reflex method, the user instance variable will already be set. Granted this feature needs to be better documented.

pjforde1978:
I actually do make use of ||= in my page#index and think I get the pattern && order of operations. Phew!

The issue is a Venn cluster of concerns around DRY code and initial state in the context of an interaction life cycle.

I’ve found that when an application progresses beyond todomvc example, the controller actions for many interfaces grow in complexity, ultimately doing potentially several joins and who knows what else. The idea of maintaining this code in two places is not great, although perhaps it’s a sign that further refactoring and moving code out of the controller is apropos.

Anyhow! I did my best to illustrate what I’m trying to solve with the slow API call example, which has three states.

It starts with a button, switches to “loading” and ends with a success message. The first render won’t have an instance variable available to set the state... although it occurs to me that this could be the else case in a switch.

I’m going to keep thinking about this. Perhaps I am overthinking what’s actually required. It’s also fine to keep this in my own bag of tricks for exceptional circumstances.

PR #28, opened 2019-09-10
Add support for declarative stimulus/reflex behavior

hopsoft:
Add declarative stimulus/reflex behavior via data attributes.

- `data-reflex` - accepts a Stimulus styled value to specify DOM event & server side reflexes i.e. `click->ExampleReflex#method`

Very much inspired by #12

Example usage here: https://github.com/hopsoft/stimulus_reflex_todomvc/pull/16

I have a question out to the Stimulus team to determine if the techniques used here are valid. i.e. calling `application.start` inside a library. https://github.com/stimulusjs/stimulus/issues/272

### DIFF
--- a/javascript/yarn.lock
+++ b/javascript/yarn.lock
@@ -2,6 +2,30 @@
 # yarn lockfile v1
 
 
+"@stimulus/core@^1.1.1":
+  version "1.1.1"
+  resolved "https://registry.yarnpkg.com/@stimulus/core/-/core-1.1.1.tgz#42b0cfe5b73ca492f41de64b77a03980bae92c82"
+  integrity sha512-PVJv7IpuQx0MVPCBblXc6O2zbCmU8dlxXNH4bC9KK6LsvGaE+PCXXrXQfXUwAsse1/CmRu/iQG7Ov58himjiGg==
+  dependencies:
+    "@stimulus/mutation-observers" "^1.1.1"
+
+"@stimulus/multimap@^1.1.1":
+  version "1.1.1"
+  resolved "https://registry.yarnpkg.com/@stimulus/multimap/-/multimap-1.1.1.tgz#b95e3fd607345ab36e5d5b55486ee1a12d56b331"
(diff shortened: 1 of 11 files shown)


hopsoft:
I'm curious what other's think about the use of 2 data attributes vs 1. The other proposal was a single data attribute that matches Stimulus semantics. See #12 for more context.

```html
<div data-reflex="click->ServerSideReflex#do_something">
```

-vs-

```html
<div data-stimulus="click" data-reflex="ServerSideReflex#do_something">
```

hopsoft:
I'm starting to lean toward the single data attribute as initially proposed by @pjforde1978

javascript/yarn.lock
```diff
@@ -2,6 +2,30 @@
 # yarn lockfile v1
 
 
+"@stimulus/core@^1.1.1":
```
andrewmcodes: Shouldn't this reflect v2.0.0?

lib/stimulus_reflex/channel.rb
```diff
@@ -22,11 +22,11 @@ def receive(data)
     reflex_name, method_name = target.split("#")
     reflex_name = reflex_name.classify
     arguments = data["args"] || []
-    options = StimulusReflex::DomElement.new(data["attrs"])
+    element = StimulusReflex::Element.new(data["attrs"])
```
andrewmcodes: Been thinking about this more and I still like element.

Seems inline with the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset)

javascript/yarn.lock
pjforde1978: Is Stimulus v2 imminent?

andrewmcodes: @pjforde1978 the changes in this PR that Nate and I paired on are breaking so I’d say yes though it’s obviously ultimately up to Nate.

hopsoft: That dep is for Stimulus itself rather than StimulusReflex.

PR #31, opened 2019-09-11
update license

andrewmcodes:
update license to reflect the current year and move file to markdown so it reflects in the GH UI.


hopsoft:
Not sure that updating the copyright date is necessary... or perhaps even desired. SEE: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated#answer-2391555

PR #37, opened 2019-09-13
Allow override of default controller

hopsoft:


pjforde1978:
Just a bit of late-breaking thought on an otherwise perfect commit...

I just implemented my first override controller, and it works well, but I had to do some extra bits to get it going. It will have to be documented, is all.

So the basic pattern is that you can/should start off by taking a copy of stimulus_reflex_controller.js from the package/gem and putting it in your app/javascript/controllers folder. I called mine reflex_controller.js.

When I first ran it, I got nasty errors because this.constructor doesn't have a register() method unless you manually set it. You do this for the default controller by assigning the register method manually.

This leaves two options:

The dev can either add 'stimulus_reflex' to their custom controller and change this.constructor.register() to StimulusReflex.register()

or:

The dev can do the manual assignment in their index.js

I'm honestly not sure which one is cleaner. I chose to do it in index.js but mostly to prove to myself that I understood what was going on. It seems like I should convert to importing and call it that way.

However, that brings up the more pertinent question: is there a reason that you didn't just import reflex in the default controller and reference it directly? I don't want to miss anything.

Right now, this is my index.html:
```
import { Application } from 'stimulus'
import { definitionsFromContext } from 'stimulus/webpack-helpers'
import ReflexController from './reflex_controller'

const application = Application.start()
const context = require.context('controllers', true, /_controller\.js$/)
application.load(definitionsFromContext(context))
ReflexController.register = StimulusReflex.register
StimulusReflex.initialize(application, ReflexController)
```

It seems like this would keep everything contained to one file that people can copy and modify without having to think about it. It would also allow you to remove the `StimulusReflexController.register = register` line from stimulus_reflex.js.

PR #38, opened 2019-09-14
Reflex is a reflex

pjforde1978:
# Bug Fix

## Description

reflex_class is going to be TodosReflex, for example, not StimulusReflex::Reflex

Need to test subclass, not equality.

## Why should this be added

Reflexes will not run unless this is fixed.

## Checklist


pjforde1978:
If anyone wants to give me a lesson in collapsing commits, I have no ego and will not be shamed.

pjforde1978:
That was actually the first thing I tried, and it didn’t work. I didn’t even realize that you could use < as an inheritance operator until last night.

Caveat, I was very tired so if is_a? decides to start working again, amazing.

PR #39, opened 2019-09-15
Callbacks and redirects

pjforde1978:
# Feature

## Description

I've added support for callbacks and redirects. They ended up being very closely related in terms of implementation, so it made sense to do both.

The Ruby library now passes enough metadata back to the JS client that it can unpack the name of the calling Stimulus controller and the method that triggered the reflex. It uses this to obtain a reference to the correct controller and run the callback.

Implicitly, it will check the controller for the existence of a callback that follows the _onMethod_ naming convention. So if my reflex is "TodosReflex#edit" then it will go to the Todos controller and attempt to call the onEdit() method if it exists.

Explicitly, the developer can set @ callback in their Reflex class to indicate which callback should be called. This is useful if you have multiple reflexes which should call the same client-side callback - no need to repeat code.

The developer can also set @ redirect in their Reflex class to initiate a redirect. The client will test for the redirect and either redirect or launch the morph operation. It first tests for the presence of Turbolinks and will call Turbolinks.visit() on whatever they set @ redirect to; barring that, it will resort to the classic window.location assignment.

Sample PR is set up here:

https://github.com/hopsoft/stimulus_reflex_todomvc/pull/19

I added simple examples of how both @ callback and @ redirect would work in the TodosReflex class. They are currently commented out.

Note: I've sort of fallen out of love with the ability to supply your own customized StimulusReflexController. I no longer think that it's necessary and suggest it could be removed. Sorry for the false start.

## Why should this be added

Callbacks and redirects are necessary to build real web applications. I did this in a way that strongly borrows from Rails' philosophy of convention over configuration while allowing flexibility.

I strongly suspect that Nate will have strong opinions about how this should be implemented and won't be sad if this all goes away. My payoff is always in getting it working.

## Checklist


pjforde1978:
Not sure what's going on with Prettier and Standard linters, sorry! They are running on my end.

Also, always happy to give Zoom demos to anyone interested.

pjforde1978:
I agree that there could be an opportunity to add an operation to cable_ready for redirect; I hope that it checks for Turbolinks because that part of it registers as genuinely clever to me.

As for the API, my suggestion is that we try out both... possibly before you push it live so we have a chance to tweak it and debate pros and cons this time. :grinning:

My thoughts are these:

I love the implicit handling of predictably named events; taking a page from Rails here, big time. The convention part is powerful.

Unfortunately, callbacks - whether implicit or explicit as DOM events would require - are not optional so far as I can tell, and I consider that true on the client and server. My wait_for_it patch on the server is not a ton of code, but it's a pattern that compels the dev into proper use and without it things like responding to a long-running call are all but impossible. It's even worse on the client; there are things like setting focus which simply cannot be done via HTML.

Luckily, the amount of actual JS that people will have to write is quite small and we can provide excellent examples that demonstrate what those edge cases look like. For example, you can see how I slightly changed cancelEdit() to prevent the reflex event call from firing twice; once on the keydown and once on the blur.

Ultimately, my concern with every dev building their own event management cascade based on the after-morph event is that you'll end up with an increasingly ugly decision tree of custom logic that each dev will have a slightly different structure for.

In the battle between one event and a case switch, and a 1:1 world where every case has its own 1-3 liner, I like the granular approach.

It's also possible to allow both styles of API. However, DHH's voice suggests that only one of them should be the recommended default approach, even if you document the others.

Anyhow, we should both try to make some increasingly more elaborate examples to test both approaches before making a decision. I'm happy to have my opinion changed!

javascript/stimulus_reflex.js
```diff
@@ -21,7 +21,29 @@ const createSubscription = controller => {
...
+                );
+                if (
+                  data.operations.morph[0].callback &&
+                  typeof targetController[data.operations.morph[0].callback] === 'function'
+                ) {
+                  targetController[data.operations.morph[0].callback]();
+                } else {
+                  let method = data.operations.morph[0].method;
+                  method = 'on' + method.charAt(0).toUpperCase() + method.slice(1);
+                  if (typeof targetController[method] === 'function') {
```
pjforde1978: See how I do this?

PR #41, opened 2019-09-16
Update to use prettier-standard for js linting/formatting

### LINKED ISSUE #40: Install StandardJS linter
After much blinking and head-desking, I finally figured out the linting situation. It wasn't what was happening... it was what isn't happening.

I learned about StandardRB through the actions set up on this project. Not as far along, but promising and I used a [post-save hook](https://github.com/emeraldwalk/vscode-runonsave) to set it up to run in VSCode because StandardRB not available as an extension yet. `bundle exec standardrb --fix $(file)` for great success.

However, I've been thoroughly flummoxed by the JS lint situation, because StandardJS is so opinionated about (no) semi-colons. What I failed to grok until just now is that my brain assumed two incorrect things: that Prettier only tackled column-width and line-breaking, and that we had StandardJS set up in our post-commit and post-pull-request callbacks on GH.

I was wrong about Prettier - it inserts "missing" semi-colons and, well, I wish you could see my face the moment I realized that when GH's Actions interface said "Run Standard Lints" it meant only StandardRB.

:face_with_head_bandage:

[Javascript Standard](https://github.com/standard/standard) is a thing of beauty. [I find] Ruby code lends itself to readability, but JS's obsession with using nested object notation (it used to be far, far worse with nested anonymous function hell) makes the Prettier + Standard combo the first thing I set up in any JS project. There are many VS Code plugins; I recommend [Prettier-Standard](https://github.com/numso/prettier-standard-vscode).

Anyhow: can we please, please, pretty please set up StandardJS to run after Prettier? This will, predictably, solve the semi-colon thing.

hopsoft:
Fixes #40

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -3,7 +3,7 @@
   "version": "2.0.0",
   "main": "./stimulus_reflex.js",
   "scripts": {
-    "prettier-check": "yarn run prettier --check ./stimulus_reflex.js"
+    "prettier-check": "yarn run prettier-standard --check ./stimulus_reflex.js"
   },
   "repository": {
     "type": "git",
@@ -17,6 +17,6 @@
     "stimulus": ">= 1.1.1"
   },
   "devDependencies": {
-    "prettier": "^1.18.2"
+    "prettier-standard": "^14.0.5"
   }
 }
--- a/bin/standardize
+++ b/bin/standardize
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 bundle exec standardrb --fix
-cd ./javascript && yarn run prettier --write stimulus_reflex.js
+cd ./javascript && yarn run prettier-standard --format stimulus_reflex.js
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -34,7 +34,7 @@ const createSubscription = controller => {
 
 const extend = controller => {
   Object.assign(controller, {
-    stimulate() {
+    stimulate () {
       clearTimeout(controller.StimulusReflex.timeout);
       const url = location.href;
       let args = Array.prototype.slice.call(arguments);
--- a/javascript/yarn.lock
+++ b/javascript/yarn.lock
@@ -2,6 +2,22 @@
 # yarn lockfile v1
 
 
+"@babel/code-frame@^7.0.0":
+  version "7.5.5"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.5.5.tgz#bc0782f6d69f7b7d49531219699b988f669a8f9d"
+  integrity sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==
+  dependencies:
+    "@babel/highlight" "^7.0.0"
+
+"@babel/highlight@^7.0.0":
+  version "7.5.0"
+  resolved "https://registry.yarnpkg.com/@babel/highlight/-/highlight-7.5.0.tgz#56d11312bd9248fa619591d02472be6e8cb32540"
+  integrity sha512-7dV4eu9gBxoM0dAnj/BCFDW9LFU0zvTrkq0ugM7pnHEgguOEeOz1so2ZghEdzviYzQEED0r4EAgpsBChKy1TRQ==
+  dependencies:
+    chalk "^2.0.0"
+    esutils "^2.0.2"
+    js-tokens "^4.0.0"
+
 "@stimulus/core@^1.1.1":
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/@stimulus/core/-/core-1.1.1.tgz#42b0cfe5b73ca492f41de64b77a03980bae92c82"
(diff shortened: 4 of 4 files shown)


hopsoft:
@andrewmcodes Will we need to update the Actions if/when this PR merges?

javascript/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "2.0.0",
   "main": "./stimulus_reflex.js",
   "scripts": {
-    "prettier-check": "yarn run prettier --check ./stimulus_reflex.js"
+    "prettier-check": "yarn run prettier-standard --check ./stimulus_reflex.js"
```
andrewmcodes: @hopsoft shouldn’t need to update the action. It just executed this script. Since you’ve updated it, it should continue working.

pjforde1978:
This is great! I really appreciate everyone being willing to go here.

One thing that I'm trying to figure out is that the prettier-standard JS module seems to output what is known as standard-semistrict, which is standard + semi-colons. I'm definitely trying to move us past the semi-colon era, so bear with me...

When I install and run standard and yarn run standard --fix it outputs as expected, including picking up a subtle error that prettier-standard somehow missed; in stimulus_reflex.js we should be using window.location.href instead of just location.href.

Anyhow, what's weird is that the prettier-standard plugin I use for VSCode is definitely running in standard mode by default (no semi-colons) so I was caught a bit off-guard by the output of the JS module.

Also still distinctly possible that I'm missing something. I'm new to linting!

@andrewmcodes thanks so much for the tip about the ruby plugin already supporting StandardRB. I am seriously so lucky to have smart people around me to learn from every day.

andrewmcodes:
@pjforde1978 not a problem! I try to be active on that repo but the code is above my head for the most part. 

Curious if you could enumerate the benefits of this library instead of just using prettier? I’m relatively familiar with the StandardJS project but have never found it really compelling.

andrewmcodes:
At the end of the day I don’t really care as long as the tool will do the formatting for me `¯\_(ツ)_/¯`

pjforde1978:
The main thing Prettier does is break up long lines and chunks things like objects into proper indentation. Also: semi-colons, apparently! I honestly never realized this until yesterday.

StandardJS has a long list of highly-opinionated lint checks that dictate everything from whitespace to 2-space tabs to how ternary operations should be formatted. It's a really long list!

https://standardjs.com/rules.html

The main answer is really just "consistency". It's exceptionally well-reasoned and a huge number of smart people have contributed. What I've found is that even the rules I didn't like, I came to appreciate.

hopsoft:
Prettier certainly seems to be winning the mindshare. https://www.npmtrends.com/prettier-vs-standard

hopsoft:
Note that ESLint defaults to semicolons. https://eslint.org/docs/rules/semi#options
Prettier also defaults to semicolons. https://prettier.io/docs/en/options.html#semicolons

I'll likely stick with the majority regarding semicolons.

hopsoft:
We need to land on an actual standard because the diffs are starting to prove too verbose and noisy with irrelevant discrepancies to effectively manage on limited time. One thing we can do in the short is run `bin/standardize` in a terminal prior to commit to prevent editor configs/preferences from overriding.

pjforde1978:
My vote isn’t to pick one, but to run standard after prettier. This post does an excellent job of explaining why most of those in the “always” camp leave themselves open to ASI introducing subtle bugs even if you stick to them religiously. 

https://feross.org/never-use-semicolons/

ESLint does semi-colons explicitly because there’s a concern that making them non-optional would expose people to thinking they could do loops without them.

Anyhow, aside from the semi-colons issue, which is cosmetic, the reason I’m pushing hard for standard is that it actually enforces consistent coding practices and catches real bugs. Prettier is... aptly named.

Consider this: there might well be a larger group of people willing to submit to standard’s standards. I don’t typically want to work on their team. :grin:

Anyhow, in short I think popularity is toxic and this graph might be more nuanced than popularity. Standard is aptly named - it’s dominant and it’s not concerned about the things Prettier focuses on.

React is really popular, too.

pjforde1978:
That the prettier-standard linters on VSCode and npm output different results is a massive FAIL.

andrewmcodes:
@pjforde1978 curious what that diff would look like if run on the entire project. Can you provide that insight.

pjforde1978:
I have seen large diffs that git won’t merge because every line appears different and it’s not pretty.

I’m actually going to push commits with linter passing js right after dinner for all of my outstanding PRs because I genuinely want you to be able to try out the concepts I’m pulling for.

andrewmcodes:
@pjforde1978 

Ultimately Nate's decision but one thing I noticed when I ran `prettier-standard` on the project is that it changes `perform(event)` to `perform (event)`...

If we are looking for consistency...this seems to go against every piece of JavaScript I have ever read. And, as a JS novice, this was very confusing at first. Seems to go against idiomatic JavaScript imo. 

One thing I noticed is that if you pass the `--lint` flag to `prettier-standard` is that it does give you some helpful information. 

ex:

```sh
  39:19  error  'location' is not defined                          no-undef
  40:11  error  'args' is never reassigned. Use 'const' instead    prefer-const
  41:11  error  'target' is never reassigned. Use 'const' instead  prefer-const
  42:11  error  'attrs' is never reassigned. Use 'const' instead   prefer-const

✖ 4 problems (4 errors, 0 warnings)
  3 errors and 0 warnings potentially fixable with the `--fix` option.
```

It occurs to me that this discussion may be more about a request to add JS linting vs simply changing our formatter, which is the reason, I believe, why the README simply says that the use of prettier is an attempt to remove bikeshedding about formatting specifically. If adding JS linting is really the underlying discussion here - I think we should formalize around that topic, because `prettier-standard` cannot autocorrect all the lints it picks up, which makes it more than just a formatter in my mind. I know you could definitely lump standard into the same category though. 

Just my opinions. Since I am way less experienced in JS than you and Nate, I will probably remove myself from this discussion because as I said before, I don't care as long as the tool will format it for me, or at least give me an easy path to fix the issues. It seems `prettier-standard` will do this and ultimately, I am happy to use whatever you two decide on. 😄

pjforde1978:
I sincerely hope that you won't remove or censor yourself. I'm not blowing smoke when I say that I've definitely learned things from you even in just this thread. Also: I'm only a javascript expert **circa 2007**:

993113996_873df6775d_o
[l-r: the guy who made IE6, the guy who made JSON, the guy who made ARIA, me, the guy who made jQuery]

...which is specifically to say that **I am a total dinosaur**. I've been blissfully insulated from over a decade of React enthusiasts justifying why their package repository has over 750k packages when other languages only have about 200k. I thawed out and became the cranky old man shaking my fist at passing cars.

Honestly, ever little bit of adjusting to change has been painful. Nobody used to run automated lints before commiting to git because everybody used svn.

At first, I really hated everything: what the hell is webpack doing to my namespaces and why can't I see any of my classes from the console inspector? Why can't I use var again? wtaf is up with **this** and why is there now a fourth way to call functions?

Anyhow, if I have a point to make it is that we're stronger together. It's actually very likely that you'll bring interesting stuff to the table that I wouldn't even be looking for because obviously I already had a plan, right? :grinning:

hopsoft:
Will be addressed by #45

PR #45, opened 2019-09-20
Setup StimulusReflex controller callbacks

hopsoft:
# Enhancement

Inspired by #39 

Demo using this feature. https://github.com/hopsoft/stimulus_reflex_todomvc/pull/27

## Description

Adds custom controller callbacks. The hope is to support more sophisticated use cases for complex UIs.

- `beforeReflex`
- `reflexSuccess`
- `reflexError`
- `afterReflex`

```js
  connect() {
    StimulusReflex.register(this);
  }

  beforeReflex() {
    // show spinner, do stuff...
  }

  afterReflex() {
    // hide spinner, do stuff...
  }
}
```

## Checklist


pjforde1978:
All tests were done with current master stimulus_reflex_todomvc + cable_ready, dom-events branch of stimulus_reflex.

1. Tried out `stimulus-reflex:before` and it worked as expected. It's definitely a step in the right direction. The problem right now is that if you're working with an app where you have multiple instances of a controller defined, how do you know you're the instance that should respond to the event? For example, if I was calling edit, I need my callback to set focus on the correct text element... and even if I only have one text element that matches a selector, I don't need to call .focus() on it more than once.

2. `stimulus-reflex:complete` and `stimulus-reflex:success` are not working as expected - no events are captured? I don't have an easy way to test :error... interestingly, putting `raise StandardError.new "This is an exception"` on the second line of the edit method in the Todos reflex raised the error but did not interfere with the rest of the page rendering and websocket broadcasting. But if I had to make a call, I don't think `stimulus-reflex:error` is working, either.

3. Edit, cancelEdit, filtering, deleting, deleting completed, toggling all are all working. Creating, updating, toggling are generating JS errors:

Screenshot from 2019-09-20 06-21-03

4. I would still love to talk about supporting redirects, whether in stimulus_reflex or in cable_ready. Especially redirect support that checks for the presence of Turbolinks and prefers Turbolinks.visit() over window.location.

5. reflexStart and reflexComplete both seemed to trigger successfully on cancelEdit but not the initial edit. I suspect that this is because cancelEdit is the only reflex called via an action handler and the stimulate() method. This seems backed up by the toggle_all and filter functions also not triggering the callbacks. This only really sizzles if it works for declarative reflexes as well.

6. These callbacks are called for every reflex method (or at least they will hopefully soon!) which means that the developer will have to put a switch statement into the function to first figure out which method the callback is for.

As currently implemented, I don't see a lot of value over just using the event capture and interrogating the details. They just make the API bigger.

The reason I like the onEdit (and you've given me a great idea in the form of beforeEdit, too!) is that the purpose of the callback is explicit and you don't have an ugly control structure.

I believe that if we have the information needed to create the reflexStart and reflexComplete callbacks, we also have the information needed to create onEdit and beforeEdit method pairs for all of the events. [If you look at my PR](https://github.com/hopsoft/stimulus_reflex/pull/39/files#r326582522), you see that I just assemble the potential function names as strings, test to see if they are defined as functions and if they are, I execute them via the object[] notation.

For the price of a few lines of JS in the library, every dev gets to use really amazing syntax.

¯\\\_(ツ)\_/¯

FWIW, I feel like I sound harsh and negative. I'm really excited for this! I just want to get this right, and I am thorough.

hopsoft:
@pjforde1978 The last commit should somewhat address why things weren't working for you. It will at least provide enough context to help developers work around any problems similar to what you encountered.

The issue was related to the DOM selector that gets stitched together when an id isn't present. Apparently `document.querySelectorAll('[data-foo.bar="baz"]')` fails as a syntax error even though dots are legal chars in the attribute name. At any rate this is a simple enough fix and should help guide developers to a simple solution.

hopsoft:
To test the error condition, simply `raise` in one of the Ruby reflex methods.

Rakefile
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
 require "bundler/gem_tasks"
-task default: :spec
+
+task default: [:test]
+
+task :test do
```
andrewmcodes: I think I’ve mentioned this before but I’d love to help with testing. Do you think just using the rails action cable test library is enough? Interested in any ideas/guidance here.

javascript/helpers.js
```diff
@@ -0,0 +1,44 @@
...
+      if (key === 'checked') continue;
+      if (key === 'selected') continue;
+      if (!attrs.hasOwnProperty(key)) continue;
+      selectors.push(`[${key}="${attrs[key]}"]`);
+    }
+    try {
+      elements = document.querySelectorAll(selectors.join(''));
+    } catch (error) {
+      console.log(
+        'StimulusReflex encountered an error identifying the Stimulus element. Consider adding an [id] to the element.',
```
andrewmcodes:
```suggestion
        'StimulusReflex encountered an error identifying the Stimulus element. Consider adding an #id to the element.',
```

javascript/helpers.js
```diff
@@ -0,0 +1,44 @@
...
+    : null;
+
+  if (evt.stimulusController) {
+    const methods = ['reflexStart', 'reflexSuccess', 'reflexError', 'reflexComplete'];
+    methods.forEach(method => {
+      if (typeof evt.stimulusController[method] === 'function') evt.stimulusController[method](detail);
+    });
+  } else {
+    console.log(
+      'StimulusReflex was unable to identify the Stimulus controller. Consider adding an [id] to the element.'
```
andrewmcodes:
```suggestion
      'StimulusReflex was unable to identify the Stimulus controller. Consider adding an #id to the element.'
```

andrewmcodes: Would it be worth abstracting this since it’s used twice?

javascript/stimulus_reflex.js
```diff
@@ -1,7 +1,9 @@
 import ActionCable from 'actioncable';
 import CableReady from 'cable_ready';
 import StimulusReflexController from './stimulus_reflex_controller';
+import { dispatch } from './helpers';
```
andrewmcodes: Is helpers the best way we can define this? Maybe just creating a file called dispatch? Since it’s only included in one file it feels like helpers is the wrong namespace to put this but I don’t hate it. Just want it to make sense for other devs reading this source code.

javascript/stimulus_reflex.js
```diff
@@ -103,7 +108,32 @@ if (!document.stimulusReflexInitialized) {
...
+      dispatch('stimulus-reflex:success', event.detail.stimulusReflex, stimulusApplication);
+      dispatch('stimulus-reflex:complete', event.detail.stimulusReflex, stimulusApplication);
+    }
+  });
+  document.addEventListener('stimulus-reflex:500', event => {
+    dispatch('stimulus-reflex:error', event.detail.stimulusReflex, stimulusApplication);
+    dispatch('stimulus-reflex:complete', event.detail.stimulusReflex, stimulusApplication);
+  });
+
+  // document.addEventListener('stimulus-reflex:before', event => {
```
andrewmcodes: What’s the reason for including these comments? Would the be better as documentation or are they future additions?

Rakefile
hopsoft: Let's plan on a pair session sometime soon to kick off this effort.

javascript/stimulus_reflex.js
hopsoft: Valid nitpick. Will think through this a bit more.

javascript/stimulus_reflex.js
hopsoft: These are debug artifacts. I'll remove them prior to wrapping up the PR.

Rakefile
andrewmcodes: @hopsoft sounds good to me 👍

hopsoft:
Pretty sure we share similar opinions regarding how to best architect a StimulusReflex project.

stimulus_reflex_todomvc 2019-09-23 06-36-05

Having said that, I'd also stress that StimulusReflex is a lower level utility and I don't feel it should impose a particular style... even though there are "best practices" that we can and should document. This is similar to Rails itself.

---

I don't agree that more specific events like `edit` and `toggle` belong in the library. These are simple enough use cases that can be built on top of the existing lib without much effort by application developers. For example, 

1. Define a custom StimulusReflex controller
1. Do typical Stimulus things to handle user interaction(s)
1. Setup an explicit call to `stimuluate`
1. Implement `reflexBefore` and `reflexComplete`

pjforde1978:
I think that there's been a [awesome] misunderstanding.

My PR [dynamically looks up and calls those events based entirely on the name of the action](https://github.com/hopsoft/stimulus_reflex/blob/fe0cec36f9a1632c15fcaf33a7934af9165fceef/javascript/stimulus_reflex.js#L40).

The entire implementation is apparently 8 lines, depending on how you count it.

Swap out line 40... we can just take event.details.target.split('#')[1] instead.

Similarily, [swap out line 32](https://github.com/hopsoft/stimulus_reflex/blob/fe0cec36f9a1632c15fcaf33a7934af9165fceef/javascript/stimulus_reflex.js#L30) so the controller reference comes from event.details.target.split('#')[0].

No changes to the Ruby lib required.

[I think it's event.details.target, but I don't have a working build right now. I think you know what I mean, though... the string with the "TodosReflex#action" in it.]

Anyhow, at no point have I ever wanted or attempted to imply that we should hard-code any action-based event handler into the library. I passionately hope that this triggers a moment of :bulb: :zap:

hopsoft:
> I think that there's been a [awesome] misunderstanding.
>
> My PR dynamically looks up and calls those events based entirely on the name of the action.

Indeed! This is a pretty slick convention. 🤔 Will think on this some more.

javascript/stimulus_reflex.js
```diff
@@ -95,6 +92,34 @@ const invokeCallback = (name, controller) => {
   if (controller && typeof controller[name] === 'function') controller[name]();
```
pjforde1978: This invokeCallback probably should have a different name because it's useful for all sorts of things that aren't callbacks. Can we call it `send` as a homage to Ruby?

Is `send` a valid function name in JS?

Probably not super important, but technically we could change the name of the controller parameter to class (or klass) because really this is a cool pattern even if you're working in React. I always like the idea that people reading the code can learn new tricks, and this isn't specifically a Stimulus thing.

Of course, it might just be confusing, too. I'm not hung up either way.

javascript/stimulus_reflex.js
```diff
@@ -74,14 +75,10 @@ const findController = (name, element) => {
 // Finds the closest StimulusReflex controller name in the DOM tree
 const findStimulusReflexControllerName = element => {
   const controllerNames = element.dataset.controller ? element.dataset.controller.split(' ') : [];
-  let controllerName = controllerNames.reduce((memo, name) => {
+  return controllerNames.reduce((memo, name) => {
```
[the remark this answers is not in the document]
hopsoft: Could be, but I typically prefer simple single condition checks over chained conditionals... especially for early returns. It just feels more readable to me. Having said that, I'm not as consistent as I could be.

javascript/stimulus_reflex.js
```diff
@@ -225,28 +250,16 @@ if (!document.stimulusReflexInitialized) {
...
-        invokeCallback('reflexSuccess', controller);
-        invokeCallback('reflexComplete', controller);
-      }, 1);
-    });
+    const { attrs } = event.detail.stimulusReflex || {};
+    const element = findElement(attrs);
+    invokeLifecycleCallback('success', element);
+    invokeLifecycleCallback('after', element);
   });
   document.addEventListener('stimulus-reflex:500', event => {
```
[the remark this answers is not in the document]
hopsoft: Yes, I emit this from the server side as a CableReady event then trap it client side and rethrow using our canonical error event name `reflexError`

javascript/stimulus_reflex.js
```diff
@@ -1,5 +1,6 @@
 import { Controller } from 'stimulus';
 import ActionCable from 'actioncable';
+import Inflection from 'inflection';
```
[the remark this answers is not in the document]
hopsoft: This is a valid concern and one that I share. I'm moving the inflector dependency to https://github.com/martinandert/inflected which supports cherry picking the methods you're interested in. It has no dependencies and shouldn't increase the bundle size any more than simply copying the code would.

pjforde1978:
Just getting this on my phone, looking forward to trying later.

As I said: I’ve verified that I have my reflex gems and packages properly redirected.... otherwise, i would have had serious issues doing anything previously as well.

I honestly wasn’t sure what, if anything, would arrive as an argument; I see now that we’re just calling the functions straight.

The only thing that seems off is your comment about how the only event todos handles is cancelEvent... I’m using dom-events branch and defining the callbacks I want for todos actions. What am I missing? In other words, where else would they be? Or put more pointedly: this is the only place that they should be...

hopsoft:
Bear in mind that this PR is still WIP. I have more changes coming tonight.

pjforde1978:
Observations about the current builds of todomvc:lifecycle-demo + SR:dom-events.

First thing: I realized last night that the reason there's no connect() + register() in todos_controller is that it **inherits** from ApplicationController. Duh!

I put a console log in beforeReflex() in application_controller. I added a beforeReflex with a console log in the todos controller. When I initiate a reflex (edit, cancelEdit or Toggle) I see the log generated by the beforeReflex() in todos but not the one from application_controller.

If I then remove the beforeReflex() from todos, the one in application_controller fires.

If don't know if this is intended functionality. You were talking about cancelling propagation although I don't see any evidence that you've implemented this in the library or the demo. [I wrote this before I realized the part about extending ApplicationController, above. But this still leads to a solid point.]

The problem I anticipate here is this is another example of possible "cross-talk"; someone joins my team and creates a chill controller that defines a beforeReflex() and suddenly the critical code we rely on that lives in application_controller is "broken" even though they didn't change any code. The new guy has no idea that he's created a side-effect.

It seems like the beforeReflex and afterReflex callbacks should fire on all controls that define them, including application_controller. Thinking about it, I don't have a strong opinion of whether to run them on the application or individual controllers first. I'm sure that there's a right answer.

(I include the reflexSuccess and reflexError callbacks in the above as well.)

**Addendum**: I realize now that the reason beforeReflex() isn't calling twice is that Todos is redefining the definition of the beforeReflex function after inheriting from ApplicationController. However, **the larger issue remains**: what if some joker creates a controller with a beforeReflex function in it, cancelling the version that's in ApplicationController? I honestly don't know how to fix this beyond putting it in the documentation that if they define a beforeReflex in any one of their controllers, it will clobber the one in ApplicationController, likely pissing someone in the office right off.

QQ: will reflexError get access to the error message from a failed reflex on the server?

Not so QQ: should you be able to return false from a beforeReflex or beforePoop and have it cancel sending the reflex entirely?

Is there a way to inject a time delay via the beforeReflex or beforePoop callback wraps up and the reflex proceeds in sending data to the server? In Ruby we have a blocking sleep call. JS has a non-blocking async setTimeout which won't prevent the callback from returning before the timeout returns. It's actually a really thorny issue but I don't want to overthink.

I was/am a little worried that we might need/want a blacklist of methods we don't want to allow reflexes for eg. def send proc lambda method_missing although I'm currently testing an InvoiceReflex#send that seems to work just fine. I guess you can .send(:send) in this crazy world.

Anyhow, great news... I have added this to my index.html.erb:

```
<section data-controller="invoice">
  <a href="#" data-reflex="click->InvoiceReflex#send" data-room="<%= session.id %>"/>Send Invoice</a>
</section>
```

When I click it, only the invoice controller calls its' beforeSend() callback. And I don't see any callbacks for todos being called inside of invoice. I think that you've nailed the problem. **Great work.**

hopsoft:
> I put a console log in beforeReflex() in application_controller. I added a beforeReflex with a console log in the todos controller. When I initiate a reflex (edit, cancelEdit or Toggle) I see the log generated by the beforeReflex() in todos but not the one from application_controller.

OOP is only one pattern people might use. I think it's reasonable to expect that they understand the repercussions and constraints if they choose to use it. When using inheritance, the appropriate way to get the desired behavior would be to call the `super`'s method.

```javascript
// app/javascript/controllers/application_controller.js
import { Controller } from 'stimulus';
import StimulusReflex from 'stimulus_reflex';

  connect() {
    StimulusReflex.register(this);
  }

  beforeReflex() {
    document.body.classList.add('wait');
  }
}
```

```javascript
// app/javascript/controllers/todos_controller.js
import ApplicationController from './application_controller';

  beforeReflex() {
    super.beforeReflex();
    // do other things
  }
}
```

I think this is something that should be documented rather than adding mechanics to protect the developer from themselves.

---

> will reflexError get access to the error message from a failed reflex on the server?

I supported this when this PR was dispatching events. I'll update to send the error message into error based lifecycle methods like `reflexError`. The method signature will be

```javascript
reflexError(sourceElement, errorMessage) { ...
```

---

> should you be able to return false from a beforeReflex or beforePoop and have it cancel sending the reflex entirely

Possibly... I like this thinking; however, I also want consistency across all lifecycle methods. Not sure what we'd cancel if the others returned false.

---

> Is there a way to inject a time delay via the beforeReflex

I've though about this for handling things like UI animations (fade & hide on delete etc...) with a promise that invokes the reflex when the animation completes. Slowing things down to create a better perceived speed and/or improve the UX. I think we should support it (make it optional), but doubt I'll add it to this PR.

pjforde1978:
I'm a little late to the party, but I thought you guys would get a kick out of this:

https://elixirforum.com/t/phoenix-liveview-callbacks-to-put-text-input-focus-on-the-selected-input-field/21148/7

https://elixirschool.com/blog/live-view-with-channels/

Can I just say again that I love the API we came up with? It kind of warms my heart that this is impossible to do in LiveView.

PR #58, opened 2019-10-03
Documentation update

pjforde1978:
Documentation WIP

## Checklist


pjforde1978:
@hopsoft At long last, here is my attempt at radically updating the documentation.

I guarantee that you won't love everything. It absolutely requires careful review and conversation. Some of the words might just be too verbose, and others might sound like I'm drunkenly calling for React developers to come out of their McIrish Pub and fight me in the street. Yes, I did bring a crowbar.

pjforde1978:
Note that if there was a way for me to access GitBook directly, I could preview and refine without you needing to change my diapers. :baby:

docs/advanced-use.md
```diff
@@ -134,3 +138,45 @@ end
...
+
+const application = Application.start()
+const context = require.context('controllers', true, /_controller\.js$/)
+application.load(definitionsFromContext(context))
+StimulusReflex.initialize(application, ImprovedReflexController)
+```
+{% endcode-tabs-item %}
+{% endcode-tabs %}
+
+And then... stuff Nate will describe becomes possible.
```
hopsoft: I've been experimenting with this and there is an oddity with Webpack where importing a controller and passing it down causes issues. Namely that the controller class is converted to a function and the expected behavior is broken. I think I'd prefer to leave this as an undocumented feature for now as I may rip this feature out at some point.

andrewmcodes:
@pjforde1978 you can test easily by creating a gitbook account and setting up a GitHub integration for your fork and can specify the branch name to import. Lmk if I can say more.

pjforde1978:
@andrewmcodes I'm a little embarrassed that I didn't think of that.

andrewmcodes:
@pjforde1978 the beauty of collaboration! 😁

pjforde1978:
@hopsoft I noticed that after merging #61 that [we still import dasherize and underscore](https://github.com/pjforde1978/stimulus_reflex/blob/93d295c06fcc7d232079cea712a5826a5f9ffb6f/javascript/stimulus_reflex.js#L3), even though they are not referenced in the file.

pjforde1978:
@hopsoft sorry for going mildly OT but I just had a thought and this seems to be where the pre-release chat is happening.

Given that the library supports late updates via things like wait_for_it, should we consider putting a guard [here](https://github.com/pjforde1978/stimulus_reflex/blob/93d295c06fcc7d232079cea712a5826a5f9ffb6f/javascript/stimulus_reflex.js#L96) to make sure that we're **still on the right page**?

Something like:

`if (data.cableReady && data.operations.morph[0].stimulusReflex.url === location.href) {`

hopsoft:
Good call on the guard. I'll add it.

PR #67, opened 2019-10-10
add generators

### LINKED ISSUE #3: Create Rails generators

andrewmcodes:
# Enhancement

## Description

```bash
bundle exec rails stimulus_reflex:install
```

This command will generate:

1. `app/javascript/controllers/application_controller.js`
2. `app/javascript/controllers/example_controller.js`
3. `app/reflexes/application_reflex.rb`
4. `app/javascript/controllers/example_reflex.rb`

```bash
bundle exec rails generate stimulus_reflex my_demo
```

This command will generate:

1. `app/javascript/controllers/application_controller.js` (If the file does not already exist)
2. `app/javascript/controllers/my_demo_controller.js`
3. `app/reflexes/application_reflex.rb` (If the file does not already exist)
4. `app/reflexes/my_demo_reflex.rb`

```bash
bundle exec rails generate stimulus_reflex:controller my_demo
```

This command will generate:

1. `app/javascript/controllers/application_controller.js` (If the file does not already exist)
2. `app/javascript/controllers/my_demo_controller.js`

```bash
bundle exec rails generate stimulus_reflex:reflex my_demo
```

This command will generate:

1. `app/reflexes/application_reflex.rb` (If the file does not already exist)
2. `app/reflexes/my_demo_reflex.rb`

Fixes #3

## Why should this be added

Help developers get up and running with some sample files

## Checklist


Rakefile
```diff
@@ -5,5 +5,6 @@ require "bundler/gem_tasks"
 task default: [:test]
 
 task :test do
-  exec "cd javascript && yarn run test"
+  system "bin/ruby_test"
+  system "cd javascript && yarn run test"
```
hopsoft: What does system return? Will it exit with a non zero code if the command fails?

andrewmcodes: @hopsoft 
```
➜ rake test
Run options: --seed 23909

# Running:

F

Failure:
StimulusReflexGeneratorTest#test_creates_named_controller_and_reflex_files [/Users/andrew.mason/work/OSS/stimulus_reflex/test/generators/stimulus_reflex_generator_test.rb:14]:
Expected file "app/reflexes/test_reflex.rb" to not exist, but does


bin/test Users/andrew.mason/work/OSS/stimulus_reflex/test/generators/stimulus_reflex_generator_test.rb:8

No value provided for required arguments 'name'
..No value provided for required arguments 'name'
..No value provided for required arguments 'name'
.

Finished in 0.014597s, 411.0434 runs/s, 1233.1301 assertions/s.
6 runs, 18 assertions, 1 failures, 0 errors, 0 skips
yarn run v1.19.0
$ yarn run mocha --require @babel/register
$ /Users/andrew.mason/work/OSS/stimulus_reflex/javascript/node_modules/.bin/mocha --require @babel/register


  attributeValue
    ✓ returns falsey for no arguments
    ✓ returns falsey for empty array
    ✓ returns expected attribute value for array with single value
    ✓ returns expected attribute value for array with multiple values
    ✓ returns expected attribute value for array with multiple values that include whitespace
    ✓ returns expected attribute value for array with mixed values

  attributeValues
    ✓ returns empty array for no arguments
    ✓ returns expected attribute values for single value
    ✓ returns expected attribute values for multiple values
    ✓ returns expected attribute values for multiple values with whitespace padding

  extractElementAttributes
    ✓ returns expected attributes for empty anchor
    ✓ returns expected attributes for anchor
    ✓ returns expected attributes for textarea
    ✓ returns expected attributes for textbox
    ✓ returns expected attributes for unchecked checkbox
    ✓ returns expected attributes for checked checkbox

  matchingControllerName
    ✓ returns expected controller name for reflex
    ✓ returns expected controller name abbreviated for event + reflex


  18 passing (40ms)

✨  Done in 0.91s.```

andrewmcodes:
I think there is now a way to share your action cache amongst other actions - will need to look into that.

pjforde1978:
Hey, @andrewmcodes - this looks like really good work. You asked for some eyes and I want to make it clear that I've never made a generator script and perhaps, more importantly, I don't frequently use them except when testing libraries that are new to me. Seeing as that's **exactly** what you're trying to do, I am grateful for your efforts.

Perhaps my only concern is that in some ways, this might be too flexible. I count four primary options, and I wonder if we should make the outcome most devs will want 99% of the time the default?

For `rails stimulus_reflex:install`... I've always thought that install was associated with installing packages, adding routes, and running migrations. Could the functionality here be merged with the generate command itself? It seems like at least we should verify that both Stimulus and StimulusReflex are installed as part of this step if we keep it.

It does seem like you could ditch the `rails stimulus_reflex:install` in favor of using the `rails generate stimulus_reflex` without a parameter. If they don't specify a name for their SR controller, you could default them to `example` or just bail on the task with an exception that says you have to provide a name.

Finally, I have a hard time imagining a developer scenario in which they would not just generate both the Ruby and JavaScript files. Providing the :controller and :reflex options is like trying to improve your restaurant profitability by adding more items to an already busy menu.

This really is going to help less experienced developers get up to speed quickly. I just hope that in the end, we can focus on providing one obvious option that works for 99% of that demographic.

PR #69, opened 2019-10-11
Tweak generators

hopsoft:


hopsoft:
@andrewmcodes I updated to run Ruby tests with Rake. Can you update the GH Ruby Test Action to work with this? https://github.com/hopsoft/stimulus_reflex/blob/hopsoft/generator-tweaks/Rakefile#L13-L15

PR #70, opened 2019-10-14
Prevent redundant `after` lifecycle callbacks

### LINKED ISSUE #68: too many afterReflex/reflexSuccess callbacks
# Bug Report

## Describe the bug

If a reflex is scoped to update multiple selectors, the reflexSuccess and afterReflex callbacks are being executed once per morph operation instead of once per cable_ready payload.

## To Reproduce

Ensure afterReflex writes to the console log. Call a reflex that is scoped to two or more selectors.

## Expected behavior

Each reflex action should fire callbacks one time.

### StimulusReflex

- Gem: 2.1.2

### External tools

- Ruby: 2.6.4
- Rails: 6.0.0

### Browser

- Chrome 77

hopsoft:
Resolves #68


pjforde1978:
Yeah... this works!

PR #74, opened 2019-10-17
Github url

benbonnet:
### Type of PR

bug fix

### Description

Wrong link in to the demo in the documentation

### Why should this be added

The previous linked repo was not providing useful examples

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -9,7 +9,7 @@ description: How to use StimulusReflex in your app
 A great user experience can be created with Rails alone. Tools like [UJS remote elements](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#remote-elements), [Stimulus](https://stimulusjs.org/), and [Turbolinks](https://github.com/turbolinks/turbolinks) are incredibly powerful when combined. Try building your application using these tools before introducing StimulusReflex.
 
 {% hint style="info" %}
-See the [Stimulus TodoMVC](https://github.com/hopsoft/stimulus_todomvc) example application if you are unsure how to do this.
+See the [Stimulus TodoMVC](https://github.com/hopsoft/stimulus_reflex_todomvc) example application if you are unsure how to do this.
 {% endhint %}
 
 ## Hello, Reflex


benbonnet:
Thx for pointing this out. I might have misunderstood the context and proposed the fix too quickly; I'll take more time to get a better understanding of it all first

pjforde1978:
Here's the thing: you might be guilty of not reading carefully, but we're also guilty of violating the principle of least surprise. We were going for a pragmatic PSA, but perhaps starting the "quick start" with an admonishment to possibly not use the library is the right idea at the wrong time.

@hopsoft and co will discuss!

pjforde1978:
@benbonnet would appreciate eyes on https://docs.stimulusreflex.com/quickstart

benbonnet:
@pjforde1978 because, on a first approach, people might read briefly through the docs and try that in a 10/15 minutes break in between other tasks during the day (at least my case), I'd link clearly the two :
```
- Classic rails (current link)
- Reflex way (other repo with reflex sample)
```

Could be interesting to add the link here https://docs.stimulusreflex.com/reflexes ? Or maybe the "next steps" link on the readme higher at the top of the page

---

About the adonishment bad idea, I kind of tend to agree. People who might be hyped by this project :
- recognize the elixir community as a very strong one (no `DHH® approved` requirement)
- understand fully what is an spa project (want to get rid of such boilerplate/pain)
- know (a lot) about rails (no extensive comparison needed)

Exagerating the idea a bit, but the only focus (at least relevant to me) is how it helps to make the js logics disappear/be moved serverside

---

Many others would already be fluent in react/vue rather than in stimulus. Apart from the title and what's installed, it does not appears as a strong requirement. Plus, if I've understood it well, [you can achieve the same in a serverside way](https://docs.stimulusreflex.com/patterns#server-side). Should one be fluent with stimulus to correctly use this gem ?

benbonnet:
not dissing anything or anyone, neither dhh; let's say i meant that a rails dev who heard of elixir and heard about liveview, will have a high potential to be onboarded.

I'll surely give a look back again at stimulus.

about my question; i'll be more accurate : [this](https://docs.stimulusreflex.com/patterns#spinners-for-long-running-actions) is very close to [this](https://docs.stimulusreflex.com/patterns#chained-reflexes-for-long-running-actions) in terms of ux, am i right to think that ?

pjforde1978:
I think I understand, finally. :grin:

The spinners snippet that you linked to is an example of something you might want to setup client-side callback handlers for. I went with spinners as an example of a global callback because it's literally something that could be useful for every server call.

Chained reflexes - which I've implemented as the wait_for_it function is a Reflex meta-programming trick that allows you to link a theoretically unlimited number of reflex updates together. Each reflex culminates in an ActionCable message with one or many morphdom updates (depending on whether you use scoping or not!).

Without something like wait_for_it, you could click a button on the client and wait for a potentially really long time for some remote API to reply, if indeed it ever does. By spinning up a 2nd thread, we're able to immediately reply to the request with an intermediate update - think "Loading..." - and then follow it up with a 2nd update when/if the data arrives.

It's not formally part of the library, although it's a strong contender. We want to make sure it tests well from a profiling perspective. But I've used it a lot so far and the sky has not fallen.

Anyhow, I hope you understand a bit better now. wait_for_it is a really exciting thing, I believe. It's a quiet killer feature. It is, however, definitely very different from client-side callbacks, which are also cool - even though we hope people won't have to use them often. We don't hate JS, we just want people to feel like they don't have to write a ton of it to get impressive functionality working.

benbonnet:
`we just want people to feel like they don't have to write a ton of it to get impressive functionality working` yep the simple fact of being able to manipulate the frontend in reflexes (in ruby) sounds highly appealing

pjforde1978:
Just to hammer home the whole "infinite" thing, here's a stopwatch:

```erb
<section data-controller="counter">
  <h2><%= Rails.cache.fetch("timer:#{session.id}") {0} %></h2>
  <button data-reflex="click->CounterReflex#stopwatch"><%= session[:clock] ? "Stop" : "Start" %></button>
</section>
```
```ruby
class CounterReflex < StimulusReflex::Reflex
  def stopwatch
    session[:clock] = session[:clock] ? !session[:clock] : true
    wait_for_it(:tick) do
      []
    end
  end

  def tick
    if session[:clock]
      Rails.cache.increment("timer:#{session.id}")
      wait_for_it(:tick) do
        sleep 1
        []
      end
    end
  end

  def wait_for_it(target)
    return unless respond_to? target
    if block_given?
      Thread.new do
        channel.receive({
          "target" => "#{self.class}##{target}",
          "args" => yield,
          "url" => url,
          "attrs" => element.attributes.to_h,
          "selectors" => selectors,
        })
      end
    end
  end
end

pjforde1978:
> `we just want people to feel like they don't have to write a ton of it to get impressive functionality working` yep the simple fact of being able to manipulate the frontend in reflexes (in ruby) sounds highly appealing

The way we hope you think about it - and tell your friends - is that for the past 20ish years, 80% of the hardest part of our jobs has been managing and syncing client state with the server. If StimulusReflex means you can design a stateful app that has no client state, then we're eliminating the 80% of work.

hopsoft:
@benbonnet I agree with the points you've made on this PR. Especially this one. 

> people might read briefly through the docs and try that in a 10/15 minutes break in between other tasks during the day

Given that the principal ethos of this project is **saving developer time** our docs should reflect it. I will amend the section you called out soon so we can help developers see the important things faster.

@pjforde1978's points above are all accurate, but I think we can do a few things better in the docs.

1. Outline the ideal person for StimulusReflex. *Namely a proficient Rails developer that's perhaps frustrated with the current state of affairs with SPA being the mainstay of conventional wisdom.*
2. Briefly state that the **attitude of questioning** whether SPAs should be the defacto solution *for your project/team* should be applied to StimulusReflex also... because Rails is very capable without it.
3. Ensure all docs are focused on clearly explaining the benefits of using StimulusReflex.

PR #76, opened 2019-10-19
Single source of truth

pjforde1978:

Enhancement

## Description

Proposed fix for the problem of multiple users clobbering each other by typing into the same input at the same time. Concept is that if a textual input element has focus, morphdom should never overwrite the input's current value even when it deviates from the server's rendered updated.

## Why should this be added

While not a complete solve for multi-user concurrency, this imposes some sanity on an otherwise imperfect status quo.

## Checklist

### DIFF
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -84,7 +84,7 @@ export const findElement = attributes => {
   return element
 }
 
-export const isTextInput = element => {
+const isTextInput = element => {
   return (
     [
       'email',
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -86,7 +86,7 @@ export const findElement = attributes => {
 
 const isTextInput = element => {
   return (
-    ['input', 'textarea'].indexOf(element.tagName) >= 0 &&
+    ['INPUT', 'TEXTAREA', 'SELECT'].indexOf(element.tagName) >= 0 &&
     [
       'color',
       'date',
@@ -98,6 +98,8 @@ const isTextInput = element => {
       'password',
       'range',
       'search',
+      'select-one',
+      'select-multiple',
       'tel',
       'text',
       'textarea',
@@ -118,7 +120,7 @@ export const receivedFocus = event => {
 export const lostFocus = event => {
   const element = event.target
   if (!isTextInput(element)) return
-  if (element.reflexPermanent && !element.reflexPermanent) {
+  if (element.reflexPermanent !== undefined && !element.reflexPermanent) {
     element.removeAttribute('data-reflex-permanent')
   }
 }
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -83,3 +83,33 @@ export const findElement = attributes => {
   const element = elements.length === 1 ? elements[0] : null
   return element
 }
+
+export const isTextInput = element => {
+  return (
+    [
+      'email',
+      'number',
+      'password',
+      'search',
+      'tel',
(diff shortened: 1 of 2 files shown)


hopsoft:
See #77 as it's related. There was a bug that assumed that connection identifiers are always Rails model instances _or at least object with an `id`_.

pjforde1978:
Great news that #77 was patched, although technically this PR was coming regardless. The server should not be able to overwrite the contents of an input element while you're typing into it.

javascript/attributes.js
```diff
@@ -83,3 +83,33 @@ export const findElement = attributes => {
...
+      'number',
+      'password',
+      'search',
+      'tel',
+      'text',
+      'textarea',
+      'url'
+    ].indexOf(element.type) >= 0
+  )
+}
```
hopsoft: I might add an additional check for `element.tagName`

pjforde1978: Interesting. I thought they were basically the same. Hard to Google. Can you explain the difference in your own

hopsoft:
```javascript
document.getElementsByTagName("INPUT")[0].tagName
"INPUT"

document.getElementsByTagName("INPUT")[0].type
"hidden"
```

This feature should ignore all tags but `input` and `textarea`. Then only apply the logic to the types you've already outlined.

javascript/attributes.js
```diff
@@ -83,3 +83,33 @@ export const findElement = attributes => {
...
+  const element = event.target
+  if (!isTextInput(element)) return
+  element.reflexPermanent = element.hasAttribute('data-reflex-permanent')
+  element.setAttribute('data-reflex-permanent', '')
+}
+
+export const lostFocus = event => {
+  const element = event.target
+  if (!isTextInput(element)) return
+  if (element.reflexPermanent !== undefined && !element.reflexPermanent) {
```
hopsoft: I'd change this guard to:

```javascript
if (element.reflexPermanent) return
```

hopsoft:
I'm currently building a chat demo on the expo project and am seeing a usecase where we actually do want the input clobbered when the reflex updates the DOM. 🤔

pjforde1978:
https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#module-custom-dom-patching ???

pjforde1978:
Just wanted to check in on this and see where your head is at. I'm happy to make the changes you suggested, but I wanted to make sure that we're still on the same page about it being a good idea.

hopsoft:
I still think this is a good idea. If someone want's to take action on one of these input fields (as my chat example above) they can still do so via a lifecycle method.

javascript/stimulus_reflex.js
```diff
@@ -285,6 +292,8 @@ if (!document.stimulusReflexInitialized) {
     invokeLifecycleMethod('error', target, element)
     invokeLifecycleMethod('after', target, element)
   })
+  document.addEventListener('focus', receivedFocus, true)
```
pjforde1978: Only just learned about [focusin](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusin_event) from your PR. I'm going to commit a patch here that switches over...

pjforde1978: Love all of the cool things I learn from @hopsoft :heart_eyes_cat:

pjforde1978:
Oh, awesome! Seemed like you were on the fence.

PR #80, opened 2019-10-21
abort reflex action if method returns false

leastbad:

Enhancement

## Description

Allow developer to abort a reflex operation before rendering the template if the Reflex class method returns false. No ActionCable broadcast will be send, and no afterX callbacks will be fired.

## Why should this be added

Flexibility!

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -25,7 +25,7 @@ let application
 // - error
 // - after
 //
-const invokeLifecycleMethod = (stage, reflex, element) => {
+const invokeLifecycleMethod = (stage, reflex, element, returnValue) => {
   if (!element) return
 
   // traverse the DOM for a matching reflex controller
@@ -57,6 +57,7 @@ const invokeLifecycleMethod = (stage, reflex, element) => {
             controller,
             element,
             reflex,
+            returnValue,
             element.reflexError
           ),
         1
@@ -70,6 +71,7 @@ const invokeLifecycleMethod = (stage, reflex, element) => {
             controller,
             element,
             reflex,
+            returnValue,
             element.reflexError
           ),
         1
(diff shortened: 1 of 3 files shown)


pjforde1978:
I know that this is coming out of nowhere, but I wanted to fully explore this idea's potential so that we could discuss.

After I implemented the initial abort functionality, I realized that it's silly to discard the return value of the function and that it could, in fact, be an excellent way to pass data back to the client callback so that the developer can respond to whatever comes back without having to resort to ugly hacks with instance variables and DOM attributes to shuttle the data around.

Right now there's no protection against someone returning a complex object that won't serialize well. Is there an equivalent in Ruby to the shallow clone schemes you see in JS? What is `JSON.parse(JSON.stringify(obj))` in Ruby?

pjforde1978:
One thing about my 2nd commit is that if we're willing to entertain the notion that this is a good (great) idea, then possibly taking false off the table as a potential return value is a dumb move.

I've introduced a `cancel_reflex!` instance method that can be called from a reflex method.

pjforde1978:
Hey @hopsoft, if one was to theoretically call CableReady::Channel.broadcast with no operations defined, it's fair to say that ActionCable would broadcast an envelope with no operations in it, right?

I was just thinking that based on all of the above, it would be cool if someone could return a value from an aborted reflex. That would allow us to send a message to the client that could be picked up by the after_Event_ callback, possibly with an additional `cancelled: true` flag set. In this way, the dev could handle non-error, cancelled reflexes programmatically.

I am working on a pretty cool example that will illustrate why this is all really powerful, but I'm keeping mum until I'm ready for feedback.

hopsoft:
> if one was to theoretically call CableReady::Channel.broadcast with no operations defined, it's fair to say that ActionCable would broadcast an envelope with no operations in it, right?

Yes

hopsoft:
Part of me thinks that this type of behavior would be better served by something like a `before_action` rather than within the internals of a reflex method.

pjforde1978:
Could you go into more detail about what you mean? It's not clear where a before_action-like construct would address returning a value from the reflex action to the client. (If you mean just for aborting an action, then yes, that is certainly another possible approach. I'm still partial to the approach I am proposing but I'm also excited to see how you would structure it differently.)

This PR definitely started as an abort mechanism, but by the time I made my last commit, I was equally as excited about the return values, which I believe could be extraordinarily useful when building afterReflex callbacks.

leastbad:
Finally, some closure thanks to the excellent work happening in https://github.com/seb1441/stimulus_reflex/pull/1

PR #86, opened 2019-10-27
add actioncable connection status classes to document.body

leastbad:

Enhancement

## Description

ActionCable provides subscription callbacks for connected and disconnected states. [Borrowed a good idea from LiveView](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#module-loading-state-and-errors) to map these states onto the document body as css classes which can be used to hide/reveal functionality that shouldn't be available unless the connection has been established. 

## Why should this be added

If you look at [this nasty insanity](https://github.com/hopsoft/stimulus_reflex_expo/blob/a4920b84460846936a0d19f5cba10c45257cf399/app/javascript/controllers/snake_controller.js#L7) you will see an excellent example of the code currently required to wait for the socket to connect. In fact, 40% of this controller exists to support this function, with [additional complexity in the view](https://github.com/hopsoft/stimulus_reflex_expo/blob/a4920b84460846936a0d19f5cba10c45257cf399/app/views/snakes/_demo.html.erb#L29).

All of this could be replaced by CSS selectors on the key elements that show and hide themselves depending on the classes present in the body element, which is a standard trick in every front-end developer's toolkit.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -85,12 +85,21 @@ const invokeLifecycleMethod = (stage, reflex, element) => {
 const createSubscription = controller => {
   const { channel, room } = controller.StimulusReflex
   const id = `${channel}${room}`
-
   const subscription =
     app.StimulusReflex.subscriptions[id] ||
     app.StimulusReflex.consumer.subscriptions.create(
       { channel, room },
       {
+        connected: () =>
+          document.body.classList.replace(
+            'reflex-disconnected',
+            'reflex-connected'
+          ),
+        disconnected: () =>
+          document.body.classList.replace(
+            'reflex-connected',
+            'reflex-disconnected'
+          ),
         received: data => {
           if (!data.cableReady) return
           const urls = [
@@ -261,6 +270,7 @@ app.StimulusReflex.subscriptions = app.StimulusReflex.subscriptions || {}
 
 if (!document.stimulusReflexInitialized) {
   document.stimulusReflexInitialized = true
+  setTimeout(() => document.body.classList.add('reflex-disconnected'), 1)
   window.addEventListener('load', () => setTimeout(setupDeclarativeReflexes, 1))
   document.addEventListener('turbolinks:load', () =>
     setTimeout(setupDeclarativeReflexes, 1)


javascript/stimulus_reflex.js
```diff
@@ -85,12 +85,21 @@ const invokeLifecycleMethod = (stage, reflex, element) => {
...
+        connected: () =>
+          document.body.classList.replace(
+            'reflex-disconnected',
+            'reflex-connected'
+          ),
+        disconnected: () =>
+          document.body.classList.replace(
+            'reflex-connected',
+            'reflex-disconnected'
+          ),
```
hopsoft: Note that each SR controller gets it's own subscription, so we'll need to move the css class to the respective controller's element. It can be used in the same manner that you proposed for the body class.

Also, I think we can get away with a single class and simplify this further.

```javascript
connected: () => controller.element.classList.add('reflex-connected'),
disconnected: () => controller.element.classList.remove('reflex-connected'),
```

pjforde1978: I originally got confused on this, too. However, this code path only executes once because of the clever way you evaluate it.

I'm 100% confident that this is supposed to be on the body element, however you want to select it.

pjforde1978: FWIW, I like having classes for both. That way you don't have to test for the absence of, which is kind of ugly from a selector perspective.

hopsoft: Ah yes you're correct about body. I still propose that we change this to be a single `connected` css class.

pjforde1978: I've explained why I proposed two separate classes. Could you explain why you want to push for one? You mentioned simplification but I don't think that's what this would accomplish. There is the `:not()` selector but `body.reflex-disconnected` is far easier to work with.

Essentially, I am suggesting that the option to craft cascades based on the potential for both classes has more flexibility than `:not()` and as a hack front-end dev I'm probably not the best avatar to defend the position, which sucks, because I think I'm right.

hopsoft: My thinking is more along the lines of mobile first responsive design. Set the default styles to be non reflex connected. i.e. `hidden`, `disabled`, etc... Then add custom css for when the reflex is connected. This feels simpler to my mind.

hopsoft: Essentially this philosophy applied to SR _(where disconnected is default)_.

> we present the mobile layout and functionality as the default configuration provided, before additional information is loaded on top of that

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Responsive/Mobile_first

hopsoft: What originally triggered my thinking on this was the use of `classList.replace` which will only work if the `disconnected` class has already been added to the `body` tag.

pjforde1978: This is true, which is technically why I did [this](https://github.com/pjforde1978/stimulus_reflex/blob/772e27df9d360d082a402f6b8c6a6e7d93348603/javascript/stimulus_reflex.js#L273) - **it starts off in the reflex-disconnected state**.

Here's how I set up my snake example to take advantage of the PR:

```Scss
body.reflex-connected .disconnected {
  display: none;
}
body.reflex-disconnected .connected {
  display: none;
}
```
```Erb
<h3 class="text-info disconnected">Connecting...</h3>
<button class="btn btn-primary connected">Start</button>
```

The idea, of course, is that you want to show "Connecting" when disconnected, and the "Start" button when connected. What's not to love?

To do it "your" way :grinning: looks like this:

```Scss
body.reflex-connected .disconnected {
  display: none;
}
body:not(.reflex-connected) .connected {
  display: none;
}
```

If you honestly find that more clear, then we simply disagree and I'll follow your lead without further grief because ultimately, you have to be the decider.

pjforde1978: Either way, it's worth celebrating that this will take an additional 9 LOC off the JS controller.

hopsoft: If we invert my proposal above and make the default state `connected`, the CSS would change to this.

```css
.disconnected { display: none; }
body.reflex-disconnected .disconnected { display: auto; }
body.reflex-disconnected .connected { display: none; }
```

```html
<div class="disconnected">Connecting</div>
<div class="connected">Connected</div>
```

hopsoft:
Another thing to consider is that SR sets up a unique subscription per room. This makes it so a single global `connected` for the library doesn't quite work as implemented. Consider the following.

```html
<a data-reflex="FirstReflex#perform">
<a data-reflex="SecondReflex#perform">
<a data-reflex="ThirdReflex#perform" data-room="special">
```

The first 2 anchors share a subscription, but the third uses a different subscription.

We'll need to factor in all the SR ActionCable subscriptions before setting the global `body.reflex-*` connection CSS classes.

javascript/stimulus_reflex.js
pjforde1978: Yeah, that’s acceptable. Not the hill I die on, either. I really will support whatever you run with.

pjforde1978:
This is true; I’m going to propose we pretend like this isn’t a big deal and just let it set the value to the same thing n times because namespacing room connections in css is awful.

I cannot think of a realistic situation where only one room would be disconnected. Can you?

leastbad:
A variation on this functionality has been added with #148. 🥳

PR #88, opened 2019-10-30
Various documentation fixes

BakiVernes:
# Enhancement

## Description

Fixed some errors in the documentation.


pjforde1978:
Thank you so much for spotting all of these! I don't even begin to know how you picked those out.

Is there any way we can help you get started with the library? We are here for you.

BakiVernes:
I have already tried it in a Hello World type app without any problems. Also, I have tried using it in a complex production app with some minor hiccups which are app specific. I was able to resolve the problems. I'll get back to you if I find any repeating problematic patterns so we add warnings to new users in the docs or we can discuss it further for any enhancements. Thanks for the quick response!

PR #89, opened 2019-10-30
specify data-reflex-allow on an input element to allow reflex operations

leastbad:

Enhancement

## Description

Adds a `data-reflex-allow` attribute to the API, which can be put on an input element to opt-out of SSoT handling.

## Why should this be added

While I still strongly believe SSoT is the responsible default, scenarios do emerge when you need an escape hatch.

## Checklist


leastbad:
Superceded by #157

PR #92, opened 2019-10-31
inherit stimulus schema

nickyvanurk:
# Enhancement

## Description

Allows for custom schema attributes.


nickyvanurk:
I am not sure if `application.schema.reflexAttribute` would cause other problems in the Reflex codebase, it probably does. However the `application.schema.actionAttribute` should cause no issues.

PR #94, opened 2019-11-03
Add schema support

### LINKED ISSUE #91: Custom Stimulus schema breaks Reflex
## Describe the bug

Reflex ignores the schema set in Stimulus. It assumes Stimulus will always use the default schema.

Line 220 in javascript/stimulus_reflex.js
```
 if (actionValue) element.setAttribute('data-action', actionValue)
```
Hardcodes the 'data-action' attribute.

In our project we use a custom schema as follows:

```
const application = Application.start(document.documentElement, {
  controllerAttribute: 'data-controller',
  actionAttribute: 'data-do',
  targetAttribute: 'data-target'
});
```
Because we had a conflict with another plugin, cropperjs.

Feedback would be appreciated.

hopsoft:
This effort was started with the work on #92 

Resolves #91


pjforde1978:
Huzzah!

PR #99, opened 2019-11-14
Docs typo

szTheory:
Docs typo


andrewmcodes:
@pjforde1978 You sure? This change makes sense to me grammatically but if you disagree maybe we could sub this word out for something else? 

`...we've traded fun and productive for...`

It's definitely clunky if its intentional.

pjforde1978:
I’m not a snoot, but I’m pretty sure that if you imagine quotes around “fun and productive” you’ll hear the voice Nate was shooting for.

That said, I’m most confident that if you can think of something better quickly, you should just change it because this is not worth the time we’re spending thinking about it.

andrewmcodes:
Reading it in @hopsoft’s voice definitely helps but I still am inclined to accept this change since not everyone has gotten the same opportunity. 

@hopsoft can we get a deciding vote here?

hopsoft:
Thank you for the PR. @pjforde1978 has the correct interpretation here, but I do want the docs to be as inclusive and understandable as possible. I'll work on rephrasing this into something less prone to misunderstanding.

PR #102, opened 2019-11-29
Squish HTML prior to sending over wire

hopsoft:
# Enhancement

A minor performance enhancement that sends fewer bits over the wire.

### DIFF
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -63,7 +63,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
   end
 
   def render_page_and_broadcast_morph(url, reflex, selectors, data = {})
-    html = render_page(url, reflex)
+    html = render_page(url, reflex)&.squish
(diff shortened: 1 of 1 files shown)


hopsoft:
I'm trying to think through use cases where you might not want `squish` enabled. The only one I'm coming up with is for `<pre>` and `<code>` tags. Let me know if anyone can think of other scenarios.

pjforde1978:
What you said; those seem to be blockers, no?

The bigger thing (not a blocker) is evaluating something like Jose Valim’s Leex in the context of Rails. I briefly investigated trying my hand at it, creating a sort of managed subset of erb with some more constraints around control flow, but frankly it was hard so i allowed myself to get sucked into other things.

I still think it would be pretty amazing to get an intercalated array of static/dynamic changed bytes happening.

hopsoft:
Not necessarily a blocker. We could scan for such tags and omit squish for those but the check could mitigate the benefits of this PR. Another option could be to explicitly opt in to the squish behavior per reflex.

PR #111, opened 2020-01-26
Fix bug related to trailing slash in URL path

hopsoft:
# Fix

## Description

A trailing slash in the URL's path was raising an error and preventing a rerender after reflex invocation.

## Why should this be added

This PR fixes the above problem.

## Checklist


hopsoft:
I dug into Rails source to see the implementation of `Rails.application.routes.recognize_path_with_request` and `Rails.application.routes.recognize_path` and realized that Rails doesn't normalize the path for `recognize_path_with_request`. Not sure if that's an oversight or if it's intentional.

PR #116, opened 2020-02-26
Explicit and implicit registering of the ActionCable consumer

hopsoft:
## Description

Support both explicit and implicit registering of the ActionCable consumer.

## Why should this be added

This will save resources by ensuring that only 1 web socket connection is established.

See hopsoft/stimulus_reflex_expo/pull/24 for reference usage. 

## Checklist


javascript/package.json
```diff
@@ -13,7 +13,7 @@
     "prettier-standard-check": "yarn run prettier-standard --check ./**/*.js"
   },
   "dependencies": {
-    "actioncable": ">= 5.2",
+    "@rails/actioncable": ">= 6.0",
```
andrewmcodes: Does this make it so this version relies on this package? 

Aka does this remove support for `"actioncable": ">= 5.2"`

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
andrewmcodes: Why is this empty? Is there any reason to include it since it's empty?

hopsoft: Artifact. I need to remove.

javascript/package.json
hopsoft: Yes.

andrewmcodes: Major version bump then? @hopsoft

hopsoft: Maybe. Not sure yet. The API hasn't changed enough to warrant this, but that dependency bump may warrant another major. Though I'm not really a fan of doing that on this one. 

Maybe I can switch actioncable back to the previous version. I'll try that first and see what happens.

javascript/consumer.js
```diff
@@ -0,0 +1,28 @@
...
+function isConsumer (object) {
+  if (object) {
+    try {
+      return (
+        object.constructor.name === 'Consumer' &&
+        object.connect &&
+        object.disconnect &&
+        object.send
+      )
+    } catch {}
```
leastbad: I know that ESsomething doesn't require catch to take an arg, but doesn't that mean this won't work unless ESsomething is present? I think you should just pass the arg.

javascript/stimulus_reflex.js
```diff
@@ -14,17 +14,21 @@ import {
 
 // A reference to the Stimulus application registered with: StimulusReflex.initialize
 //
-let app
+let stimulusApplication
```
leastbad: Are you renaming this for clarity, or is there a functional change?

javascript/stimulus_reflex.js
```diff
@@ -42,12 +46,13 @@ const resetImplicitReflexPermanent = event => {
 // controller - the StimulusReflex controller to subscribe
 //
 const createSubscription = controller => {
+  actionCableConsumer = actionCableConsumer || getConsumer()
   const { channel, room } = controller.StimulusReflex
-  const id = `${channel}${room}`
+  const identifier = JSON.stringify({ channel, room })
```
leastbad: We should do up a PR to remove room from the client and server. Nobody has mentioned it except as something that was confusing since we removed it from the docs.

javascript/stimulus_reflex.js
hopsoft: Both clarity and consistency given that we're now also tracking `actionCableConsumer`

PR #120, opened 2020-02-28
authentication.md: Suggest storing only the session ID string

henrik:
At least in Rails 6, `session.id` is an instance of `Rack::Session::SessionId`.

Without the `to_s`, `cookies.encrypted[:session_id]` will end up storing a hash like `{"public_id"=>"abc123"}`, which still seems to work, but seems unnecessary.


henrik:
Thanks!

leastbad:
Nice catch, @henrik!

PR #130, opened 2020-03-25
Fix docs spring binstub command

seb1441:
Doc.

## Description

Fixed the last command which removes spring binstubs in the setup page.


leastbad:
Nice catch!

PR #140, opened 2020-04-08
independently test for both morphs and dispatched events

### LINKED ISSUE #139: Non-morph operations are not executed by CableReady on errors
The event listener for `stimulus-reflex:500` will short-circuit when it receives a broadcast that has no morph operations. 

https://github.com/hopsoft/stimulus_reflex/blob/bd4ceb22df5fa18d3a1001afa7ffc25380ee14b7/javascript/stimulus_reflex.js#L58

500s do not return any morph operations, but you would still want CableReady to run the `dispatchEvent` operation to the error can be trapped. 

## To Reproduce

Cause an error to be raised by the reflex and attempt to trap it in an error event.

## Expected behavior

The error should cause a `stimulus-reflex:500` error event to be dispatched.

### StimulusReflex

- Gem: 3.0.0
- Node package: 3.0.0

### External tools

- Ruby: 2.6.5
- Rails: 6.0.2.2
- Node: 10.19.0

### Browser

- Browser Chrome
- Version 80.0.3987.163

leastbad:
# Bug Fix

## Description

SR client currently returns before emitting stimulus-reflex:500 event. This patch breaks out individual CableReady operation payloads and executes them independently.

Fixes #139 

## Why should this be added

Errors are not currently being raised.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -55,12 +55,15 @@ const createSubscription = controller => {
     actionCableConsumer.subscriptions.create(channel, {
       received: data => {
         if (!data.cableReady) return
-        if (!data.operations.morph || !data.operations.morph.length) return
-        const urls = [
-          ...new Set(data.operations.morph.map(m => m.stimulusReflex.url))
-        ]
-        if (urls.length !== 1 || urls[0] !== location.href) return
-        CableReady.perform(data.operations)
+        if (data.operations.dispatchEvent && data.operations.dispatchEvent.length)
+          CableReady.perform({ dispatchEvent: data.operations.dispatchEvent })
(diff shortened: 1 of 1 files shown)


hopsoft:
Is there a reason we should block all CableReady operations on this channel except for `morph` and `dispatchEvent`? I don't recall why we added the explicit return for non morphs in the first place.

leastbad:
https://discordapp.com/channels/629472241427415060/629472241427415062/698963235629432922

hopsoft:
Did some digging and realized that there wasn't a good reason to return early without allowing other CableReady operations. I agree that it's not really a recommended practice, but I also don't want to prevent use cases we aren't considering or might be blind to. I'm going to merge #145 in lieu of this one as it also fixes the bug without adding restrictions to CableReady.

PR #142, opened 2020-04-09
Allow #stimulate to use promises

dark-panda:
# Feature

## Description

The basic idea here is that we keep a table of promises indexed by
randomly generated UUIDs. The UUIDs are sent to the server on each
reflex call and are returned with responses, and thus can be used to
look up the promise we kept in our table, which we then resolve or
reject. This allows us to do stuff like

```javascript
this.stimulate('MyReflex#action')
  .then(() => this.doSomething())
  .catch(() => this.handleError())
```

## Why should this be added

Primar benefits:

* Using promises gives you the option to keep callback code near the location of `#stimulate` calls, as well as implement per-call behaviour as opposed to generalized callback methods.

* You can take advantage of the full `Promise` API, including functions like `Promise.all`, `Promise.allSettled`, and `Promise.race`, allowing for simple reflex orchestration when you are making multiple calls to reflexes and need to ensure behaviour based on their overall executions.

* You can use `await/async` syntactic sugar to further simplify some code patterns.

## Checklist

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -37,7 +37,8 @@
     "@rails/actioncable": ">= 6.0",
     "cable_ready": ">= 4.1",
     "inflected": ">= 2.0",
-    "stimulus": ">= 1.1"
+    "stimulus": ">= 1.1",
+    "uuid": ">= 7.0.3"
   },
   "devDependencies": {
     "@babel/core": "^7.6.2",
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -1,5 +1,6 @@
 import { Controller } from 'stimulus'
 import CableReady from 'cable_ready'
+import { v4 as uuidv4 } from 'uuid'
 import { defaultSchema } from './schema'
 import { getConsumer } from './consumer'
 import { dispatchLifecycleEvent } from './lifecycle'
@@ -19,6 +20,8 @@ let stimulusApplication
 // A reference to the ActionCable consumer registered with: StimulusReflex.initialize or getConsumer
 //
 let actionCableConsumer
+
+const promises = {}
 
 // Initializes implicit data-reflex-permanent for text inputs.
 //
@@ -96,6 +99,7 @@ const extendStimulusController = controller => {
       }
       const attrs = extractElementAttributes(element)
       const selectors = getReflexRoots(element)
+      const reflexId = uuidv4()
       const data = {
         target,
         args,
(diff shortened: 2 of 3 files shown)


leastbad:
Having slept on this, I'm still excited about it, but there are a few things we should cover if we're going to merge this concept. It's a Venn diagram:

1. Can we make the data that you'd get back from a callback or DOM event available to the closure inside of then()? And the error info within catch()? I don't have much experience working with Promises so perhaps it's already there - but I don't think so. If you look at this data we make available in the callbacks for success and failure, this is what I have in mind: [element, reflex, [error]](https://docs.stimulusreflex.com/lifecycle#conventions).
2. Can we make the UUID available to the callback and DOM events? I would add it as a third method to the signature eg. element, reflex, uuid and element, reflex, uuid, error

Essentially, I'm trying to get all three potential methods of responding to a Reflex operation to have similar data available to them; equal if possible.

Apologies in advance if any of the above seems to miss obvious points. Again, I'm not super-confident working with Promises.

dark-panda:
Fixed and re-rolled. You now get a single argument (which is apparently all the `Promise` API supports which I had forgotten) which contains the following:

* `data`, which contains all of the data structure that was sent in the first place to the reflex, which itself contains the promise UUID

* `element` is the... element... 

* `event` as dispatched by the reflex

hopsoft:
This looks solid. Thank you for putting in the work on this. I'll be testing tomorrow and over the weekend.

hopsoft:
I've tested this and think it's a great addition. It also helped tease out a bug (that I'll fix shortly) with client side error handling of server errors. 

Also, here's an example of how to extract data from the promise payloads.

```js
this.stimulate('MyReflex#example')
  .then(payload => {
    const { data, element, event } = payload
    const { attrs, reflexId } = data
    // * attrs - an object that represents the attributes of the element that triggered the reflex
    // * data - the data sent from the client to the server over the web socket to invoke the reflex
    // * element - the element that triggered the reflex
    // * event - the source event
    // * reflexId - a unique identifier for this specific reflex invocation
  })
  .catch(payload => {
    const { data, element, event } = payload
    const { attrs, reflexId } = data
    const { error } = event.detail.stimulusReflex
    // * attrs - an object that represents the attributes of the element that triggered the reflex
    // * data - the data sent from the client to the server over the web socket to invoke the reflex
    // * element - the element that triggered the reflex
    // * error - the error message from the server
    // * event - the source event
    // * reflexId - a unique identifier for this specific reflex invocation
  })
```

PR #148, opened 2020-04-14
Check the ActionCable connection on stimuluate

hopsoft:
# Feature

## Description

Raise and exception if stimuluate is called before the ActionCable connection is open and ready.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -75,10 +75,10 @@ GEM
     concurrent-ruby (1.1.6)
     crass (1.0.6)
     erubi (1.9.0)
-    faraday (0.17.3)
+    faraday (1.0.1)
       multipart-post (>= 1.2, < 3)
-    faraday-http-cache (2.1.0)
-      faraday (~> 0.8)
+    faraday-http-cache (2.2.0)
+      faraday (>= 0.8)
     github_changelog_generator (1.15.2)
       activesupport
       faraday-http-cache
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -64,6 +64,7 @@ export const findElement = attributes => {
     let selectors = []
     for (const key in attributes) {
       if (key.includes('.')) continue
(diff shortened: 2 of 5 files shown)


leastbad:
This is a really good step in the right direction.

1. Can we also merge #86 alongside this? It probably needs to be rebased, but conceptually this is valid and viable.
2. What happens if stimulate is triggered by a declared reflex? I guess the error will raise but there's no opportunity to capture it.

julianrubisch:
👆🏻 2. was also my first thought?

hopsoft:
1. I'm not sure how I feel about #86 after revisiting it. I now consider it to be a concern that should be managed by users of the library. _But, we can and should provide documentation to help users understand how to best work with the socket connection in their controllers._ Here's an example:

```js
  connect () {
    super.connect()
    this.doStuff()
  }

  doStuff () {
    if (!this.isActionCableConnectionOpen()) 
      return setTimeout(this.doStuff.bind(this), 10)
    this.stimulate('ExampleReflex#do_stuff')
  }
}
```

2. This is already a problem today. The primary difference is that users don't currently get a helpful error message. I expect that in the future we may be able to implicitly check the connection status on `stimulate` and retry until the connection is ready, but we'll need to be thoughtful about this. Also, I see that feature as landing as a separate feature in a new PR.

leastbad:
1. I hope you revisit your revisitation! I believe that having the option to use style cascading (as nature intended) to hide or change UI items that aren't currently available is useful and powerful. Having to write a bunch of event handlers to do the same thing is just boilerplate.

2. I'm excited to help you flesh this out going forward. It's surprising to me that it hasn't been the source of more issues. I know that my Snake demo - aside from being thoroughly silly - frequently doesn't work if you don't wait a few seconds for WS to upgrade the connection.

hopsoft:
You talked me into it :)

PR #157, opened 2020-04-19
replace implicit data-reflex-permanent with skipMorph flag

leastbad:
# Enhancement

## Morph current focused input without mutating text value

**Important note: this PR is dependent on the acceptance of [this PR on the morphdom project](https://github.com/patrick-steele-idem/morphdom/pull/188) as there is no skipMorph checking in today's morphdom.**

morphdom is a miracle of software engineering, but it takes an all-or-nothing approach when it comes to updating an element; there's no facility to tweak or whitelist/blacklist what properties/attributes get updated during an operation.

StimulusReflex and LiveView both take pains to make sure that the client is a single source of truth; the server should not accidentally update the value of a text input that you're currently typing into. We created the `data-reflex-permanent` attribute to protect hierarchies of elements from all morph updates, but what if you just want to protect the value of the element you're typing into? When something is made permanent, you can no longer do any updates to it, even if they are desirable. Live feedback via CSS updates beome impossible, for example. You cannot dynamically disable or reenable an element.

Instead of setting the currently focused input to permanent, let's instead set a boolean value called `skipMorph` on the element which can be checked by morphdom. If `skipMorph` is true, allow morph updates unless they attempt to change the text value.

## Why should this be added

success

With this PR, developers will be able to change the CSS and other non-text attributes of the currently focused input element with a morph without overwriting the current text value. The client remains the single source of truth.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -23,22 +23,6 @@ let actionCableConsumer
 
 const promises = {}
 
-// Initializes implicit data-reflex-protect-value for text inputs.
-//
-const initializeImplicitReflexProtectValue = event => {
-  const element = event.target
-  if (!isTextInput(element)) return
-  element.skipMorph = true
-}
-
-// Resets implicit data-reflex-protect-value for text inputs.
-//
-const resetImplicitReflexProtectValue = event => {
-  const element = event.target
-  if (!isTextInput(element)) return
-  element.skipMorph = false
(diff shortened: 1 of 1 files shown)


leastbad:
Happy to say that this issue was closed #176.

Also posted [a summary on the morphdom issue tracker](https://github.com/patrick-steele-idem/morphdom/pull/188).

PR #160, opened 2020-04-21
add reflex callbacks

seb1441:
# Feature

## Description

- Add Reflex callbacks: `before_reflex`, `around_reflex` , `after_reflex`.
- All callbacks can receive multiple method names, an optional block and the following options: `only`, `except`, `if`, `unless`.
- Can halt a reflex via `throw :abort` in a `before_reflex` callback. Added by @hopsoft.

Examples:
```rb
  # will run only if the element has the step attribute, can use "unless" instead of "if" for opposite condition
  before_reflex :do_stuff, if: proc { |reflex| reflex.element.dataset[:step] }

  # will run only if the reflex instance has a url attribute, can use "unless" instead of "if" for opposite condition
  before_reflex :do_stuff, if: :url

  # will run before all reflexes
  before_reflex :do_stuff

  # will run before increment reflex, can use "except" instead of "only" for opposite condition
  before_reflex :do_stuff, only: [:increment]

  # will run around all reflexes, must have a yield in the callback
  around_reflex :do_stuff_around

 # will run after all reflexes
  after_reflex :do_stuff

  # Example with a block
  before_reflex do 
    # callback logic
  end

  # Example with multiple method names
  before_reflex :do_stuff, :do_stuff2

  # Example with halt
  before_reflex :run_checks

  def increment
    # reflex logic
  end

  def decrement
    # reflex logic
  end

  private

  def run_checks
    throw :abort # this will prevent the reflex from re-rendering the page
  end

  def do_stuff
    # callback logic
  end

  def do_stuff2
    # callback logic
  end

  def do_stuff_around
    # before
    yield
    # after
  end
end
```

## Why should this be added
https://stimulus-reflex.discourse.group/t/server-side-reflex-callbacks/26

## Checklist

### DIFF
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -55,9 +55,9 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     optional_params = method.parameters.select { |(kind, _)| kind == :opt }
 
     if arguments.size == 0 && required_params.size == 0
-      reflex.public_send method_name
+      reflex.process_action(method_name)
     elsif arguments.size >= required_params.size && arguments.size <= required_params.size + optional_params.size
-      reflex.public_send method_name, *arguments
+      reflex.process_action(method_name, *arguments)
     else
       raise ArgumentError.new("wrong number of arguments (given #{arguments.inspect}, expected #{required_params.inspect}, optional #{optional_params.inspect})")
     end
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -28,7 +28,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     begin
       reflex_class = reflex_name.constantize
       raise ArgumentError.new("#{reflex_name} is not a StimulusReflex::Reflex") unless is_reflex?(reflex_class)
-      reflex = reflex_class.new(self, url: url, element: element, selectors: selectors)
+      reflex = reflex_class.new(self, url: url, element: element, selectors: selectors, reflex_name: method_name)
       delegate_call_to_reflex reflex, method_name, arguments
     rescue => invoke_error
       message = exception_message_with_backtrace(invoke_error)
@@ -55,9 +55,9 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     optional_params = method.parameters.select { |(kind, _)| kind == :opt }
 
     if arguments.size == 0 && required_params.size == 0
-      reflex.process_action(method_name)
+      reflex.process_reflex(method_name)
     elsif arguments.size >= required_params.size && arguments.size <= required_params.size + optional_params.size
-      reflex.process_action(method_name, *arguments)
+      reflex.process_reflex(method_name, *arguments)
     else
       raise ArgumentError.new("wrong number of arguments (given #{arguments.inspect}, expected #{required_params.inspect}, optional #{optional_params.inspect})")
     end
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -1,10 +1,14 @@
 # frozen_string_literal: true
 
 class StimulusReflex::Reflex
+  include ActiveSupport::Callbacks
+
   attr_reader :channel, :url, :element, :selectors
 
   delegate :connection, to: :channel
   delegate :session, to: :request
+
+  define_callbacks :process_action
 
   def initialize(channel, url: nil, element: nil, selectors: [])
     @channel = channel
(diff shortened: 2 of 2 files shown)


hopsoft:
This is very promising. Thank you for working on it. 

I'm thinking through the API and would prefer something more along these lines.

```ruby
  before_reflex :do_stuff_before
  before_reflex :do_stuff_before_increment, only: [:increment]
  after_reflex do 
    puts "Hello from the after callback of all reflex methods"
  end

  def increment
    # reflex logic goes here...
  end

  def decrement
    # reflex logic goes here...
  end

  private

  def do_stuff_before
    puts "Hello from the before callback of all reflex methods"
  end

  def do_stuff_before_increment
    puts "Hello from the before callback of the increment reflex method"
  end
end
```

You're pretty close already. What do you think about expanding to include options like `only` and `except` as well as both symbols and blocks?

julianrubisch:
now I remember where I posted the feature request :D

to be fair, we treat discourse a bit shabby...

leastbad:
> What do you think about expanding to include options like only and except as well as both symbols and blocks?

Yes, please!

leastbad:
@hopsoft I'm still eagerly anticipating hearing more about what you meant in #80 when you said that callbacks could achieve the same outcome. Do you mean that you could return false from a before callback and prevent the action from happening?

hopsoft:
Yes that's certainly one way to handle it; however, I'm rethinking what a **no-render** or **no-broadcast** should be or look like. I believe it's a valid use case, but I'm not quite sure what shape it should have yet.

seb1441:
Sorry when I wrote the PR I didn't see the Reflex callbacks issue that was created 3 days ago, I hope nobody started working on it and that I didn't just steal the task. It honestly was just a coincidence that I created the PR 3 days after the issue on GitHub was created, all I saw was the request on discourse when I started working on that :facepalm:

As @julianrubisch mentioned in the issue, I also found Rails AbstractController::Callbacks when I was searching on how Rails implemented this functionality. You will notice that my code is taking example from that module.

@hopsoft About the naming of the new callbacks that you suggested, how about naming it `method` instead of `reflex`? I think this would also follow Stimulus naming conventions. How I see it, a reflex is an instance of `StimulusReflex::Reflex` and a method is well.. a reflex method. Examples with `method`: `before_method`, `method_name`, `process_method`.
We could also still name it action like in Rails controllers :thinking:. Need your opinions! :)

I should make another commit tonight with the except and only options and more.

julianrubisch:
> Sorry when I wrote the PR I didn't see the Reflex callbacks issue that was created 3 days ago, I hope nobody started working on it and that I didn't just steal the task. It honestly was just a coincidence that I created the PR 3 days after the issue on GitHub was created, all I saw was the request on discourse when I started working on that 🤦

No worries, it was both me 😁

leastbad:
@seb1441 I try (but don't always succeed) to be consistent with my terminology in the docs, having mostly settled on calling methods in a Reflex class "reflex actions". So by your logic, the real "argument" should be to use before_action. ;)

In reality, I can make arguments for all three. Like Nate, I think I *prefer* before_reflex aesthetically, but action or method are arguably more proper. I do think that it's important to remember that we have some artistic license to treat Reflex classes as a DSL.

julianrubisch:
„before_reflex“ and „after_reflex“ would be consistent with the JavaScript side of things, FWIW

seb1441:
Added a commit, should I squash any new commits or I leave that to the person who will merge?
- Renamed action to reflex
- Callbacks can now receive optional blocks. Passing method name symbols and a block at the same time will also work, the block will be called at the end.
- Added around_reflex callback which requires a yield. Note: For some reason, I couldn't make it work with a block directly, this callback for now can only be used with method name symbols.
- Added only/except and if/unless options, which required a new attribute to Reflex. Note: It might be a bit confusing in StimulusReflex::Channel#receive to pass method_name to the reflex class reflex_name attribute when there is also a variable named reflex_name above. 

Updated the PR's description to include new possibilities.

hopsoft:
@seb1441 Don't worry about squashing. We'll take care of that on merge.

hopsoft:
This is looking great. I'm looking forward to testing it. Should be soon.

seb1441:
Merged @hopsoft's changes.
He replaced some metaprogramming and added support for halting a reflex via throw :abort in a before_reflex method.

Will add his example to the PR description.

jonathan-s:
It would be nice with some documentation in this PR as well as some examples in the documentation for when this useful.

leastbad:
Documentation suggestions are welcome! I’m usually more on top of documenting everything obsessively but I’m juggling real life atm as many of you are.

It’s not a hard rule but historically we’ve done our documentation edits in GitBook. Sometimes people make documentation PRs and I just copy them over manually and give thanks in the commit message.

hopsoft:
@leastbad Of note, this will make `process` a reserved word/method-name in reflexes.

leastbad:
Congrats to @seb1441 for leading this charge and thanks to @julianrubisch for his ideas and inspiration. This PR is epic and makes StimulusReflex a better library.

PR #163, opened 2020-04-21
add console log messages for every reflex call

marcoroth:

Enhancement

## Description

This PR adds logging functionality to show which reflexes were called and what data was sent to the backend. 

This is heavily inspired by https://github.com/LogRocket/redux-logger

## Screenshot

log-reflex

## Why should this be added

I think this is useful for debugging and tracing purposes. For example if you want to know what data was sent to the reflex or in which order the reflexes were called.

## Discussion points
- [x] Maybe it makes sense to provide an option to disable the functionality completely (for example in the production environment or based on a log level) 

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -130,7 +130,7 @@ const extendStimulusController = controller => {
       const promise = new Promise(
         (resolve, reject) => (promises[reflexId] = { resolve, reject, data })
       )
-      promise.catch(() => {}) // noop default catch
+      if (debugging) promise.catch(() => {}) // noop default catch
       return promise
     },
 
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -321,7 +321,7 @@ if (!document.stimulusReflexInitialized) {
 
     const element = findElement(attrs)
     const promise = promises[event.detail.stimulusReflex.reflexId]
-    const response = { data: promise?.data, element, event }
+    const response = { data: (promise && promise.data), element, event }
 
     if (options.logging) {
       Log.response(response)
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -2,11 +2,10 @@ import { Controller } from 'stimulus'
 import CableReady from 'cable_ready'
 import { v4 as uuidv4 } from 'uuid'
 import { defaultSchema } from './schema'
-import { defaultOptions } from './options'
 import { getConsumer } from './consumer'
 import { dispatchLifecycleEvent } from './lifecycle'
 import { allReflexControllers } from './controllers'
-import { logReflex } from './log'
+import Log from './log'
 import {
   attributeValue,
   attributeValues,
@@ -24,6 +23,12 @@ let stimulusApplication
 let actionCableConsumer
 
 const promises = {}
+
+// This object holds the options for StimulusReflex
+//
+let options = {
+  logging: false
+}
 
 // Initializes implicit data-reflex-permanent for text inputs.
 //
@@ -132,15 +137,19 @@ const extendStimulusController = controller => {
 
       dispatchLifecycleEvent('before', element)
 
-      if (stimulusApplication.options.logging) {
-        logReflex(reflexId, this.context.scope.identifier, target, element, args[0])
-      }
-
       subscription.send(data)
 
-      return new Promise((resolve, reject) => {
+      const promise = new Promise((resolve, reject) => {
         promises[reflexId] = { resolve, reject, data }
-      })
+      });
+
+      if (options.logging) {
+        Log.request(reflexId, target, this.context.scope.identifier, element, args[0])
+        // promise.then(response => Log.response(response));
+        promise.catch(response => Log.error(response));
+      }
+
+      return promise;
     },
 
     // Wraps the call to stimulate for any data-reflex elements.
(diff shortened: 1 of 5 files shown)


jonathan-s:
Neat! I can definitely see that this is useful. Though I would also want a way to turn this off.

leastbad:
This is interesting. I also remember that at one point @julianrubisch was looking at what it would take to create a Chrome plugin (was it for Stimulus or StimulusReflex) and that's another way that this kind of thing might be even more useful. The sky and your free time is truly the limit.

As @jonathan-s said, this would have to be off by default, as logging some applications would get very noisy fast. However, I'm intrigued by the different ways it could be enabled as it might not just be all-or-nothing at load time. We could explore exposing a boolean value on each controller, allowing us to dynamically turn on and off the log flow very granularly.

julianrubisch:
Maybe overkill, but how about https://github.com/flyerhzm/uniform_notifier/ ?

leastbad:
Cool library, but seems again like overkill for what we need in this particular case, as you said yourself.

Just my $0.02! I've never used Uniform Notifier (yet!)

julianrubisch:
I think the question mainly is how much customization we want to permit?

marcoroth:
I came up with some ideas on how to enable this logging functionality. I don't know if every option is technically doable, but those are just some ideas that came to me mind.

I agree, maybe it makes sense to disable the functionality by default. You would have to opt-in the enable this extensive logging. That's why I added the discussion point 😊

## Option A - Enable in Initializer

Enable the logging in a initializer. Very Rails-like but it's probably "hard" to get this config to the client in a nice way.

```ruby
# config/initializers/stimulus_reflex.rb

StimulusReflex.config do |config|
  config.enable_logging = true
  # ...
end

```

## Option B - Enable with Data Attribute

Set a data attribute on the body element to enable the logging for all reflexes. Very lightweight and simple, would fit the to other data attributes which are very Stimulus-like.

```html
<body data-reflex-logging="true">
  <div class="container" data-controller="example">
    <button data-reflex="ExampleReflex#increment">+</button>
  </div>
</body>
```


## Option C - Enable in Reflex

Enable the logging functionality on a per reflex basis or enable it on the `ApplicationReflex` to enable it for all reflexes.

But we have the same question here, how do we get the configuration to the client in an easy way without introduction a lot of new techniques.

```ruby
# for every reflex
class ApplicationReflex < StimulusReflex::Reflex
  self.enable_logging = true
end

# or for a single reflex
  self.enable_logging = true

  def increment
    # ...
  end
end
```

## Option D - Enable in JavaScript

I think the easiest way to implement this, is to provide an option in the StimulusReflex initialization script. Maybe by adding an extra argument to the `initialize` method or by providing an extra function on the `StimulusReflex` class. 

It also makes sense to add the configuration to the JavaScript part of the library, because this option doesn't affect any part of the backend.

```javascript
// app/javascript/controllers/index.js

// ..

options = { logging: true }
StimulusReflex.initialize(application, { consumer }, options)

// or

StimulusReflex.initialize(application, { consumer })
StimulusReflex.enableLogging()
```

## Other Options

I'm sure there are a lot of other ways to implement this. But I think it shouldn't be an overkill solution. 
But I'm very open about any other ideas. At the end it should fit into the StimulusReflex library and should simple to enable.

What do you guys think, which option would you prefer?

seb1441:
@marcoroth I like option A, also this config could eventually be used for other things.

I personally don't see a situation where I would like logging only on a specific view or reflex.

Option D first example where you pass the logging option to the initializer is nice too so that we have separate toggles for the JavaScript side and Ruby side.

jonathan-s:
Personally I'd prefer option D or some variant thereof as this primarily concerns javascript. If you want to have debugging per reflex you could perhaps make that work with the specific javascript controller that is the frontend part of the rails reflex.

leastbad:
Relevant:

[](http://www.youtube.com/watch?v=MZvmYaFkNJI&t=780 "")

marcoroth:
@leastbad right, I also saw that. It's funny because I watched this video right after I posted the comment yesterday and thought "Well this is exactly what I just wrote in Option D" 😅 

Personally I would also prefer option D.

marcoroth:
Okay, I just implemented Option D as a proof-of-concept. With those changes StimulusReflex can now get initialized like this: 

```javascript
// app/javascript/controllers/index.js

// Default behavior - no logging - as is
StimulusReflex.initialize(application, { consumer })

// Option 1 - Pass logging option with initialize
StimulusReflex.initialize(application, { consumer, logging: true })

// Option 2 - Enable logging by calling the enableLogging() method
StimulusReflex.initialize(application, { consumer })
StimulusReflex.enableLogging();

// For completeness there is also a disableLogging() method
StimulusReflex.disableLogging();
```

javascript/stimulus_reflex.js
```diff
@@ -266,18 +273,28 @@ const getReflexRoots = element => {
 // - options
 //   * controller - [optional] the default StimulusReflexController
 //   * consumer - [optional] the ActionCable consumer
+//   * logging - [optional] option wheter reflex logging should be enabled or not
 //
 const initialize = (application, options = {}) => {
   const { controller, consumer } = options
   actionCableConsumer = consumer
   stimulusApplication = application
+  stimulusApplication.options = { ...defaultOptions, ...options }
```
hopsoft, line 282: I wouldn't hang our options off of the Stimulus application given that this is only for StimulusReflex. I'd probably track this as a file scope level variable similar to what we're doing with [promises](https://github.com/hopsoft/stimulus_reflex/blob/master/javascript/stimulus_reflex.js#L24).

```js
const options = {
  logging: false
}

// ...

if (options.logging) {
  // ...
}

// ...

export default {
   initialize,
   register,
   options
 }
```

marcoroth:
Thanks for the review @hopsoft

I changed the definition of the options. They are now defined in the file. I tried to apply your suggestion to export the options, but it seems like it doesn't work that way.
 
If you export the options and try to set it with `StimulusReflex.options.logging = true` it will write it in the exported object which isn't the same as the one defined inside `stimulus_reflex.js`. 

I swapped the reflexId with the date as you suggested.

Also there is now support for response and error logging. If a stimulate call has multiple `data-reflex-root` it will display a response log for every root. They are indicated with the number at the end (i.e. `[1]`). The invoking code could be refactored a little bit once #171 is resolved.

Screenshot 2020-04-26 at 04 59 02

And with errors:

Screenshot 2020-04-26 at 04 58 28

If you have logging disabled and an error occurs it now shows the right error message. Before it just logged `Unhandled Promise Rejection: [Object object]`

Screenshot 2020-04-26 at 05 07 14

Let me know if we should add/change something 😄

hopsoft:
Makes sense. I'm looking forward to testing this. Looks like we have a conflict that will need to be resolved too.

hopsoft:
Solid addition to the library. Great job on this.

PR #167, opened 2020-04-25
Setup instructions for django

jonathan-s:
This is a start of documenting how the stimulus-reflex technique is applied in django and the sockpuppet library. 

## Description
This is a start for documenting how the stimulus-reflex technique is applied in the sockpuppet library in django. 

## Why should this be added
I'd be happy to add documentation in one place. It makes the project bigger, and sockpuppet uses the javascript part of stimulus reflex so any changes made to that would need to be taken into consideration by the sockpuppet library. For that reason it makes sense to me to keep the documentation in the same place. 

I'm adding this as a draft PR and will make this a normal PR once all documentation is finished. I'd be happy to hear any thoughts or concerns that you might have.


jonathan-s:
@hopsoft Reading this, what questions would you ask?

jonathan-s:
Closing this as documentation is now @ https://sockpuppet.argpar.se

PR #169, opened 2020-04-25
setupDeclarativeReflexes export with UJS support

### LINKED ISSUE #161: Reflex not binding to ajax loaded content
## Describe the bug

Reflex not binding to Ajax loaded content that is loaded onto page from Rails UJS remote call. 

## To Reproduce

- Use remote UJS call to load html onto page
- New html content has data-reflex attribute
- Reflex is not binding

## Expected behavior

Reflex should bind to ajax loaded content

### StimulusReflex

- stimulus_reflex 3.1.3

### External tools

- Ruby: 2.6.3
- Rails: 6.0.2
- Turbolinks 5.2.1

### Browser

- Chrome

leastbad:
# Enhancement

Scan DOM for `data-reflex` attributes after UJS operations.

Also export `setupDeclarativeReflexes` so that it can be called from a controller that utilizes a template engine.

Fixes #161 

## Why should this be added

Cover a new use case and make the library more flexible.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -287,6 +287,9 @@ if (!document.stimulusReflexInitialized) {
   document.addEventListener('cable-ready:after-morph', () =>
     setTimeout(setupDeclarativeReflexes, 1)
   )
+  document.addEventListener('ajax:complete', () =>
+    setTimeout(setupDeclarativeReflexes, 1)
+  )
   // Trigger success and after lifecycle methods from before-morph to ensure we can find a reference
   // to the source element in case it gets removed from the DOM via morph.
   // This is safe because the server side reflex completed successfully.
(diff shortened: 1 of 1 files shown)


nflorentin:
Hi!

I don't see anymore the code of this commit in master (setupDeclarativeReflexes on ajax:complete), is there any reason for that ?
It is handle by another piece of code ? or the developer has to trigger it manually ? Will trigger manually the event "cable-ready:after-morph" after an ajax call do the job ?

Thanks in advance for your answer

leastbad:
Hi @nflorentin! The SR client has been extensively refactored without breaking its public API over the past minor versions.

You shouldn't need to manually trigger `setupDeclarativeReflexes` ever. If you look at the function on the [master branch](https://github.com/hopsoft/stimulus_reflex/blob/fb94b906cbc0f0c18d3f8ad3498ae37b45443c95/javascript/stimulus_reflex.js#L454) you'll see that we run our own MutationObserver which should pick up all DOM changes and scan for declared Reflexes.

Are you having a specific issue that we can help with? Jump on Discord and we can talk about it.

PR #105, opened 2019-12-23
Support devise authenticated routes

hopsoft:
# Fix

## Description

Simplify how the request is reconstructed.
Also use the request when parsing the URL.

This change supports Devise authenticated routes.

Fixes #104

Related https://stimulus-reflex.discourse.group/t/cannot-get-sr-with-devise-to-work/14

## Checklist

- [x] My code follows the style guidelines of this project
- [x] Checks (StandardRB & Prettier-Standard) are passing

PR #172, opened 2020-04-25
Support devise authenticated routes

### LINKED ISSUE #173: Devise authenticated routes not supported anymore
# Bug Report

## Describe the bug

When SR is re-rendering the page (accessible for Devise authenticated user only) after a reflex has been triggered, this message appears in the logs:

```
StimulusReflex::Channel Failed to re-render http://localhost:3000/ undefined method `authenticate?' for nil:NilClass /Users/dbo/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/devise-4.7.1/lib/devise/rails/routes.rb:479:in `block in constraints_for'
```

## To Reproduce

- A route scoped in  devise `authenticated`
such as:

```
  authenticated :user do
    root to: "CONTROLLER#ACTION", as: :SOME_PATH
  end
```

## Expected behavior

Reflex should re-render the page and morph it into the page in the browser.

```
StimulusReflex::Channel Failed to re-render http://localhost:3000/ undefined method `authenticate?' for nil:NilClass /Users/dbo/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/devise-4.7.1/lib/devise/rails/routes.rb:479:in `block in constraints_for'
```

### StimulusReflex

- Gem: 3.1.3
- Node package: 3.1.3

### External tools

- Ruby: 2.6.5
- Rails: 6.0.2.2
- Node: v10.15.3
- Devise: 4.7.1

### Browser

- Firefox
- 75

db0sch:

Bug fix

## Description

According to this PR https://github.com/hopsoft/stimulus_reflex/pull/105 , the devise authenticated routes were supposed to be supported. But I have the feeling that the rewriting of some code with the newest versions introduced this issue once again.

I tried to implement a fix to highlight were the problem is according to me. But I'm not over confident on these topics (request, action dispatch...), so I solved it in a (too?) simple way. Hope this can be helpful.

The main thing I don't like in my code is calling `Rails.application.routes.recognize_path_with_request` once again, in addition to `StimulusReflex::Reflex#url_params`. But at this stage, I'm not sure how to avoid this call.

Happy to discuss it with you, and see how we can improve this, or, find how to solve this issue in another way.

Fixes #173 

## Why should this be added

Authenticated devise routes are quite common in the rails project. Not supporting this would require developers to find workarounds of not using, say, a specific `root_path` for their authenticated users. It has been fixed once by @hopsoft , so I guess we should re-introduce it again.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -17,9 +17,8 @@ class StimulusReflex::Reflex
     @request ||= begin
       uri = URI.parse(url)
       path = ActionDispatch::Journey::Router::Utils.normalize_path(uri.path)
-      path_params = Rails.application.routes.recognize_path(path)
       query_hash = Rack::Utils.parse_nested_query(uri.query)
-      ActionDispatch::Request.new(
+      req = ActionDispatch::Request.new(
         connection.env.merge(
           Rack::MockRequest.env_for(uri.to_s).merge(
             "rack.request.query_hash" => query_hash,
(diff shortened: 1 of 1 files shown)


hopsoft:
Good catch. Thanks for this PR.

db0sch:
Thank you @hopsoft 🙏

leastbad:
Yes, this is a life-saver.

PR #175, opened 2020-04-26
147 extract multiple checkbox values

### LINKED ISSUE #147: Allow extractElementAttributes to use a checkbox list
# Feature Request

`extractElementAttributes` for the moment returns multiple values for `select` tags with `multiple: true`:

https://github.com/hopsoft/stimulus_reflex/blob/480ee76dccf4ddc212602d1c3dee01216b6369ab/javascript/attributes.js#L37

A common scenario is to use a checkbox list for an association collection, as in `collection_check_boxes`:

https://devdocs.io/rails~6.0/actionview/helpers/formoptionshelper#method-i-collection_check_boxes

It would be nice if SR supported that out of the box.

The hidden complexity lies in the fact that we'd have to consider multiple elements in `extractElementAttributes`, not just the one triggering the reflex. Still I think a library such as SR should take on that complexity instead of imposing it on the back end developer.

Specifically, I discovered this shortage when working on the permalink expo example and trying to employ a button group: https://github.com/hopsoft/stimulus_reflex_expo/pull/26/files#diff-23c58647594fccfb30d0b54554851d59R15

julianrubisch:
# Enhancement

## Description

I circled around the implementation and finally came up with this, because it allows for maximum extensibility (IMHO) should more cases arrive in the future.

I also delete `value` from `attrs` if there are multiple values present because IMHO it's confusing.

Two tests were also added.

Fixes #147 

## Why should this be added

When preparing [this expo example](https://github.com/hopsoft/stimulus_reflex_expo/pull/26) I noticed that it would be good to have native support for multiple selects that are realized as checkboxes (as in Rails' `collection_check_boxes`)

## Checklist


javascript/attributes.js
```diff
@@ -31,23 +33,52 @@ export const extractElementAttributes = element => {
...
-      const checkedOptions = Array.prototype.slice.call(
-        element.querySelectorAll('option:checked')
-      )
-      attrs.values = checkedOptions.map(o => o.value)
-    } else if (element.selectedIndex > -1) {
-      attrs.value = element.options[element.selectedIndex].value
+
+  if (hasMultipleOptions(element)) {
+    attrs.values = collectCheckedOptions(element)
+    delete attrs.value
```
hopsoft, line 42: Wondering if we should leave value assigned (as a comma delimited list) for consistency. That would mean selects would look like this on the server.

```ruby
# single value
element[:value]  #=> 1
element[:values] #=> [1]

# multiple values
element[:value]  #=> "1,2,3"
element[:values] #=> [1,2,3]
```

javascript/attributes.js
```diff
@@ -31,23 +33,52 @@ export const extractElementAttributes = element => {
...
+const hasMultipleOptions = element => {
+  const multipleElementCount = document.querySelectorAll(
+    `input[type="${element.type}"][name="${element.name}"]`
+  ).length
+
+  return (
+    (element.tagName.match(/select/i) && element.multiple) ||
+    multipleElementCount > 1
+  )
+}
```
hopsoft, line 64: We could simplify or eliminate this guard function if we adopt my proposal above as all we'd need to verify is that we're dealing with a select.

javascript/attributes.js
julianrubisch: good idea! we should mention that in the docs, too

javascript/attributes.js
julianrubisch: I'm not sure I'm following. What about multiple `input`s ?

hopsoft: Something like this is what I was thinking when I wrote that comment. Not sure if it covers all edge cases though.

```js
const hasMultipleOptions = element => {
  if (element.tagName === 'SELECT' && element.multiple) return true 
  if (element.tagName === 'CHECKBOX' && document.querySelectorAll(`[name="${element.name}"]`).length > 1)
    return true
  return false
}
```

julianrubisch: I see... well I wanted it to be a bit more generic because I'm not 100% sure that checkboxes just cover all scenarios... there might be some exotic cases, as you said, of other multiple elements...

julianrubisch: but I'll rework if you think it's necessary

PR #176, opened 2020-04-26
Remove implicit permanent for text inputs

### LINKED ISSUE #159: Input field values sometimes remain
# Bug Report

When removing a child record from a form, values sometimes remain in an input field.

## Describe the bug

I have a form that includes nested child records. Each child record contains a "Remove" button that sets the `_destroy` attribute on the record to `true` and then sends serialized form data to the server for validation. The server correctly handles the request, and the html coming back from the server is correct. But input fields sometimes appear with data from the deleted child record still showing.

This seems to be a diffing problem. The HTML coming back from the server is right but is not correctly inserted into the page.

## Expected behavior

Replaced HTML should be faithful to the rendered server response.

Screen recording input fields

### StimulusReflex

- Gem: 3.1.3
- Node package: 3.1.3

### External tools

- Ruby: 2.7
- Rails: 6.0.2.2

### Browser

- Browser Safari <= Note: I have not been able to reproduce this bug in Chrome
- Version 13.1

hopsoft:
# Enhancement and bug fix

## Description

CableReady now handles the responsibility of protecting text input values which allows us to remove this concern from StimulusReflex.

This provides a much more robust protection for text input values.

Designed to work with CableReady changes in PR https://github.com/hopsoft/cable_ready/pull/33

Hopefully this will resolve #159 Should also resolve the concerns that #157 was aimed at.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     stimulus_reflex (3.1.3)
-      cable_ready (>= 4.1)
+      cable_ready (>= 4.1.2)
       nokogiri
       rack
       rails (>= 6.0)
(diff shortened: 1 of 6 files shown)


hopsoft:
I'm hoping that this may also address #159

leastbad:
https://discordapp.com/channels/629472241427415060/683317077423292582/703972444649553940

PR #178, opened 2020-04-29
 Pluralize the generated class name, so that will match with the file name

darkrubyist:
bug fix

## Description
I notice in the SR screencast - Build a Twitter clone in 10 minutes with Rails, CableReady, and StimulusReflex: https://www.youtube.com/watch?v=F5hA79vKE_E that the Reflex generator is generating a singular class name. 

## Why should this be added
This PR should pluralize the generated class name, so that will match with the file name.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,15 +122,10 @@
 
 - Reload session prior to each reflex accessing it [\#131](https://github.com/hopsoft/stimulus_reflex/pull/131) ([hopsoft](https://github.com/hopsoft))
 - tweak prettier-standard and add actions caching [\#125](https://github.com/hopsoft/stimulus_reflex/pull/125) ([andrewmcodes](https://github.com/andrewmcodes))
-- Propose post install message [\#122](https://github.com/hopsoft/stimulus_reflex/pull/122) ([julianrubisch](https://github.com/julianrubisch))
 
 **Closed issues:**
 
 - Cannot read property 'stimulusReflexController' of null [\#127](https://github.com/hopsoft/stimulus_reflex/issues/127)
-
-**Merged pull requests:**
-
-- Bump actionview from 6.0.2.1 to 6.0.2.2 [\#128](https://github.com/hopsoft/stimulus_reflex/pull/128) ([dependabot[bot]](https://github.com/apps/dependabot))
 
 ## [v2.2.2](https://github.com/hopsoft/stimulus_reflex/tree/v2.2.2) (2020-03-04)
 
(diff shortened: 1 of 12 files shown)


marcoroth:
Good observation! I agree that the filename should match the defined class in it.

But is this really the wanted behaviour for every case? I wouldn't say that the reflex name should always be pluralized, because the reflex doesn't have to match to an according model. I think it would make sense if the behaviour would match the Rails controller generator.

This means that:
* `rails g stimulus_reflex posts` generates a `posts_reflex.rb` file which defines the `PostsReflex` class
* and `rails g stimulus_reflex post` generates a `post_reflex.rb` file which defines the `PostReflex` class

What do you think @darkrubyist?

darkrubyist:
Thanks, @marcoroth. This commit should support both singular and plural use cases.

marcoroth:
Nice, I like this addition! 🎉 

One little detail: Does it also make sense to update the reflex class references for consistency in the js comments of [custom_controller.js](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/generators/templates/custom_controller.js) ?

darkrubyist:
Thanks @marcoroth .  Updated the custom_controller.js and added tests for both scenarios.

I notice the `bundle exec rails d stimulus_reflex User` will remove the `application_controller.rb` and `application_controller.js` as well.

Do you think there is a way to not do that?

marcoroth:
Yes there is. You can check the value of the `behavior` variable. The value is either `:invoke` or `:revoke`. 

Something like this should do the trick:
```ruby
if behavior == :invoke
  template "application_reflex.rb", File.join(REFLEX_BASE_PATH, "application_reflex.rb")
end
```

hopsoft:
Great observation and addition. I'm happy to merge this once conflicts are resolved and we update to prevent removing of the application controller files.

leastbad:
Huzzah! Congratulations to @darkrubyist on a great PR. Thanks @marcoroth for your contributions.

leastbad:
QQ: Just so everything lines up just right, do I need to tweak anything in this section? https://docs.stimulusreflex.com/quickstart#stimulusreflex-generator

darkrubyist:
Thanks @hopsoft, @leastbad, @marcoroth for help. :)

Now, we support both singular and plural versions, so users can use:

`bundle exec rails generate stimulus_reflex user`
`bundle exec rails generate stimulus_reflex users`

`application_reflex.rb` and `application_controller.js` will not be removed anymore if the user will run `bundle exec rails d stimulus_reflex user` or `bundle exec rails d stimulus_reflex users`.

@leastbad, @hopsoft recommended the use of the plural version, similar to rails controllers.

PR #180, opened 2020-05-01
Allow channel exceptions to be rescuable

dark-panda:
# Feature

## Description

Exceptions in reflexes are currently logged and swallowed. This will give reflexes the ability to handle the exceptions via `rescue_from` in much the same manner as you would with ActionController.

## Why should this be added

We'd like to track reflex errors via an external exception tracker, but the errors are currently swallowed with no chance to handle them.

This PR allows you to use `rescue_from` in reflexes thusly:

```ruby
class MyTestReflex < StimulusReflex::Reflex
  rescue_from StandardError do |exception|
    ExceptionTrackingService.error(exception)
  end

  # ...
end
```

## Checklist

### DIFF
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -31,6 +31,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
       reflex = reflex_class.new(self, url: url, element: element, selectors: selectors)
       delegate_call_to_reflex reflex, method_name, arguments
     rescue => invoke_error
+      reflex.rescue_with_handler(invoke_error)
       message = exception_message_with_backtrace(invoke_error)
       return broadcast_error("StimulusReflex::Channel Failed to invoke #{target}! #{url} #{message}", data)
     end
@@ -38,6 +39,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     begin
       render_page_and_broadcast_morph reflex, selectors, data
     rescue => render_error
+      reflex.rescue_with_handler(render_error)
       message = exception_message_with_backtrace(render_error)
       broadcast_error "StimulusReflex::Channel Failed to re-render #{url} #{message}", data
     end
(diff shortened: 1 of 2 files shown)


lib/stimulus_reflex/channel.rb
```diff
@@ -31,13 +31,15 @@ def receive(data)
       reflex = reflex_class.new(self, url: url, element: element, selectors: selectors)
       delegate_call_to_reflex reflex, method_name, arguments
     rescue => invoke_error
+      rescue_in_reflex(reflex, invoke_error)
```
hopsoft, line 34: Could we simplify by just calling `reflex.rescue_with_handler(exception)` here?

lib/stimulus_reflex/channel.rb
```diff
@@ -31,13 +31,15 @@ def receive(data)
...
       message = exception_message_with_backtrace(invoke_error)
       return broadcast_error("StimulusReflex::Channel Failed to invoke #{target}! #{url} #{message}", data)
     end
 
     begin
       render_page_and_broadcast_morph reflex, selectors, data
     rescue => render_error
+      rescue_in_reflex(reflex, render_error)
```
hopsoft, line 42: Could we simplify by just calling reflex.rescue_with_handler(exception) here?

lib/stimulus_reflex/channel.rb
```diff
@@ -117,4 +119,8 @@ def broadcast_error(message, data = {})
   def exception_message_with_backtrace(exception)
     "#{exception} #{exception.backtrace.first}"
   end
+
+  def rescue_in_reflex(reflex, exception)
+    reflex.handle_channel_exception(exception)
+  end
```
hopsoft, line 125: I think we may be able to omit this method. See comments above.

lib/stimulus_reflex/reflex.rb
```diff
@@ -41,4 +43,8 @@ def request
   def url_params
     @url_params ||= Rails.application.routes.recognize_path_with_request(request, request.path, request.env[:extras] || {})
   end
+
+  def handle_channel_exception(exception)
+    rescue_with_handler(exception)
+  end
```
hopsoft, line 49: I think we may be able to omit this method. See comments above.

leastbad:
Thank you for making StimulusReflex a better library, @dark-panda. This PR is excellent.

PR #181, opened 2020-05-02
Replace uuid4 dependency with function in repo

jonathan-s:
An enhancement. 

## Description
I've replaced the dependency of uuid with our own function. So one less dependency + this function is shorter and should take less space than the original. 

## Why should this be added
It'll decrease the bundle size, if even marginally + less dependencies are better. 

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -1,6 +1,5 @@
 import { Controller } from 'stimulus'
 import CableReady from 'cable_ready'
-import { v4 as uuidv4 } from 'uuid'
 import { defaultSchema } from './schema'
 import { getConsumer } from './consumer'
 import { dispatchLifecycleEvent } from './lifecycle'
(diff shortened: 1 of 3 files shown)


leastbad:
On one hand, this should get tree shaken such that wire size isn't too horrible.

On the other hand... is that really a viable uuid4 implementation? If so, wow!

jonathan-s:
Tangentially related, it doesn't look like inflected is treeshaked. So that might need some investigation as well.

hopsoft:
I love the idea of removing dependencies if we can simply pull in the relevant bits of code after vetting the functionality. It will improve security and possibly package size. For cross reference, here's the implementation from the uuid package. https://github.com/uuidjs/uuid/blob/master/src/v4.js

javascript/stimulus_reflex.js
```diff
@@ -22,6 +21,17 @@ let actionCableConsumer
 
 const promises = {}
 
+function uuidv4 () {
```
hopsoft, line 24: I'm actually changing all function definitions to use function expressions so they're block scoped and a little more strict. I'd prefer to see this defined like so.

```js
const uuidv4 = () => {
```

jonathan-s: Thank you, now I learnt something new.

> Function expressions are invoked to avoid polluting the global scope. Instead of your program being aware of many different functions, when you keep them anonymous, they are used and forgotten immediately.

PR #182, opened 2020-05-02
Introduce websocket adapter to stimulusreflex for more flexibility

jonathan-s:
An enhancement! 

## Description
This is a proposal to adapt the code to use an adapter for websockets. Which would give the user of the library a bit more flexibility, and it would make it easier to replace actioncable with something else. 

If this proposal is accepted, it would make it possible to split the javascript into `stimulus-reflex` which would be for rails alone and `stimulus-reflex-core` which is completely independent of actioncable. That in turn would make the bundle smaller for python users, as they would unnecessarily need to install actioncable as it is now. 

The adapter pattern would also allow for transport over http as I will explain in comments below. 

As a reference on how this affects django-sockpuppet -> https://github.com/jonathan-s/django-sockpuppet/pull/13

## Checklist

### DIFF
--- a/javascript/adapter.js
+++ b/javascript/adapter.js
@@ -1,39 +1,31 @@
 import CableReady from 'cable_ready'
 
-class ConsumerAdapter {
+class AbstractConsumerAdapter {
   constructor (consumer) {
     this.consumer = consumer
-  }
-
-  find_subscriptions (identifier) {
-    throw Error('Not implemented')
-  }
-
-  create_subscription (channel) {
-    throw Error('Not implemented')
-  }
-
-  isConnected () {
-    throw Error('Not implemented')
-  }
-
-  send (identifier, data, options = {}) {
-    throw Error('Not implemented')
-  }
-
-  connect () {
-    throw Error('Not implemented')
-  }
-
-  disconnect () {
-    throw Error('Not implemented')
+    if (new.target === AbstractConsumerAdapter) {
+      throw new TypeError('Cannot construct an abstract instance directly')
+    }
+    const methods = [
+      // takes an argument identifier
+      'find_subscription',
+      // takes an argument channel
+      'create_subscription',
+      'isConnected',
+      // takes arguments identifier, data, options
+      'send',
+      'connect',
+      'disconnect'
+    ]
+    for (const method of methods) {
+      if (typeof this[method] === undefined) {
+        throw new TypeError(`Must override the method ${method}`)
+      }
+    }
   }
 }
 
-class ActionCableAdapter extends ConsumerAdapter {
-  constructor (consumer) {
-    this.consumer = consumer
-  }
+class ActionCableAdapter extends AbstractConsumerAdapter {
 
   find_subscription (identifier) {
     return this.consumer.subscriptions.findAll(identifier)[0]
@@ -72,4 +64,4 @@ class ActionCableAdapter extends ConsumerAdapter {
   }
 }
 
-export { ActionCableAdapter, ConsumerAdapter }
+export { ActionCableAdapter, AbstractConsumerAdapter }
--- a/javascript/adapter.js
+++ b/javascript/adapter.js
@@ -25,8 +25,11 @@ class AbstractConsumerAdapter {
   }
 }
 
-class ActionCableAdapter extends AbstractConsumerAdapter {
+const getAbstractClass = () => {
+  return AbstractConsumerAdapter
+}
 
+class ActionCableAdapter extends AbstractConsumerAdapter {
   find_subscription (identifier) {
     return this.consumer.subscriptions.findAll(identifier)[0]
   }
@@ -64,4 +67,4 @@ class ActionCableAdapter extends AbstractConsumerAdapter {
   }
 }
 
-export { ActionCableAdapter, AbstractConsumerAdapter }
+export { ActionCableAdapter, getAbstractClass }
--- a/javascript/adapter.js
+++ b/javascript/adapter.js
@@ -1,0 +1,75 @@
+import CableReady from 'cable_ready'
+
+class ConsumerAdapter {
+  constructor (consumer) {
+    this.consumer = consumer
+  }
+
+  find_subscriptions (identifier) {
+    throw Error('Not implemented')
+  }
+
+  create_subscription (channel) {
+    throw Error('Not implemented')
+  }
+
+  isConnected () {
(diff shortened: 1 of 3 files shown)


javascript/adapter.js
```diff
@@ -0,0 +1,76 @@
+import CableReady from 'cable_ready'
+
+class ConsumerAdapter {
+  constructor(consumer) {
+    this.consumer = consumer
+  }
+
+  find_subscriptions(identifier) {
+    throw Error('Not implemented')
+  }
```
jonathan-s, line 10: The idea here is to have an abstract class, not sure what the pattern is for javascript. So would be helpful to hear some feedback!

javascript/stimulus_reflex.js
```diff
@@ -27,24 +28,15 @@ const promises = {}
...
-          if (urls.length !== 1 || urls[0] !== location.href) return
-        }
-        CableReady.perform(data.operations)
-      }
-    })
+  let subscription =
+    adaptedConsumer.find_subscription(identifier) ||
+    adaptedConsumer.create_subscription(channel)
+
+  controller.StimulusReflex.subscription = subscription
```
jonathan-s, line 39: We don't really need the subscription at a later point. So we could actually delete this line.

javascript/stimulus_reflex.js
```diff
@@ -107,6 +100,7 @@ const extendStimulusController = controller => {
       element.reflexData = data
 
       dispatchLifecycleEvent('before', element)
+      adaptedConsumer.send(JSON.stringify({ channel }, data, stimulateOptions))
```
jonathan-s, line 103: With an options we have increased flexibility in how sending stuff over the wire works. This would allow a user to create an adapter that actually sends stuff over http instead of websockets. 

So the options object could contains something like `{"method": "POST"}` or similar that would be needed to send things over http.

julianrubisch:
You beat me to it 😉 but there’s still work on the Ruby side of it... maybe I can tackle that tomorrow

jonathan-s:
@julianrubisch Yeah, I don't really have any ideas for what is needed (or what your plans were) on the ruby side. But since this is fully backward compatible it wouldn't affect anything if this went in.

julianrubisch:
Totally! I just added some annotations based on my experience, nothing critical though, afaict your implementation looks solid!

jonathan-s:
This still needs some work, so I'm going to take a look at this.

javascript/stimulus_reflex.js
julianrubisch: I think it's actually a good idea to tack it onto a global

I even also added the consumer:

```suggestion
  controller.StimulusReflex.subscription = subscription
  controller.StimulusReflex.consumer = adaptedConsumer
```

julianrubisch: you can then actually call it on line 61:

```
   return this.StimulusReflex.consumer.isConnected()
```

I guess you forgot that method you defined on the consumer

javascript/stimulus_reflex.js
```diff
@@ -115,6 +108,7 @@ const extendStimulusController = controller => {
 
       dispatchLifecycleEvent('before', element)
 
+      adaptedConsumer.send(JSON.stringify({ channel }, data, stimulateOptions))
       subscription.send(data)
```
julianrubisch, line 112: this line produces an error, because `subscription` isn't defined

javascript/stimulus_reflex.js
julianrubisch: I think you got the parens/braces wrong here... you'll want to send the channel identifier as it is, and the data as stringified JSON:

```suggestion
      this.StimulusReflex.consumer.send(
        channel,
        JSON.stringify(data),
        stimulateOptions
      )
```

javascript/stimulus_reflex.js
julianrubisch: I'd probably also rename `isActionCableConnectionOpen` to something more generic, `isConnectionOpen` for example

javascript/stimulus_reflex.js
jonathan-s: You're right, I got that wrong here. I do think that the `channel` needs to be stringified as well. I'll have to test it to see what works. Once I've tested it, it'd be good if someone tested on rails too, to make sure.

javascript/adapter.js
jonathan-s: That's an excellent answer. Why on earth is it marked as duplicate. At least I've got some SO powers so I voted for the question to be reopened.

javascript/consumer.js
```diff
@@ -4,7 +4,7 @@ function isConsumer (object) {
   if (object) {
     try {
       return (
-        object.constructor.name === 'Consumer' &&
+        object.constructor.name == 'ActionCableAdapter' &&
```
jonathan-s, line 7: The one thing I'm unsure about is this. We need to assume that the constructor name can be different than `ActionCableAdapter`. Is there a way to go around this @julianrubisch ?

jonathan-s:
It's worth noting that merging this in would break implementations that provide their own consumer (ie django-sockpuppet as an example). 

The standard case is to not provide your own consumer, and in that regard the above code is backwards compatible.

julianrubisch:
Yep well as already pointed out on discord, this PR(s) would necessitate rethinking package architecture a bit, I.e. making it pluggable. It’s easy to do on the ruby side, haven’t thought about JavaScript yet...

jonathan-s:
Let me know if you have any further ideas regarding package architecture @julianrubisch. From my point of view the above works for the time being, keen to hear your thoughts though.

julianrubisch:
It sure works! In ruby, we can swap adapters at config time. Specify a gem dependency „stimulus_reflex-messagebus“ and define the adapter in a config file. In JavaScript, we’d need to do a similar thing. However, that would first require a commitment to split up packages.

jonathan-s:
as a counter point, having batteries included is a nice feature as well.

julianrubisch:
I would certainly leave ActionCable as the default

javascript/consumer.js
jonathan-s: Also on line 27 in this file. We should consider returning the adapted consumer directly.

leastbad:
Closing this with sincere thanks, @jonathan-s, as the spirit of this PR will be fully realized in v4 but with a different implementation.

💙

PR #183, opened 2020-05-03
Introduce a transport adapter for cable_ready

julianrubisch:

Enhancement

## Description

A suggestion along https://github.com/hopsoft/stimulus_reflex/pull/182 to pull out some `cable_ready` dependent code into an agnostic transport adapter.

This is of course only a first step because for other transports like https://github.com/discourse/message_bus we'd need to wrap the whole `channel.rb` in an adapter. However, it adds a first layer of indirection and provides a clearer API to add new adapters that could implement `transmit_morphs` and `transmit_errors` differently.

## Checklist


julianrubisch:
@leastbad pointed out that everything from line 26 on in channel.rb could also be extracted. I could try that too in this PR, if you like

leastbad:
Closing this for spring cleaning / Marie Kondo reasons, but the spirit of this is already captured in v3.5 and will be fully realized in v4.

💙

PR #184, opened 2020-05-03
Replace camelize with homegrown version

jonathan-s:
Enhancement. 
## Description
In the spirit of #181 I thought it would be a good idea to remove the dependency of inflected as well, as that will reduce the bundle with a few more bytes. 

Compare with how it's done at inflected -> https://github.com/martinandert/inflected/blob/master/src/camelize.js

and rails -> https://apidock.com/rails/String/camelize

## Why should this be added
Bundle size gets smaller and we rely and fewer dependencies. 

## Checklist

### DIFF
--- a/javascript/lifecycle.js
+++ b/javascript/lifecycle.js
@@ -1,4 +1,19 @@
-import { camelize } from 'inflected'
+const camelize = (value, uppercaseFirstLetter = true) => {
+  if (typeof value !== 'string') return ''
+  str = str
+    .replace(/[\s_](.)/g, $1 => {
+      return $1.toUpperCase()
+    })
+    .replace(/[\s_]/g, '')
+    .replace(/^(.)/, $1 => {
+      return $1.toLowerCase()
+    })
+
+  if (uppercaseFirstLetter) {
+    str = str.substr(0, 1).toUpperCase() + str.substr(1)
+  }
+  return str
+}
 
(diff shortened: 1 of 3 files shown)


leastbad:
Since when does camelCase ever capitalize the first letter by default?

jonathan-s:
camelCase and camelize are slightly different. 

```
Loading development environment (Rails 6.0.2.2)
2.7.0 :001 > 'string_me'.camelize()
 => "StringMe"
2.7.0 :002 >
```

leastbad:
You are, of course, completely correct.

javascript/lifecycle.js
```diff
@@ -1,4 +1,18 @@
-import { camelize } from 'inflected'
+function camelize (str, uppercase_first_letter = true) {
```
hopsoft, line 1: I'm actually changing all function definitions to use function expressions so they're block scoped and a little more strict. Also since were in JavaScript, let's use `camelCase` for the argument. I think I may also prefer having guard clause on this one.

I'd prefer to see this defined like so.

```js
const camelize = (value, uppercaseFirstLetter = true) => {
  if (typeof value !== 'string') return ''
```

Also, let's use arrow functions in the internals.

jonathan-s: This should now be resolved.

PR #188, opened 2020-05-05
Allow to call stimulate without a reflex target

### LINKED ISSUE #179: Stimulate without a reflex target
# Feature Request

Turbolinks doesn't currently handle get requests from forms (https://github.com/turbolinks/turbolinks/issues/272). StimulusReflex handles this perfectly without the hacky workarounds, except that the `stimulate` JavaScript function requires a reflex target, but the Ruby side doesn't need to do anything except refresh the page.

In our specific case, have a `get` method search form. When the user clicks the search button, we use `history.pushState` to update the search query in the URL. Our users like to bookmark searches, so we need to be able to write real URLs instead of using session values.

We then call `stimulate` which rewrites the page. This works great, except in this case we've had to provide an empty Reflex target method.

```ruby
class NilReflex < ApplicationReflex
  def nothing
  end
end
```

```javascript
this.stimulate("NilReflex#nothing")
```

`stimulate` can take no arguments and still work as expected, re-rendering the current controller action and view. 

```javascript
// Ideal
this.stimulate()
```

## Additional context

Calling `stimulate` with no options currently raises this error:

```
Uncaught TypeError: Cannot read property 'split' of undefined
    at invokeLifecycleMethod (lifecycle.js:16)
    at HTMLDocument.<anonymous> (lifecycle.js:43)
    at dispatchLifecycleEvent (lifecycle.js:72)
    at Controller.stimulate (stimulus_reflex.js:114)
    at Controller._perform (project_search_controller.js:59)
    at invokeFunc (index.js:164)
    at trailingEdge (index.js:209)
    at timerExpired (index.js:197)
```

marcoroth:
# Enhancement

## Description

This PR allows to call the `stimulate()` method in a Stimulus Controller without to provide a reflex target.

### Setup

The basic setup looks like this:

```html
  <input type="text" data-action="keyup->example#example">
</div>
```

```javascript
// example_controller.js

  connect () {
    StimulusReflex.register(this)
  }

  example () {
    this.stimulate()
  }

}
```

It also works if you just specify an event in the `data-reflex` attribute on a html element without the need to call the `stimulate` method in a Stimulus Controller
```html
<input type="text" data-reflex="keyup">
```

Resolves #179

## Why should this be added

It makes it easier for any use case you just need to reload the page with a reflex. But with this you don't have to define an empty reflex on the server side to just reload the page. 

/cc @bbugh

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -88,7 +88,7 @@ const extendStimulusController = controller => {
     stimulate () {
       const url = location.href
       const args = Array.from(arguments)
-      const target = args.shift()
+      const target = args.shift() || 'StimulusReflex::Reflex#default_reflex'
       const element =
         args[0] && args[0].nodeType === Node.ELEMENT_NODE
           ? args.shift()
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -96,4 +96,8 @@ class StimulusReflex::Reflex
   def halted?
     !!@halted
   end
+
+  def default_reflex
+    # noop default reflex to force page reloads
+  end
 end


javascript/stimulus_reflex.js
```diff
@@ -78,7 +78,10 @@ const extendStimulusController = controller => {
     stimulate () {
       const url = location.href
       const args = Array.from(arguments)
-      const target = args.shift()
+      let target = args.shift()
+
+      if (!target) target = 'StimulusReflex::Reflex#default_reflex'
+
```
hopsoft, line 94: We could probably simplify to:

```js
const target = args.shift() || 'StimulusReflex::Reflex#default_reflex'
```

leastbad:
I'm thrilled to see this go from concept to real so quickly.

leastbad:
QQ: is the event in the `<input type="text" data-reflex="keyup">` example relevant? That is, can I capture any event? Does it have to be a browser native event or can it be a custom DOM event?

This is seriously awesome.

marcoroth:
@leastbad I don't know how it works with custom DOM events, but it does work for any native event. If the triggering with a custom DOM event worked before it probably also works here.

leastbad:
Legend.

bbugh:
Looks great! Thanks for doing this @marcoroth.

PR #192, opened 2020-05-08
2.2.3 - Use recognize_path_with_request instead of recognize_path

jasoncharnes:

Bug fix

We use route constraints and in version 2.2.3 `Rails.application.routes.recognize_path(path)` cannot find routes within a constraint. (This is a known issue from Rails)

## Why should this be added

This has been fixed in a later version. (Obviously, because I copied it from master 😎) This shouldn't go into master, but I wanted to open a PR to see if this is worth backporting to version 2 for those on Rails 5.2

## Checklist


jasoncharnes:
Opened on the advice of @andrewmcodes ✅

hopsoft:
I resolved conflicts but think it may not have produced the desired outcome. Let's discuss.

Issue #170, opened 2020-04-25
"Uncaught (in promise)" error after failed declarative reflex


leastbad:
# Bug Report

When a declared (eg. `data-reflex`) reflex fails on the server, the promise rejection generates an error.

## To Reproduce

Vanilla Rails 6 app w/SR + scaffolded resource. Declare a reflex that goes to a non-existant action in a reflex class eg:

```html
<h1 data-reflex="click->ApplicationReflex#meat">MEAT</h1>
```

This generates the following log output:

```
StimulusReflex::Channel Failed to invoke ApplicationReflex#meat! http://localhost:3000/posts/new undefined method `meat' for class `ApplicationReflex' /home/leastbad/.rvm/gems/ruby-2.6.5/gems/stimulus_reflex-3.1.3/lib/stimulus_reflex/channel.rb:53:in `method'
```

and the following error in the console:

![uncaught](https://user-images.githubusercontent.com/38150464/80281354-7445a280-86d8-11ea-86b1-eec27156ecac.png)

## Expected behavior

No errors in the console.

### StimulusReflex

- Gem: 3.1.3
- Node package: 3.1.3

### External tools

- Ruby: 2.6.5
- Rails: 6.0.0.2
- Node: 13.10.0

### Browser

- Browser: Chrome
- Version: 81

leastbad:
@hopsoft I'm trying to work through this https://discord.com/channels/629472241427415060/708177464894947362/710151681689518130

The question really is: assuming unnecessary scary errors are lazy APIs, should we / could we make an effort to determine whether to return a promise or not based on how `stimulate()` was called (via declared reflex or explicit JS)

leastbad:
Closed by #193

PR #193, opened 2020-05-09
Add halted lifecycle event

### LINKED ISSUE #190: Add a halted lifecycle event
# Feature Request

Currently clients cannot react to `halted` noop reflexes.

When a reflex callback chain is halted, the client is not made aware of this condition. This is because the lifecycle currently only supports:

- before
- success
- error
- after

This proposal would introduce a `halted` lifecycle event.

- before
- success
- error
- halted 👈 
- after

Adding a `halted` lifecycle event would allow clients to handle things properly and cleanup any UI tweaks made in the `before` event.

seb1441:
# Feature

## Description

Added the **halted** lifecycle event which gives the ability on client side to know when a reflex has been halted with the following ways:
- Listening on event `stimulus-reflex:halted`.
- Custom lifecycle method. Example for the `increment`reflex: `incrementHalted`.

### Note(update: Done)
The way `error` and `halted` work is pretty similar. They both come from the server side, broadcasted by CR.
- `error` will trigger the `stimulus-reflex:500` event which will then get picked up by the client side to dispatch the lifecycle event `stimulus-reflex:error`.
- `halted`, it's almost the same except that it doesn't retrieve an error and the event names are `stimulus-reflex:abort` and `stimulus-reflex:halted`.

Perhaps we should merge these together? Example, instead of having two different events `stimulus-reflex:500` and `stimulus-reflex:abort` to get messages from the server, we merge that into one event `stimulus-reflex:server` which will have an optional error. If it has an error with it, dispatch the `error` lifecycle, if not, `halted`.

What do you guys think?

## Why should this be added

Closes #190.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,10 +28,6 @@
 
 [Full Changelog](https://github.com/hopsoft/stimulus_reflex/compare/v3.2.0-pre0...v3.2.0.pre1)
 
-**Fixed bugs:**
-
-- Add guard to morph that checks stimulusReflex [\#191](https://github.com/hopsoft/stimulus_reflex/pull/191) ([hopsoft](https://github.com/hopsoft))
-
 ## [v3.2.0-pre0](https://github.com/hopsoft/stimulus_reflex/tree/v3.2.0-pre0) (2020-05-07)
 
 [Full Changelog](https://github.com/hopsoft/stimulus_reflex/compare/v3.2.0.pre0...v3.2.0-pre0)
@@ -52,6 +48,7 @@
 
 **Fixed bugs:**
 
+- Add guard to morph that checks stimulusReflex [\#191](https://github.com/hopsoft/stimulus_reflex/pull/191) ([hopsoft](https://github.com/hopsoft))
 -  Pluralize the generated class name, so that will match with the file name [\#178](https://github.com/hopsoft/stimulus_reflex/pull/178) ([darkrubyist](https://github.com/darkrubyist))
 
 **Closed issues:**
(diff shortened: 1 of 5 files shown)


hopsoft:
This is great. I can't wait to review and test. Thanks!

hopsoft:
@seb1441 

> Perhaps we should merge these together? 

👍  I like this idea. Please update the PR to include this refactor. 

Side note: I've never really cared for `stimulus-reflex:500` but it seemed like an obscure enough name that I wouldn't need to worry about future naming collisions.

seb1441:
@hopsoft Thanks for the feedback.
Added a commit which merges some error and halted related methods into a generic server message method.

I kept the previous behavior for errors, which is to reject promises. However, this is what's causing the error message in the console which @leastbad , from #170, would like to remove. Do we want both error and halted to resolve the promise?

leastbad:
Are we talking about renaming and/or slightly changing the behavior of an existing error? I guess this means a significant version bump?

seb1441:
@leastbad No, it won't change any current behavior.

seb1441:
@leastbad To be more specific, the event `stimulus-reflex:500` was only used internally to send the error from server side to client side but it was never mentioned in the docs. Renaming it should not affect any user.
Users will still be able to use `stimulus-reflex:error` event and callback, nothing will change there

leastbad:
When you and @hopsoft are ready to move forward with this, I'll quickly edit the docs to include client halted callbacks. 🥇

leastbad:
Huzzah!

PR #196, opened 2020-05-11
Messagebus exploration

julianrubisch:
# Proof of Concept

## Description

This is a Proof of Concept for discourse/message_bus as an alternative transport

I tested it against expo, to make this work you need to make the following changes there:

in application.rb, insert the message bus rack middleware:

```ruby
require "message_bus"
...
    config.middleware.use MessageBus::Rack::Middleware
```

in routes.rb, mount the StimulusReflex engine:
```ruby
  mount StimulusReflex::Engine, at: "stimulus_reflex"
```

I think this is valuable as a general exploration that alternative transports are possible, using https://github.com/hopsoft/stimulus_reflex/pull/183 and an adapted version of @jonathan-s 's https://github.com/hopsoft/stimulus_reflex/pull/182 where I fixed some issues.

Note that when I tested this, some features of expo (e.g. validation, book search) became glitchy both for message bus and action cable, while others (tabular, todo list) work. This might be configuration/timing issues of MessageBus, or some issues introduced in the extraction of adapter.js, or new bugs creeping in - at this point I cannot really tell. I would ask @jonathan-s to review his PR and double check that it works so I can then pull this in.

In my opinion, a HTTP transport would solve scalability considerations some users have uttered, while it is clear that probably MessageBus is not the right choice due to the hoops I had to jump through while making it talk to `cable_ready.js` properly.

Then, it occurred to me that this very JS part of CableReady is actually backend agnostic and calling anything called "Cable..." from an HTTP transport might be confusing.

So all in all, this PR is for you to explore, and a good point that the abstractions I chose in https://github.com/hopsoft/stimulus_reflex/pull/183 were correct.


leastbad:
I still look at this code frequently, but it's fair to say that it's now just an artifact that will inspire v4 greatly.

PR #200, opened 2020-05-18
Combine dataset with data-attributes from parent elements on reflex call

marcoroth:

Enhancement

## Description

This PR adds the functionality to combine all data attributes from all parents and includes them in the reflex call. 

To enable this you need to add the `data-reflex-dataset="combined"` attribute on the element with the `data-reflex` attribute. The dataset from the element and from all parents will be combined and sent to the backend. The combined data-attributes will then be available in the `dataset` of the element (see example below).

## Why should this be added

This helps to simplify and dry up the required html to call a reflex from the client side.

### Example with delecrative reflexes:

Before: 

```html
<div class="folder-details">
  <div class="action-details">
    <button data-folder-id="<%= folder.id %>" data-action-id="<%= action.id %>" data-whatever="1" data-reflex="click->FolderReflex#finish">Finish</button>
    <button data-folder-id="<%= folder.id %>" data-action-id="<%= action.id %>" data-whatever="1" data-reflex="click->FolderReflex#edit">Edit</button>
    <button data-folder-id="<%= folder.id %>" data-action-id="<%= action.id %>" data-whatever="1" data-reflex="click->FolderReflex#close">Close</button>
  </div>
</div>
```

After: 

```html
<div class="folder-details" data-folder-id="<%= folder.id %>">
  <div class="action-details" data-action-id="<%= action.id %>" data-whatever="1">
    <button data-reflex="click->FolderReflex#finish" data-reflex-dataset="combined">Finish</button>
    <button data-reflex="click->FolderReflex#edit" data-reflex-dataset="combined">Edit</button>
    <button data-reflex="click->FolderReflex#close" data-reflex-dataset="combined">Close</button>
  </div>
</div>
```

In the reflex on the server side the attributes can be accessed like before:

```ruby
class FolderReflex < ApplicationReflex
  def finish
    element.dataset["folder-id"]
  end
end
```

**Note:** If a data attribute appears several times, the deepest one in the DOM tree is taken. In the example below `data-id` would be `2`.

```html
<div data-id="1">
  <button data-id="2" data-reflex="ExampleReflex#whatever" data-reflex-dataset="combined">Click me</button>
</div>
```

### Example with `stimulate()`

The `data-reflex-dataset="combined"` attribute also works with the `stimulate()` function.

```html
<div data-folder-id="<%= folder.id %>" data-controller="folders">
  <button data-action="click->folders#edit" data-reflex-dataset="combined">Edit</button>
</div>
```

```javascript
// folders_controller.js

import StimulusReflex from 'stimulus_reflex';

  connect() {
    StimulusReflex.register(this)
  }

  edit(event) {
    this.stimulate("FolderReflex#edit", event.target)
  }
}
```

Thoughts? Let's discuss!

## Checklist


leastbad:
Interesting!

How is the parent determined? I assume that this is just for stimulate() calls and not data-reflex attrs?

marcoroth:
I keep traversing up the DOM tree with `element.parentElement` beginning from the element on which we strip out the current attributes. 

The current implementation is intended for the `data-reflex` attributes. I thought you anyway need to provide the attributes you want to pass to `stimulate()` yourself, so it wouldn't make sense to implicitly put them in there. But I'm open.

hopsoft:
I like this idea and agree that it can/will help to dry up several use cases. We'll need to be explicitly clear about how attribute conflicts are handled in the docs around this.

leastbad:
So it's basically the exact opposite of what I understood. 😆 
That'll teach me not to browse issues on my phone when I'm trying to fall asleep.

Conceptually, this seems really clever, but like Nate I worry about conflicts and generally scooping up far more than is intended. If I understand *better* now, this will keep scooping right up to the body element, no?

Just throwing it out there, but it might be good to make this feature opt-in by adding `data-reflex-scoop` to the element with the `data-reflex`, in the same way we add `data-reflex-permanent`.

And in the same vein, it could be useful to check those ancestors for a `data-reflex-scoop-break` or something to halt the scooping.

marcoroth:
> ... this will keep scooping right up to the body element, no?

That's right. This POC implementation naively goes up to the end of the DOM tree.


> ... but it might be good to make this feature opt-in by ...

I agree somewhat. On the one hand you don't want side effects and to break existing code but on the other hand you just add another attribute instead of the one you just moved up to the parent. 

In the following example you would even end up with more data-attributes than you had before, to "inherit" the data attribute from the parent element. But I think this is just an "issue" for the more simple examples. 

Before: 
```html
<div>
  <button data-example-id="<%= example.id %>" data-reflex="click->ExampleReflex#action1">Action 1</button>
  <button data-example-id="<%= example.id %>" data-reflex="click-> ExampleReflex#action2">Action 2</button>
</div>
```

After:
```html
<div data-example-id="<%= example.id %>">
  <button data-reflex-inherit data-reflex="click->ExampleReflex#action1">Action 1</button>       
  <button data-reflex-inherit data-reflex="click->ExampleReflex#action2">Action 2</button>
</div>
```

But you are probably right @leastbad. In the end it makes sense to opt-in for the "data attribute inheritance" you avoid confusion and that the developer should know what he's doing.

But I think the benefit is still here even with the extra attribute. Especially if you have a lot of different data attributes across your views or when they are nested and split up in different partials and/or components.

leastbad:
It is legitimately an interesting thought puzzle. The extra bytes are few. This is really just concern that someone could wonder why they are seeing what they perceive as "extra" attributes, and potential for confusion over conflict resolution.

The more I think about it, the more I'm coming around to it. :)

marcoroth:
Alright, I added the `data-reflex-inherit` attribute to opt-in for the parent data-attribute inheritance. Does the concept and the naming makes sense to you?

The push it a little bit further: What do you thing about the following situation:

```html
<div data-parent-id="123">
  <div data-reflex-inherit>
    <button data-reflex="click->ExampleReflex#do_something">Something</button>
    <button data-reflex="click->ExampleReflex#do_another_thing">Another thing</button>
  </div>
</div>
```

Should we treat `data-reflex-inherit` the same way we treat `data-reflex-permanent` and `data-reflex-root`? So if you define it on a parent element it also applies for its children? 
Like in the example above: should there be parent data-attribute inheritance for the two buttons?

leastbad:
I thought I got it, but your last example throws my brain off the cliff a little bit.

Is your intention to use `data-reflex-inherit` to say "this element, as well as any children of this element should recurse up the DOM tree to body, looking for data attributes to gobble up"?

My first instinct was that this is an awful lot of DOM hierarchy navigation every time someone clicks a data-reflex element; first you have to check your ancestors for any occurrence of `data-reflex-inherit`, then you have to do it again to pick up extra attributes. Then I remembered that computers are pretty fast, now.

Would you humor me and benchmark a realistic app (eg. lots of nested divs, like buttons on a Bootstrap card) with and without your `inherit` mechanism enabled?

marcoroth:
> Is your intention to use `data-reflex-inherit` to say "this element, as well as any children of this element should recurse up the DOM tree to body, looking for data attributes to gobble up"?

Exactly. Like you define the `data-reflex-root`. You say, starting from this element, I want all children to look up the data attributes from their parents.

> My first instinct was that this is an awful lot of DOM hierarchy navigation every time someone clicks a data-reflex element.

You are right. Maybe we can optimize that by annotating all implicit `data-reflex-inherit` in the same style we do it with the "conversion" from `data-reflex` to `data-action="stimulus-reflex#__perform"` in `setupDeclarativeReflexes`.

hopsoft:
Here's the data attributes API that I'd prefer to see.

```html
<div data-parent-id="123">
  <button data-reflex="click->ExampleReflex#do_something" data-reflex-inherit>Something</button>
  <button data-reflex="click->ExampleReflex#do_another_thing">Another thing</button>
</div>
```

In this example, the first button would inherit the `parent-id` but the second would not.

---

It may also make sense to expose inherited data to the Ruby side somewhere other than `element.dataset`. Something like, `element.inherited["parent-id"]`. Thoughts?

marcoroth:
@hopsoft Great. The current implementation works as you described in your example.

To be honest I'm not sure about `element.inherited[...]`. I think it makes the reflex code more obtrusive than it needs to be.  You would always need to check if want want to get the value 
 from `element.dataset["parent-id"]` or  `element.inherited["parent-id"]`. Do you see any benefit having them separated / any downside having them together? 

I could see that we provide the `inherited` additionally. `attributes` really just includes the attributes the element has and `dataset` contains all data-attributes from the element itself and the inherited ones. But somehow I feel like this makes it even more complex. I don't know.

marcoroth:
From the implementation point of view we would send `attributes` and `inherited` from the client.  As far as I understand `dataset` currently is anyway "just" an elegant accessor for the `attributes` hash.  Like this you can access just the `attributes`, just the `inherited` ones are access them from the combined `dataset` if you don't care where you attribute values is coming from.

hopsoft:
My primary concern is that we advertise the `element` exposed in the reflex as a representation of the DOM element. Merging inherited dataset attributes into the same dataset blurs this 1:1 relationship. What if we provided a helper method on `StimulusReflex::Element` to fetch a value? Maybe something like this:

```ruby
def get_attribute(name)
  attributes[name] || dataset[name] || inherited_dataset[name]
end
```

leastbad:
I'd say that, having slept on it, if we're going to add this functionality, let's not add artificial hoops for them to jump through to access the attributes. The 1:1 thing is pretty fuzzy anyhow; we just explain it properly in the docs, methinks. I kind of like @marcoroth's core idea that you pick up each new attribute at the lowest point it first shows up in the DOM. Then we give them a nice bucket of attributes, and if they don't want to use some of them, well, that's exactly what they already do today.

marcoroth:
Alright, I think I found a good tradeoff. The big difference now is, that we send an extra `dataset` in the reflex payload. If you now access `element.dataset` it is accessing the `@dataset` variable on the `Element` class instead of accessing and filtering the `attributes`.  

In my opinion this is has multiple advantages:
 * `element.attributes` just contains the attributes which are on the actual DOM element (as before)
* If you access`element.dataset` it just contains all the regular data-attributes which are on the DOM element itself (as before + the awesome improvements from #210) 
* If you now setup a `data-reflex-inherit` on the element, all the inherited attributes will now also be accessible on `element.dataset`. Like this the interface will stay the same for the developer. If you did setup the inheritance you know that you did that. In my opinion you then also want to access those inherited data-attributes in the same way. If you haven't setup the reflex inheritance there will be no difference (just that the inherited attributes aren't available  😉). 

I also tried out the other ideas (including the `element.inherited` and `get_attribute`) in my Boxdrop app, but I have to say it somehow feels not right in my opinion.  

I'm curious what you think about it.

javascript/attributes.js
```diff
@@ -62,6 +64,38 @@ export const extractElementAttributes = element => {
...
+// Extracts dataset of an element and includes inherited data attributes from parents if requested.
+//
+export const extractElementDataset = element => {
+  let attrs = extractDataAttributes(element) || {}
+
+  if (element.attributes[defaultSchema.reflexInheritAttribute]) {
+    let parent = element.parentElement
+
+    while (parent) {
+      attrs = { ...extractDataAttributes(parent), ...attrs }
```
marcoroth, line 79:
```suggestion
      attrs = { ...parent.dataset, ...attrs }
```
We could use `parent.dataset` here (and on line 73) instead of relying on the `extractDataAttributes()` function. But we would then need to convert the snakeCase key-names back to dasherized keys on the backend. 

But if we would do that, we could get rid of the `extractDataAttributes()` function.

hopsoft: I think it's OK to keep this computation/responsibility on the client.

lib/stimulus_reflex/element.rb
```diff
@@ -4,10 +4,11 @@ class StimulusReflex::Element < OpenStruct
...
   def initialize(attrs = {})
-    @attributes = HashWithIndifferentAccess.new(attrs || {})
-    @data_attributes = attributes.each_with_object(HashWithIndifferentAccess.new) { |(key, value), memo|
-      memo[key.delete_prefix("data-")] = value if key.start_with?("data-")
-    }
+    @attributes = HashWithIndifferentAccess.new(attrs.to_h["attrs"] || {})
+    @data_attributes = (attrs.to_h["dataset"] || {}).each_with_object(HashWithIndifferentAccess.new) { |(key, value), memo|
+      memo[key.delete_prefix("data-")] = value
+    }.freeze
+
```
hopsoft, line 11: Not sure the `to_h` call is necessary on `attrs`. Did you hit conditions where the argument wasn't a `Hash`?

marcoroth: Nope I didn't face a condition where it wasn't a Hash. It's just to be extra safe because it's now a two level deep Hash and before it was just one level since I now pass the whole `data` object in there.

But maybe we should rename that local variable in there from `attrs` to `data` since it's not just the `attrs` anymore.

javascript/attributes.js
```diff
@@ -62,6 +64,38 @@ export const extractElementAttributes = element => {
...
   }
+  return attrs
+}
+
+// Extracts dataset of an element and includes inherited data attributes from parents if requested.
+//
+export const extractElementDataset = element => {
+  let attrs = extractDataAttributes(element) || {}
+
+  if (element.attributes[defaultSchema.reflexInheritAttribute]) {
```
hopsoft, line 75: We can't use `defaultSchema` here because consumers of the library may change the name of this attribute.

We'll need to update to use `stimulusApplication.schema`. There are a few ways to accomplish this, but I'll let you experiment and find something that works well.

javascript/schema.js
```diff
@@ -1,5 +1,6 @@
 export const defaultSchema = {
   reflexAttribute: 'data-reflex',
   reflexPermanentAttribute: 'data-reflex-permanent',
-  reflexRootAttribute: 'data-reflex-root'
+  reflexRootAttribute: 'data-reflex-root',
+  reflexInheritAttribute: 'data-reflex-inherit'
```
hopsoft, line 5: I propose a rename of this new attribute given that all other attributes are noun based descriptors rather than verbs indicating behavior. 

`data-reflex-inherit` -> `data-reflex-dataset=combined`

javascript/attributes.js
marcoroth: TIL that you can rename the `data-reflex-*` attributes 😉

javascript/schema.js
marcoroth: Updated it to use `data-reflex-dataset=combined` 🙌

lib/stimulus_reflex/element.rb
```diff
@@ -3,11 +3,12 @@
...
-    @data_attributes = attributes.each_with_object(HashWithIndifferentAccess.new) { |(key, value), memo|
-      memo[key.delete_prefix("data-")] = value if key.start_with?("data-")
-    }
+  def initialize(data = {})
+    @attributes = HashWithIndifferentAccess.new(data["attrs"] || {})
+    @data_attributes = (data["dataset"] || {}).each_with_object(HashWithIndifferentAccess.new) { |(key, value), memo|
+      memo[key.delete_prefix("data-")] = value
+    }.freeze
+
     super attributes.merge(attributes.transform_keys(&:underscore))
```
hopsoft, line 12: I'm thinking that I'd prefer to include the combined dataset in the dot notation API available on the element. We'd need to change the call to `super`:

```ruby
super attributes.merge(attributes.merge(data_attributes).transform_keys(&:underscore))
```

marcoroth: Isn't this already happening on line 16?

Maybe we could move that inside to constructor and get rid of the `dataset` accessor method

hopsoft: No this is for constructing the `StimulusReflex::Element` as an `OpenStruct`. Note the call to `super`

Also, I prefer dataset get lazily initialized to help with performance if/when it isn't used.

PR #203, opened 2020-05-19
Map hashes in incoming arguments using with_indifferent_access

jaredcwhite:
# Enhancement

## Description

I've always been a stickler for wanting to use symbol keys in hashes and have usually been spoiled with Rails providing them. So when I noticed that hashes (or arrays containing hashes, etc.) coming into a reflex argument only provided string keys (right out of the JSON parser I'm guessing), I knew I needed to do something about that. 😄

If this is a potential performance concern, maybe we could put the map function behind a configuration option. Default true would be preferable!

## Checklist

### DIFF
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -23,6 +23,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     reflex_name, method_name = target.split("#")
     reflex_name = reflex_name.classify
     arguments = data["args"] || []
+    arguments.map! { |argument| map_hashes_in(argument) }
     element = StimulusReflex::Element.new(data["attrs"])
 
     begin
@@ -50,6 +51,13 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
   end
 
   private
+
+  def map_hashes_in(argument)
+    return argument.with_indifferent_access if argument.is_a?(Hash)
+
+    argument.map! { |nested_argument| map_hashes_in(nested_argument) } if argument.is_a?(Array)
+    argument
+  end
 
   def is_reflex?(reflex_class)
     reflex_class.ancestors.include? StimulusReflex::Reflex


lib/stimulus_reflex/channel.rb
```diff
@@ -51,6 +52,13 @@ def receive(data)
 
   private
 
+  def map_hashes_in(argument)
```
jaredcwhite, line 55: This could probably be renamed to something clearer…

hopsoft:
I agree with the sentiment about indifferent access, but am also trying to decide if this is something that should be handled by the library... for arguments.

leastbad:
I'm waffling back and forth on this. I think I'll abstain, because I don't end up using arguments very often.

jaredcwhite:
@leastbad @hopsoft Maybe I'm approaching this like a n00b, but I'm expecting JSON payloads would come in for most reflexes I write. How else do I access incoming data if not for arguments? (I know I can grab data attributes of element(s), but that feels more DOM/Stimulus-specific and less server-related.)

jaredcwhite:
Actually I'm looking at PR #204 now and I'm thinking that might take things in a better direction — use params instead of arguments for a more structured Rails-controller-params pattern. Only those params are being auto-serialized. I'll go make a comment there about providing optional client-side programmatic control over params.

hopsoft:
@jaredcwhite I'm leaning toward keeping explicit calls to `stimuluate` with arguments as is. _i.e. custom arguments should be handled by developers in their respective apps._

How do you feel about using mechanics from #204 to setup params _(with or without a form)_? Would that satisfy your desire to have Rails style params in your reflexes without the addition of this PR? See: https://github.com/hopsoft/stimulus_reflex/pull/204#issuecomment-632096254

jaredcwhite:
@hopsoft Yep, that works for me. 👍

andrewmcodes:
I totally support this. I think symbolizing the keys would be a great addition. So does @brascoder.

hopsoft:
I'm starting to come around on this, but may have some proposed refactoring for you @jaredcwhite. Keep an eye peeled.

lib/stimulus_reflex/channel.rb
```diff
@@ -23,6 +23,7 @@ def receive(data)
     reflex_name, method_name = target.split("#")
     reflex_name = reflex_name.classify
     arguments = data["args"] || []
+    arguments.map! { |argument| map_hashes_in(argument) }
```
hopsoft, line 26: Looks like `with_indifferent_access` will handle the nested values for us. Here's an example:

```ruby
{foo: 1, bar: [{baz: 2}]}.with_indifferent_access["bar"][0]["baz"] # => 2
```

I'm thinking we could simplify to this a one liner and get rid of the helper method.

```ruby
arguments.map! { |arg| arg.try(:with_indifferent_access) || nil }
```

Thoughts?

jaredcwhite: @hopsoft Well, I was trying to be too clever by half by mapping incoming arrays that might have hashes inside of them instead of assuming each argument is just a hash… If we do go this direction with the arguments, I think we should at least make the attempt to check things in a recursive fashion somehow.

hopsoft: Note that my solution doesn't assume every argument is a `Hash` because it's using ActiveSupport's `try` functionality. So if `with_indifferent_access` isn't supported by the argument, we simply return the arg itself.

The one use case this doesn't handle would be an array argument that contains Hashes. 🤔  I'm considering if we need to cover that use case.

jaredcwhite: That's what I was trying to do, yeah. Hash arg is easy, array arg means looping through looking for hashes.

leastbad: Arrays containing hashes are reasonably common. *goes back to lurking*

hopsoft: I'm sold. What about something like this:

```ruby
arguments.map! { |arg|
  arg.try(:with_indifferent_access) || arg.try(:map, &:with_indifferent_access) || arg
}
```

Too clever?

hopsoft:
Accidentally resolved my last comment. Here it is again.

---

I'm sold. What about something like this:

```ruby
arguments.map! { |arg|
  arg.try(:with_indifferent_access) ||
    arg.try(:map) { |a| a.try(:with_indifferent_access) || a } ||
    arg
}
```

Too clever?

jaredcwhite:
@hopsoft That is clever, but if there's a hash within an array within an array, it wouldn't get touched, so…I still think if we do it it should be fully recursive through nested arrays.

hopsoft:
Ah yes. Recursion it is then. Will approve and merge if you can address conflicts.

jaredcwhite:
🥳

PR #204, opened 2020-05-20
Form parameters

### LINKED ISSUE #199: ActionController Parameters
# Feature Request

The ability to access "params" inside a reflex.

I have a large form that has two requirements:

1. It needs to autosave the entire form after a field is changed
    * This includes misc reflexes that may do something unrelated to just persisting, but still needs to persist the form as a side-effect
2. It should go through the normal Rails validation process when saving
    * If it fails, I want to keep all the fields in the form on re-render (keep the bad data, show validations, keep instantiated nested has many, etc.)

Given the size of the form, and the requirements listed, I'd like to send all the form fields to Stimulus Reflex. I can do this using arguments, like so:

```erb
<% # app/views/posts/edit.html.erb %>
<%= form_with model: @post, data: { controller: "posts" } do |form| %>
  <%= form.hidden_field :id %>
  <%= form.text_field :name, data: { action: "change->posts#submitPost" } %>
<% end %>
```

```javascript
// app/javascript/controllers/posts_controller.js
get params() {
  // Using https://www.npmjs.com/package/form-serialize
  return serialize(this.element, { hash: true, empty: true });
}

submitPost(event) {
  this.stimulate("PostReflex#submit_post", event.target, this.params);
}
```

```ruby
# app/reflexes/post_reflex.rb
class PostReflex < ApplicationReflex
  def submit_post(params = {})
    @post = Post.find(params.dig('post', 'id')
    @post.assign_attributes(params.dig('post'))
    @post.save
    true # to always trigger a re-render
  end
end
```

However, after building the page, we end up with multiple reflexes that handle different separate, but similar requirements. 

Without assigning the post from the attributes each time, we would lose access to the form fields and could potentially (and our case, likely) trigger a re-render without the persisted data.

A benefit to this approach is that for a "small amount" of code, and by sending the attributes each time, when we attempt to autosave we implicitly get `#validate`. When StimulusReflex re-renders we have error messages and css for free, and we didn't lose data! 🎉 

But, this means we need to repeat the 1. find, 2. assign, 3. save process for each reflex. I'd like to extract that into before and after callbacks but, to my knowledge, we can't access arguments in callbacks.

@hopsoft showed me how he handles sending a form with parameters to a reflex, stores the attributes in the session, and re-instantiates the model with the params. (It's really cool!)

Unfortunately, we use cookie store. While it's on the docket to potentially change to a database session store in the future, we're not there yet. That means, we don't have access to the session. We also don't want to use Rails cache, because we don't want to lose the data due to eviction (we have a pretty large cache in terms of available size, but also in terms of how much it's used). 

This leaves us with a fair amount of duplication for sending form data and attempting to persist:
- Having to take params as an argument for **each** reflex
- Having to find the record
- Performing the "reflex" (if it's not just an autosave)
- Attempting to save the record

The ability to send params to a `Reflex` and have it available as an `attr_reader` inside the reflex so it's available in callbacks.

* Send params from StimulusReflex JS
* Instantiate params as `ActionController::Parameters` instance
* Pass params instance to the reflex constructor
* Make params available `attr_reader` inside a reflex

Here's how our reflex looks (code changed to "posts") at work after implementing `params` in our fork of StimulusReflex:

```erb
<% # app/views/posts/edit.html.erb %>
<%= form_with model: @post, data: { controller: "posts" } do |form| %>
  <%= form.hidden_field :id %>
  <%= form.text_field :name, data: { action: "change->posts#submitPost" } %>
  <%= form.fields_for :categories, @post.categories do |category_form| %>
    ... form for adding a new category while creating a new post
  <% end 
<% end %>
```

```javascript
// app/javascript/controllers/posts_controller.js

// Sent implicitly with `this.stimulate` as to not override the existing argument functionality
get params() {
  // Using https://www.npmjs.com/package/form-serialize
  return serialize(this.element, { hash: true, empty: true });
}

submitPost(event) {
  this.stimulate("PostReflex#submit_post", event.target);
}
```

```ruby
# app/reflexes/post_reflex.rb
class PostReflex < ApplicationReflex
  before_reflex -> { @post = Post.find(params.dig(:post, :id)) }
  before_reflex -> { @post.assign_attributes(post_params) }
  after_reflex -> { @post.save; true }, except: :build_new_category

  # In this case, all our functionality is handled by callbacks
  def submit_post; end

  # If using nested attributes, the field will be rendered when reflex re-renders
  # By passing params each time, we don't have to keep "state" of categories
  # because params is our "source of truth" and also handles validations.
  def build_new_category
    @post.categories.build
  end

  private

  def post_params
    params.require(:post).permit(:name, categories_attributes: [:id, :name])
  end
end
```

What I like about this approach:
- It feels like a Rails controller
  - `params` are a familiar concept
- The form functions like a server-rendered/submitted Rails form
  - If I "click submit" (autosave) and there's an error, I don't lose any data and I get errors
  - If I refresh with unsaved data (an error occurred) it doesn't persist the data

What I don't like about this approach:
- It could be expensive to always re-render the page (though I'm worried about the amount of code we'll end up with otherwise)
- I don't love that `get params()` is sent implicitly. @excid3 had an idea for `stimulateWithParams` that I'd be glad to implement, as well
- That I might be misunderstanding Reflex 😏

## NOTESSSSSS

I'm still new to StimulusReflex, so I may be missing the bigger picture. This is the result of a 40-hour week of thinking about my given problem.

* If this isn't functionality that Reflex is interested in taking on, I completely understand ❤️ 
* There ~may be~ is probably a better way to implement either:
  * My problem/requrements
  * The proposed solution
* If _there is_ something here, I'm **more than glad** to do the work. Like I said, we have this POC working in our app and I'd be glad to upstream/adjust what we've done
* If you want to see our implementation, you can [view it here](https://github.com/podia/stimulus_reflex)

I'd love to hear from others on:
1. How they'd solve this
2. If they could/would use this proposal

Sorry for the big issue. I wanted to explain in detail what I was looking at/considering. I'm very appreciative for anyone taking the time to read this and for all of you who have contributed to make StimulusReflex awesome. ❤️ 

CC @excid3 @basilkhan05 @ideasasylum

jasoncharnes:
# Enhancement

## Description

After discussion with @hopsoft I have a PR for (potentially) introducing `params` to StimulusReflex.

This code:
* Sends `params` to StimulusReflex from JavaScript
  * Adds the [`form-serialize`](https://www.npmjs.com/package/form-serialize) package from NPM which has 0 dependencies and is less than ~300kb~ [1kb](https://bundlephobia.com/result?p=form-serialize@0.7.2) This allows us to serialize a form (including nested parameters) into a hash with one line.
* Casts and exposes `params` to `StimulusReflex::Reflex` as a `ActionController::Parameters` instance

Here's what a reflex can do with the new params:

```ruby
class Post < ApplicationRecord
  validates :name, presence: true
  has_many :categories
  accepts_nested_attributes_for :categories
end

class Category < ApplicationRecord
  validates :name, presence: true
  belongs_to :post
end
```

```ruby
class PostsController < ApplicationController
  def edit
    # Memoizing means the instance set in the reflex (from params) will be reused
    # when the page re-renders from SR, this is an important part of the magic.
    @post ||= Post.find(params[:id])
  end
end
```

```erb
<%= form_with model: @post, data: { reflex: "submit->PostReflex#submit", signed_id: @post.to_sgid.to_s } do |form| %>
  <% if @post.errors.any? %>
    <!-- keep in mind, if you have error CSS classes defined, they will automatically pick up, too -->
    <% @post.errors.full_messages.each do |message| %>
      <li><%= message %>
    <% end %>
  <% end %>
  <div>
    <%= f.label :name %>
    <!-- changing this field will trigger the autosave
            if the field is changed while you have invalid categories, 
            they will stick around on re-render, but will show validation errors 
            because of accepts_nested_attributes and the category validations --> 
    <%= f.text_field :name, data: { reflex: "change->PostReflex#submit" } %>
  </div>
  <%= form.fields_for :categories, @post.categories do |category_form| %>
    <%= category_form.hidden :id %>
    <%= category_form.label :name,  %>
    <!-- changing this doesnt autosave, it's the responibility of either the submit button
            or the autosave that happens on name above.
            this is a bit silly of an example I've made, but it's to show how both autosaving
            and regular "form submission" can work together. -->
    <%= category_form.text_field :name
  <% end %>
  <%= link_to "New category", "#", data: { reflex: "click->PostReflex#build_category" } %>
  <%= form.submit %>
<% end %>
```

```ruby
class PostReflex < ApplicationReflex
  before_reflex do
    @post = GlobalID::Locator.locate_signed(element.dataset.dig("signed-id")
    @post.assign_attributes(post_params)
  end

  def submit
    @post.save
  end

  # Triggers re-render with new category.
  # Since we're building the object from params in the callback above, we can
  # click this as many times as we want to keep getting new categories.
  def build_category
    @post.categories.build
  end

  private

  def post_params
    params.require(:post).permit(:name, categories_attributes: [:id, :name])
  end
end
```

Fixes #199 

## Why should this be added

This has allowed us, at Podia, to take an edit form that requires auto-saving a model that `has_many` nested attributes and make it interactive with only **8 lines** of JavaScript.

We treat this form like a plain old Rails form (PORF™) by building our form like it's server-rendered, but stimulating it with StimulusReflex.

Adding params to the Reflex allows us to work with forms in StimulusReflex as first-class citizens, and feel (not quite 100%) like working with a standard server-rendered Rails form.

The benefits we've seen so far:
* Autosaving is as simple as adding `data: { reflex: "change->Reflex#update" }` to each field that should autosave (since the field is inside the `<form>` all the params are automatically serialized and send to StimulusReflex)
  * For any "JavaScript sprinkles" we need, we can utilize lifecycle methods in Stimulus
* Because we've defined our error styling, we get form errors for free
  * We instantiate an object from params, if it fails to save that object is still assigned to the instance variable and when the page re-renders errors are just there
* We have access to `params` in callbacks (`before_reflex`, `after_reflex`) that allow us to make decisions/instantiate objects from params
* Params are as accessible from the Reflex as from a controller
  * `params.require(:post).permit(:name, categories_attributes: [:id, :_destroy, :name])`
* Building a new record for a nested `has_many` association requires no JavaScript
  * Let me say that again, **no JavaScript** 😱
  * We just call a reflex that builds the new record: `@post.categories.build`
  * Because we have a Rails form that knows about `fields_for :categories, @post.categories` the re-render populates the empty form field
  * BUT THAT'S NOT ALL! Because we have access to `params`, and we instantiate an object from `params` each time we don't lose the built record if saving the record fails, it re-renders just like it would if a controller action went to update/save the record and had to bail out with error messages

It packs a lot of punch for us, with a "minimal" amount of code (in terms of implementation as well as code added to StimulusReflex)

I've been through several iterations of this, and after pairing with @hopsoft today, this is the proposal we've come up with.

I'm glad to answer any questions/hear feedback!

❤️ 

## Checklist

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "@rails/actioncable": ">= 6.0",
     "cable_ready": ">= 4.1.2",
-    "form-serialize": "^0.7.2",
+    "form-serialize": ">= 0.7.2",
     "stimulus": ">= 1.1"
   },
   "devDependencies": {
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -36,6 +36,7 @@
   "dependencies": {
     "@rails/actioncable": ">= 6.0",
     "cable_ready": ">= 4.1.2",
+    "form-serialize": "^0.7.2",
     "stimulus": ">= 1.1"
   },
   "devDependencies": {
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -78,7 +78,7 @@ class StimulusReflex::Reflex
       )
       path_params = Rails.application.routes.recognize_path_with_request(req, url, req.env[:extras] || {})
       req.env.merge(ActionDispatch::Http::Parameters::PARAMETERS_KEY => path_params)
-      req.env.merge!("action_dispatch.request.parameters" => @params)
+      req.env["action_dispatch.request.parameters"] = @params
       req.tap { |r| r.session.send :load! }
     end
   end
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class StimulusReflex::Reflex
+  include ActionController::StrongParameters
   include ActiveSupport::Rescuable
   include ActiveSupport::Callbacks
 
@@ -42,7 +43,7 @@ class StimulusReflex::Reflex
     end
   end
 
-  attr_reader :channel, :url, :element, :selectors, :method_name, :params
+  attr_reader :channel, :url, :element, :selectors, :method_name
 
   delegate :connection, to: :channel
   delegate :session, to: :request
@@ -53,7 +54,7 @@ class StimulusReflex::Reflex
     @element = element
     @selectors = selectors
     @method_name = method_name
-    @params = ActionController::Parameters.new(params)
+    @params = params
   end
 
   def request
@@ -77,6 +78,7 @@ class StimulusReflex::Reflex
       )
       path_params = Rails.application.routes.recognize_path_with_request(req, url, req.env[:extras] || {})
       req.env.merge(ActionDispatch::Http::Parameters::PARAMETERS_KEY => path_params)
+      req.env.merge!("action_dispatch.request.parameters" => @params)
       req.tap { |r| r.session.send :load! }
     end
   end
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -42,17 +42,18 @@ class StimulusReflex::Reflex
     end
   end
 
-  attr_reader :channel, :url, :element, :selectors, :method_name
+  attr_reader :channel, :url, :element, :selectors, :method_name, :params
 
   delegate :connection, to: :channel
   delegate :session, to: :request
 
-  def initialize(channel, url: nil, element: nil, selectors: [], method_name: nil)
+  def initialize(channel, url: nil, element: nil, selectors: [], method_name: nil, params: {})
     @channel = channel
     @url = url
     @element = element
     @selectors = selectors
     @method_name = method_name
+    @params = ActionController::Parameters.new(params)
   end
 
   def request
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class StimulusReflex::Reflex
-  include ActionController::StrongParameters
   include ActiveSupport::Rescuable
   include ActiveSupport::Callbacks
 
@@ -105,5 +104,9 @@ class StimulusReflex::Reflex
 
   def default_reflex
     # noop default reflex to force page reloads
+  end
+
+  def params
+    @_params ||= ActionController::Parameters.new(request.parameters)
   end
 end
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -103,7 +103,10 @@ const extendStimulusController = controller => {
         permanent_attribute_name:
           stimulusApplication.schema.reflexPermanentAttribute,
         reflexId: reflexId,
-        params: serializeForm(element.closest("form"), { hash: true, empty: true })
+        params: serializeForm(element.closest('form'), {
+          hash: true,
+          empty: true
+        })
       }
       const { subscription } = this.StimulusReflex
       const { connection } = subscription.consumer
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -1,5 +1,6 @@
 import { Controller } from 'stimulus'
 import CableReady from 'cable_ready'
+import serializeForm from 'form-serialize'
 import { defaultSchema } from './schema'
 import { getConsumer } from './consumer'
 import { dispatchLifecycleEvent } from './lifecycle'
@@ -101,7 +102,8 @@ const extendStimulusController = controller => {
         selectors,
         permanent_attribute_name:
           stimulusApplication.schema.reflexPermanentAttribute,
-        reflexId: reflexId
+        reflexId: reflexId,
+        params: serializeForm(element.closest("form"), { hash: true, empty: true })
       }
       const { subscription } = this.StimulusReflex
       const { connection } = subscription.consumer
--- a/javascript/yarn.lock
+++ b/javascript/yarn.lock
@@ -2006,6 +2006,11 @@ form-data@~2.3.2:
     combined-stream "^1.0.6"
     mime-types "^2.1.12"
 
+form-serialize@^0.7.2:
+  version "0.7.2"
+  resolved "https://registry.yarnpkg.com/form-serialize/-/form-serialize-0.7.2.tgz#b0a2ff0c22026fb6d3d15c9d33f6de6a432e4732"
+  integrity sha1-sKL/DCICb7bT0VydM/beakMuRzI=
+
 fs.realpath@^1.0.0:
   version "1.0.0"
   resolved "https://registry.yarnpkg.com/fs.realpath/-/fs.realpath-1.0.0.tgz#1504ad2523158caa40db4a2787cb01411994ea4f"
--- a/lib/generators/templates/custom_reflex.rb
+++ b/lib/generators/templates/custom_reflex.rb
@@ -11,6 +11,7 @@ class <%= class_name %>Reflex < ApplicationReflex
   #   - session - the ActionDispatch::Session store for the current visitor
   #   - url - the URL of the page that triggered the reflex
   #   - element - a Hash like object that represents the HTML element that triggered the reflex
+  #   - params - parameters from the element's closest form (if any)
   #
   # Example:
   #
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -24,11 +24,12 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     reflex_name = reflex_name.classify
     arguments = data["args"] || []
     element = StimulusReflex::Element.new(data["attrs"])
+    params = data["params"] || {}
 
     begin
       reflex_class = reflex_name.constantize
       raise ArgumentError.new("#{reflex_name} is not a StimulusReflex::Reflex") unless is_reflex?(reflex_class)
-      reflex = reflex_class.new(self, url: url, element: element, selectors: selectors, method_name: method_name)
+      reflex = reflex_class.new(self, url: url, element: element, selectors: selectors, method_name: method_name, params: params)
       delegate_call_to_reflex reflex, method_name, arguments
     rescue => invoke_error
       reflex.rescue_with_handler(invoke_error)


hopsoft:
Look like form-serialize is much smaller than noted above. Less than 1k gzipped & minified. https://bundlephobia.com/result?p=form-serialize@0.7.2

hopsoft:
This is an exciting PR as it promotes forms to first class status in the library. It treats form **state** similar to a standard HTTP request _which is incredibly robust_. This should help people using cookie based sessions that also want to create highly dynamic and complex forms. In fact, I will likely port some of my existing projects to use this technique instead of relying on session/cache/db based **state** storage. 

I'm looking forward to hearing what others think about it.

jonathan-s:
I'd say that this looks solid. At first I was a bit taken a back by the claim that the dependency was smaller than 300kb (which is a lot). Though checking the dependency as hopsoft mentioned, it's very small. 

For what this does it seems like a great improvement.

jasoncharnes:
Haha I meant to say 3k, but 1k is ever better

julianrubisch:
wonderful. I found myself needing this _today_.

now pardon my ignorance, but don't we have `request.params` in our hands in the reflex? why serialize the form a second time? I'm sure I'm missing something

julianrubisch:
ah, cause it's not your normal `POST`, the form params won't be included, right?

jasoncharnes:
> ah, cause it's not your normal POST, the form params won't be included, right?

Exactly! The only "params" passed, currently, are method arguments and an element's info (value, dataset, etc)

hopsoft:
@julianrubisch Makes a great point here though. I wonder if we should keep form params in a private instance variable and then merge them into the `request.params` when it gets instantiated. That feels a bit more Railsy to me. We could also `delegate :params, to: :request`. Thoughts?

leastbad:
@jasoncharnes this is genuinely impressive; I can't currently fathom why this wouldn't make the world better. I am genuinely wowed by how little code this requires. Also, I am relieved to encounter someone similarly unsparing with their prose as I am on this project. Everyone else here communicates like they are paying by the two-byte pair. 😁 

I do have some questions and requests, but it's all housekeeping / explain it to this simple country lawyer stuff.

1. If a reflex gets called on a form element or one of its children - not just elements but buttons and divs and anything else in the hierarchy - then the params for the whole form will be serialized and passed along as a params accessor. A reflex triggered by an element not inside of a form will have a params accessor that returns an empty `ActionController::Parameters`. Right?

2. When you declare a data-reflex, I see you using "Post" instead of "PostReflex". Is that a simple omission? (If so, let's update the PR text)

3. You have two before_reflex callbacks; is there a reason you couldn't just do it all in one callback? (If so...)

4. Is there a functional difference between using a lambda and a block syntax callback?

5. I'm feeling a bit thick trying to figure out how the input element connected to Post#name doesn't get wiped out by reflex updates, unless there's an implicit autosave **required** on every element. If a change event handler is mandatory, let's update the PR text.

6. Can you please add an update action to the reflex with at least one change event reflex declared in the markup to demonstrate the autosave you envision? Pretend we have no capacity for imagination. Will autosave kick off validations for the entire model? I'm trying to figure out whether this would allow someone to, for example, enter a value to an autosaving text on an unsaved model that has multiple `validates presence: true` without showing error messages for text inputs they haven't reached yet.

7. Can you please add an example of what minimalist error handling would look like in the markup (and Reflex class, if applicable).

jasoncharnes:
> I wonder if we should keep form params in a private instance variable and then merge them into the `request.params` when it gets instantiated. 

@hopsoft @julianrubisch This is interesting! I'd be glad to do that. I think it makes sense for the request to own the `params`. I can update the PR accordingly. 

Do you think we should delegate params to request in StimulusReflex, or leave that up to the user?

--- 

@leastbad thanks for the encouragement! I've been pretty reluctant/nervous to commit to open source in the past and this has been a welcoming experience. 🤗 

> If a reflex gets called on a form element or one of its children - not just elements but buttons and divs and anything else in the hierarchy - then the params for the whole form will be serialized and passed along as a params accessor. A reflex triggered by an element not inside of a form will have a params accessor that returns an empty ActionController::Parameters. Right?

Correct. `form-serialize` will return an empty JavaScript object, which by the time it's reached the Reflex will be an empty hash, which is instantiated as an empty `ActionController::Parameters` instance.

> When you declare a data-reflex, I see you using "Post" instead of "PostReflex". Is that a simple omission? (If so, let's update the PR text)

That is the late-night brain 😎 I will fix that in the PR text.

> You have two before_reflex callbacks; is there a reason you couldn't just do it all in one callback? (If so...) Is there a functional difference between using a lambda and a block syntax callback?

There's no specific reason I have it split up. It could all be done in a single callback. 👍 Regarding the difference in lambda/block- I'm unaware (read, ignorant) of the functional differences in this scenario. I tend to reach for lambdas in callbacks out of habit, but for not specific reason.

(I'll probably research this now, though 😅)

> I'm feeling a bit thick trying to figure out how the input element connected to Post#name doesn't get wiped out by reflex updates, unless there's an implicit autosave required on every element. If a change event handler is mandatory, let's update the PR text.

Not if the reflex happens _within_ the form. The params will always be sent and as long as you're instantiating an object/using `assign_attributes` with the params (and your controller memoizes `@post ||= ` instead of `@post =`) you shouldn't lose any data. (Am I misunderstanding the question?)

This brings up a good point, too. 

If you fire a reflex event _outside_ of the form, then it _would_ lose the attributes and wipe out the form. This is an interesting case I haven't considered, because all my reflexes happen (currently) inside my form at work. 

Do you have any ideas on how to mitigate this? Or I pre-maturely trying to solve a problem that doesn't exist yet?

> Can you please add an update action to the reflex with at least one change event reflex declared in the markup to demonstrate the autosave you envision? Pretend we have no capacity for imagination. Will autosave kick off validations for the entire model? I'm trying to figure out whether this would allow someone to, for example, enter a value to an autosaving text on an unsaved model that has multiple validates presence: true without showing error messages for text inputs they haven't reached yet.

I'm glad to update the text with that example. 

Autosave _will_ kick off validations for the entire model. In the case of our edit form at work, this is okay for our use case, because the form should always be in a valid state. In the case of a form for a new object, autosaving may be undesirable.

> Can you please add an example of what minimalist error handling would look like in the markup (and Reflex class, if applicable).

Sure thing!

---

Let me know if I missed anything/didn't answer a question. Thanks for the thoughtful feedback and responses.

jasoncharnes:
I updated the examples in the PR body to include:

- What the models might look like (validations, nested attributes)
- How the controller would look
- An autosave example on name
- Some error messages in the form
- Contextual comments to explain what's happening

leastbad:
Thanks for filling in the blanks, @jasoncharnes. I'm going to stare at this for a while.

FWIW, I actually kind of like the lambda syntax for a single-liner, but I do feel like it's probably more idiomatic to use a do..end or pass a symbol. I was just curious if it conveyed some additional magic!

Speaking of highly subjective: if you're using the same Reflex action (submit) to autosave, you might likely settle on instead calling that action save. 🤷 

I suspect that there's no practical general solution to the issue of reflexes called outside the form wiping your values........ unless there was some mechanism by which you could use `form-serialize` (or something like it) to actually scoop up all of the input element state for your entire DOM. That might be inception-level bonkers, but fun to think about.

Closer to earth, it really comes down to designing your UX around these principles. It's very possible that the "[hybrid reflex](https://discordapp.com/channels/629472241427415060/629472241427415062/712045568729546793)" process I described would allow people to build up their UIs, thinking of them as components instead of full pages.

hopsoft:
> Do you think we should delegate params to request in StimulusReflex, or leave that up to the user? -@jasoncharnes 

I think we should delegate so the reflex feels similar to a standard Rails controller.

jasoncharnes:
@hopsoft I updated the code to move params into the request. 

However, in a controller the `ActionController::Parameter` instantiation happens in a module that's included into `ActionController::Base`. This means that in order to get a `Parameters` object, I have to take the params back out of the request and instantiate them. 

I'm not sure that's much cleaner, though. 🤔 Also, I could just suck at Rails internals. 😎 It would be neat if putting the params into the request _automatically_ cast it, but I can't see that it does.

Let me know if I'm missing something, here.

leastbad:
There are some deeply weird things that we have to do to make the context which exists during a Reflex behave the way people would reasonably expect it to. In actual fact, we use the term "bubble universe" more frequently than I am comfortable because we have to be on our toes that, for example, when someone sets a session variable, it's actually set for the application and not just a mockery of a session variable that only exists for the duration of the Reflex.

[](http://www.youtube.com/watch?v=pu8JjjESu-I "")

jaredcwhite:
This looks really cool @jasoncharnes! I'm wondering if we can expose the params on the client-side so (optionally) they can be programmatically constructed, rather than default to a form serialization under the hood. The main reason I ask is because I could then send JSON payloads via params rather than arguments per se, and my #203 PR becomes unnecessary. What do you think about that?

hopsoft:
Interesting idea @jaredcwhite. Perhaps a custom event or callback that could be used to modify the serialized params prior to invoking ActionCable's `send`. 

```js
document.addEventListener('stimulus-reflex:finalize-params', event => {
  const { params } = event.detail
  params.foo = "bar"
})
```

Maybe @jasoncharnes and I could pair on this tonight or tomorrow. 🤞

pablo-co:
I have had both use cases, reflexes where the whole data contained within the form is needed, and cases where there is some reflex global state (IDs, value of some variable or field, etc) needs to be available on all reflexes. I have solved both through serialization of data in `element.dataset`.

hopsoft:
💡 I just realized that we already have a lifecycle hook that can be used to modify the `params` before we `send` data over the wire to the server. Here are some examples of how to do it.

```js
document.addEventListener('stimulus-reflex:before', event => {
  const params = event.target.reflexData
  event.target.reflexData.params = { ...params, foo: true, bar: false }
})
```

```js
  beforeReflex(element) {
    const { params } = element.reflexData
    element.reflexData.params = { ...params, foo: true, bar: false }
  }
}
```

@leastbad We'll want to add a section to the docs on **form params serialization** that describes how to modify params before sending.

lib/stimulus_reflex/reflex.rb
```diff
@@ -76,6 +77,7 @@ def request
       )
       path_params = Rails.application.routes.recognize_path_with_request(req, url, req.env[:extras] || {})
       req.env.merge(ActionDispatch::Http::Parameters::PARAMETERS_KEY => path_params)
+      req.env["action_dispatch.request.parameters"] = @params
```
hopsoft, line 80: What happens if we modify to this?

```ruby
req.env.merge(ActionDispatch::Http::Parameters::PARAMETERS_KEY => path_params.merge(@params))
```

lib/stimulus_reflex/reflex.rb
```diff
@@ -103,4 +105,8 @@ def halted?
   def default_reflex
     # noop default reflex to force page reloads
   end
+
+  def params
+    @_params ||= ActionController::Parameters.new(request.parameters)
+  end
```
hopsoft, line 111: I'd prefer to see this implemented as:

```ruby
delegate :params, to: :request
```

lib/stimulus_reflex/reflex.rb
jasoncharnes: This was the first thing I tried yesterday. Because of how [Rails works under the hood](https://github.com/rails/rails/blob/777595212e5bda9f804da7cbb46d5775918f3457/actionpack/lib/action_dispatch/http/parameters.rb#L66), this will only assign the `path_parameters` and ignore the `request_parameters`.

This leaves us with just path parameters when all is said and done, hence the extra env assignment.

lib/stimulus_reflex/reflex.rb
jasoncharnes: I'm down to do that. 

Where would we get the `ActionController::Parameters` instance, then? My understanding of ActionController is that a request isn't responsible for the instantiation of strong params. I believe it's a module that's included in `ActionController::Base` that reads the params out of the request. The method here is actually from that module.

hopsoft: I'm expecting Rails internals to set that up for us. What happens when if we delegate? Is `params` and instance of `ActionController::Parameters`?

lib/stimulus_reflex/reflex.rb
hopsoft: Makes sense.

lib/stimulus_reflex/reflex.rb
hopsoft: I finally caught up to you on this and can see that you simply implemented what the controller is doing. This makes sense to me now.

javascript/package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "@rails/actioncable": ">= 6.0",
     "cable_ready": ">= 4.1.2",
+    "form-serialize": "^0.7.2",
```
hopsoft, line 39: I wonder if we should loosen this dependency up a bit. `"form-serialize": ">= 0.7.2"` Thoughts?

leastbad:
@jasoncharnes thank you for this amazing contribution to the library. #epic

PR #205, opened 2020-05-20
Loosen Rails requirement to 5.2 with instructions

jasoncharnes:
# Um... Enhancement? Maybe?

## Description

We're dinosaurs at Podia running on Rails 5.2. 👀 Nate helped us "shove" StimulusReflex into our application. Except, there was no shoving involved. We only had to loosen the requirement on Rails down to 5.2 and use the new `@rails/actioncable` npm package, instead of `actioncable`.

This PR loosens the requirement and adds instructions in the docs on how to upgrade your ActionCable JavaScript code, based on our experience.

## Why should this be added

I believe allowing Rails 5.2 apps into ReflexLand could help adoption from companies like ours. We're working hard on our Rails 6 upgrade, but we're not there yet.

Without this, we have to run a fork of StimulusReflex- which means:
* Pointing bundler to our GH repo for StimulusReflex
* Hosting another repo just for the JavaScript that we can point yarn to
  * The `javascript` folder is nested and I can't get yarn to point to a nested folder in a git repo 😅 

## Help!

I'm not 100% sure that our positive experience will translate to others running Rails 5.2. Do you see any reason why this _wouldn't_ work in a Rails 5.2 app?

## Checklist

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -5,7 +5,7 @@ description: How to prepare your app to use StimulusReflex
 # Setup
 
 {% hint style="warning" %}
-StimulusReflex v3 has been released, and there are some big changes. **Rails 6+ and server-side session storage are now required.**
+StimulusReflex v3 has been released, and there are some big changes. **Server-side session storage is now required.**
 
 You can find additional information for supporting Rails 5.2+ below.
 {% endhint %}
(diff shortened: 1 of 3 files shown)


docs/setup.md
```diff
@@ -117,16 +117,16 @@ StimulusReflex supports both client and server logging of Reflexes.
...
-```ruby
-bundle add stimulus_reflex --version "~> 2.2.3"
-yarn add stimulus_reflex@2.2.3
-```
-
-While we don't have the resources to maintain two distinct package versions, we're proud of 2.2.x and consider it stable. In the unfortunate case of a critical security issue, we will make every attempt to backport hotfixes.
+1. Replace `actioncable` with `@rails/actioncable` in `package.json`
+  * `yarn remove actioncable`
+  * `yarn add @rails/actioncable`
+2. Replace any instance of `import Actioncable from "actioncable"` with `import { createConsumer } from "@rails/actioncable"`
```
jonathan-s, line 125: Tangentially related to this PR. Having our own `getConsumer` would be a good idea, so that it is agnostic. 

There is a proposal, but it's brewing I suppose :). Just a reminder. https://github.com/hopsoft/stimulus_reflex/pull/182

PR #210, opened 2020-05-23
Make element even more user friendly

hopsoft:
# Enhancement

## Description

Update `StimulusReflex::Element` to inherit from `OpenStruct` and expose attribtes via dot notation. Also, exposes `dataset` as an `OpenStruct` with the same behavior.

### Example

```html
<input id="name" type="text" value="hopsoft" 
  data-reflex="click->ExampleReflex#do_stuff" data-foo="bar">
```

```ruby
  def do_stuff
    element.id    # => name
    element[:id]  # => name
    element["id"] # => name

    element.value # => hopsoft

    element.dataset.foo    # => bar
    element.dataset[:foo]  # => bar
    element.dataset["foo"] # => bar
  end
end
```

## Why should this be added

Makes working with the element representation in the reflex even better.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/element.rb
+++ b/lib/stimulus_reflex/element.rb
@@ -1,15 +1,17 @@
 # frozen_string_literal: true
 
 class StimulusReflex::Element < OpenStruct
-  attr_reader :attributes, :dataset
+  attr_reader :attributes, :data_attributes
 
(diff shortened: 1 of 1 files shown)


jasoncharnes:
This is great! I was expressing the other day how great it would be to access `element.value`

Issue #186, opened 2020-05-05
Unnecessary body update after text_content


sevenmaxis:
I was following the tutorial [Twitter clone](https://www.youtube.com/watch?v=F5hA79vKE_E)

And I've noticed it updates the entire page after `text_content` operation
The first part of the log is where text_content happening and the second part is where the body update occurs.

```
StimulusReflex::Channel#receive({"target"=>"PostsReflex#like", "args"=>[], "url"=>"http://localhost:3000/posts", "attrs"=>{"class"=>"card-link", "data-reflex"=>"click->PostsReflex#like", "data-id"=>"25", "href"=>"#", "data-controller"=>"stimulus-reflex", "data-action"=>"click->stimulus-reflex#__perform", "checked"=>false, "selected"=>false, "tag_name"=>"A"}, "selectors"=>[], "permanent_attribute_name"=>"data-reflex-permanent", "reflexId"=>"4a031db8-a943-4e74-888a-518ffc4f9271"})
  Post Load (0.3ms)  SELECT "posts".* FROM "posts" WHERE "posts"."id" = ? LIMIT ?  [["id", 25], ["LIMIT", 1]]
  ↳ app/reflexes/posts_reflex.rb:7:in `like'
  Post Update All (1.9ms)  UPDATE "posts" SET "likes_counter" = COALESCE("likes_counter", 0) + ? WHERE "posts"."id" = ?  [["likes_counter", 1], ["id", 25]]
  ↳ app/reflexes/posts_reflex.rb:8:in `like'
[ActionCable] Broadcasting to timeline: {"cableReady"=>true, "operations"=>{"textContent"=>[{"selector"=>"post-25-likes", "text"=>8}]}}
Processing by PostsController#index as HTML
TimeLineChannel transmitting {"cableReady"=>true, "operations"=>{"textContent"=>[{"selector"=>"post-25-likes", "text"=>8}]}} (via streamed from timeline)
TimeLineChannel transmitting {"cableReady"=>true, "operations"=>{"textContent"=>[{"selector"=>"post-25-likes", "text"=>8}]}} (via streamed from timeline)
  Rendering posts/index.html.erb within layouts/application
  Rendered posts/_form.html.erb (Duration: 1.0ms | Allocations: 399)
  Post Load (0.2ms)  SELECT "posts".* FROM "posts" ORDER BY "posts"."created_at" DESC
  ↳ app/views/posts/index.html.erb:8
[...]

PR #211, opened 2020-05-23
Morph Modes: page, selector and nothing

leastbad:
## Description

This PR adds the concept of **Morph Modes** to StimulusReflex; *Morph Selector*, *Morph Nothing* and the current behavior, *Morph Page*.

Thanks to the thoughtful design of the existing library on both the client and server, it was possible to bolt on two entirely new modes of operation without any dramatic changes to existing functionality. Both benefit from the same logging, callbacks, events and promises as the classic refresh mode.

~~I suspect that a strict interpretation of SemVer will force a *minor* version bump.~~

**Selector Morphs** are similar to classic Page Morph operations, except that they do not call ActionDispatch - the controller is not run and the view template is not re-rendered. Instead, the developer can call the `morph` method one or more times to push an arbitrary string to an arbitrary DOM element. Selector morphs are syncronous - allowing you to morph a selector to display a spinner, call a slow API, and then remove the spinner... all in the same morph. Selector morphs also accept a hash syntax, allowing multiple morphs to be executed asyncronously in the same CableReady broadcast.

Yes, not hitting the routing engine OR ActionDispatch means that **you're going to be seeing a lot of 0-1ms updates**. Yes, this means that you can return a rendered partial collection, a ViewComponent or a string.

**Morph Nothing** is for calling Reflex actions that initiate long-running processes like ActiveJobs or brewing your morning coffee. They are the perfect place to make CableReady broadcasts that surgically update tiny bits of your DOM, like message notifications. When a Morph Nothing Reflex completes, no morphing occurs but all of the usual suspects (callbacks etc) fire.

I have added an optional 2nd parameter to `stimulate`. It has to be an object that contains at least one of the following keys: **attrs**, **reflexId**, **selectors**. If this options parameter is passed, it is `shift()-ed` out of the arguments splat.

Yes, you can now supply your own reflexId, and even override the attrs if you really know what you're doing. selectors is supposed to be an array of strings; if you pass `#poop` it will quietly coerce it into `['#poop']` for you. ~~In the next version of the library, `data-reflex-morph-target` will replace `data-reflex-root` (now **deprecated**).~~

~~Page reflexes that currently specify a data-reflex-root will get notified of both the switch to data-reflex-morph-target, and that morph_target will be ignored for Page reflexes starting with the next version.~~

Before I forget: I made the left/right arrows on @marcoroth's logging mechanism more defined. I'm getting old and my eyes were squinty trying to figure out whether I was looking at requests or responses.

Okay, demonstration time!

happening

```html
<div data-controller="evolve">
<div data-controller="evolve">
  <h4>Stimulus controller triggers</h4>
  <button type="button" data-action="evolve#page">Page</button>
  <button type="button" data-action="evolve#pageWithSelector">Page with Selector</button>
  <button type="button" data-action="evolve#selector">Selector</button>
  <button type="button" data-action="evolve#nothing">Nothing</button>

  <br><br>

  <h4>Declarative Reflex actions</h4>

  <button type="button" data-reflex="click->Evolve#page">Page</button>
  <button type="button" data-reflex="click->Evolve#selector">Selector</button>
  <button type="button" data-reflex="click->Evolve#nothing">Nothing</button>

  <br><br>

  <div id="poop">I'm sad <%= "and stinky" if @stimulus_reflex %></div>
  <div id="yuck">When will it end</div>

  <div id="pop"></div>
  <div id="mike"></div>

  <br><br>
  
  <div>A random number is: <%= rand(100) %></div>
</div>
```
```js
import consumer from '../channels/consumer'
import CableReady from 'cable_ready'

  connect () {
    StimulusReflex.register(this)
    consumer.subscriptions.create('EvolveChannel', {
      received (data) {
        if (data.cableReady) CableReady.perform(data.operations)
      }
    })
  }

  page () {
    // 2nd argument is an optional object containing *at least one of*: reflexId, selectors, attrs
    this.stimulate('Evolve#page', { reflexId: 'covid' })
  }

  pageWithSelector () {
    // accepts a string or an array of selectors, now
    this.stimulate('Evolve#page', { selectors: '#poop' })
  }

  selector () {
    this.stimulate('Evolve#selector')
  }

  nothing () {
    this.stimulate('Evolve#nothing').then(payload =>
      console.log('Works with promises!', payload.morphMode)
    )
  }

  afterPage () {
    console.log('Page rendered!')
  }

  afterSelector () {
    console.log('Selector rendered!')
  }

  afterNothing () {
    console.log('Nothing rendered!')
  }
}
```
```ruby
class EvolveReflex < StimulusReflex::Reflex
  include CableReady::Broadcaster
  def page
    puts "This is a classic SR full-page morph"
  end

  def selector
    puts "This morphs the target element"
    # ApplicationController.render(partial: "layouts/navbar/notification", locals: { notification: notification })
    # ApplicationController.render(ConfirmEmailComponent.new(user: current_user))
    morph "#poop", "I am full of it!" # default syntax is `morph "CSS selector", string` and it will be synchronous
    morph "#yuck" # string content defaults to an empty string; this will be a 2nd CableReady broadcast
    morph "#pop": "rocks", "#mike": "ike" # you can pass an implicit hash of multiple morphs that will be broadcast asynchronously
    # morph :nothing # this would raise an error because once you switch from page to selector or nothing, there's no going back
  end

  def nothing
    puts "This is where you initiate an ActiveJob or CableReady broadcast; it does not morph the client"
    morph :nothing
  end
end
```

Fixes #39

## Why should this be added

Many newcomers request the ability to update sections of the page and/or updating from specified URLs that might differ from the current page. While we understand *why* they are asking for these tools, the design of SR was intentional and it doesn't actually make sense to update the current page with the rendered content of a different page. There are simpler, clearer ways that won't confuse TF out of users. Morph Selector Reflexes are the perfect solution to this legitimate request.

Also, as the Ruby community lights up with the possibilities of CableReady, a Morph Nothing reflex is the rug that really ties the room together - it will make it easy for people to kick off long-running processes and call CableReady broadcasts. It is the perfect corollary to CableReady, in that it is a command sent from the client to the server.

## Checklist


marcoroth:
@leastbad Nice! Great work on this PR! 
I really love the idea of modes and can see them become very useful! (I can already see where I can improve stuff in my exisiting apps 😉)

The only concern I have is, that we "clutter" the framework by introducing a lot of new and sometimes specific data attributes (not particular to the features you just proposed, but in general). Is it just me to be overcautious?

julianrubisch:
I think @marcoroth is right that we have to watch very carefully how the API evolves. @leastbad‘s addition looks pretty straightforward in this respect.

hopsoft:
I have a set of integration tests stood up against the expo site but it's not complete and we need to make such testing more accessible to all contributors.

leastbad:
@marcoroth I agree with you 100% about being extremely careful about adding data attributes.

... says the guy intending to add mode - the first one since he added/championed permanent and root 🤠

leastbad:
Just throwing it on the table: there *is* a reasonable and possibly desirable alternative to introducing a new data attribute, and that is to revisit the syntax and conventions of `data-reflex`.

We borrow the Stimulus -> for refresh.

What if >> indicated a launch? What if ~> indicated an update?

While we’re brainstorming, @hopsoft is there a dimension close to ours where you’d consider making the string fragment Reflex optional, as it is completely redundant? All declared calls point to a reflex class, and all reflex classes end in Reflex. We could lop off six characters from every declaration.

marcoroth:
On my first read-through I didn't get that you put the `data-reflex-mode` attributes on the buttons and wondered how you decide which type of mode you are going to use. I first thought it's based on the reflex method names 🙃 

I just had a quick thought. I know, those are technically 3 new data-attributes but would eliminate the need for `data-reflex-mode`.

```html
<!-- data-reflex defaults to "refresh" -->
<button type="button" data-reflex="click->EvolveReflex#refresh">Refresh</button>
<button type="button" data-reflex-refresh="click->EvolveReflex#refresh">Refresh</button>
<button type="button" data-reflex-launch="click->EvolveReflex#launch">Launch</button>
<button type="button" data-reflex-update="click->EvolveReflex#update" data-reflex-root="#poop">Update</button>
```

This one has the disadvantage that you have 4 attributes which technically almost do the same thing. I don't know 🤷‍♂️ 

 I personally don't know which style I'd prefer, but I just wanted to share it with you anyway 😉

leastbad:
I also wanted to say, for posterity, that I’m not emotionally attached to refresh, update or launch as terms. I entertained many options, and these happened to float to the top.

It’s entirely likely that there’s a better word than launch, for example.

leastbad:
Sorry @marcoroth. We’re big kids here, you know I love you. Hard pass on introducing three new attributes.

seb1441:
Personally I think I prefer the `data-reflex-mode` attribute as it's more open if one day we wanna add more modes. And the behavior seems also more obvious than `~>`, `>>`.

marcoroth:
@leastbad I'm fine with this since I’m also not a big fan of introducing 3 new attributes (after I wrote that we should be careful about adding new ones 🙈)

I just wanted to share the idea anyway 😄

leastbad:
@seb1441 good points!

### Arguments for modifying data-reflex

- shorter syntax
- puts update and launch on equal footing with refresh (no extra markup required)
- no need to expand the schema

### Arguments for adding data-reflex-mode

- more easily allows future expansion
- syntax is more obvious
- work is already done

hopsoft:
I really like the ideas here and have been considering the public API for a while now. While I like how terse **modifying data-reflex** attribute values would be, it poses a higher risk of being confusing and prone to user error/misunderstanding (especially for developers new to the library). It also breaks with our convention of repurposing Stimulus semantics, so my vote is for **data-reflex-mode**.

Having said that, I'd like to propose a new name for **data-reflex-mode** because I think the current semantics on the PR are a bit too presumptive about the types of things people might use these features for. Here's my proposed _re_-naming scheme.

- `data-reflex-mode=refresh` to `data-reflex-render=page` **default** 
- `data-reflex-mode=update` to `data-reflex-render=partial`
- `data-reflex-mode=launch` to `data-reflex-render=none`

I realize that using `partial` as a value is making some assumptions; however, I think it's acceptable because (1) if feels very much like Rails (2) the meaning serves a dual purpose, so even if you render a `String` or a `ViewComponent`, it's still a partial page update.

I'd love to hear what others think.

julianrubisch:
Very thoughtful, 100% agree

marcoroth:
I like the idea of renaming it to `data-reflex-render`.  I just think it could be a little bit confusing to distinguish between `data-reflex-render=partial` and `data-reflex-render=page` with a `data-reflex-root`.

hopsoft:
@marcoroth Makes a great point and his concerns make me wonder if we should consider deprecating `data-reflex-root` as part of this PR. It would force a major version bump, but I think removing similar/competing features from the library is worth the effort and potential pain. At the very least we should consider deprecating **data-reflex-root** in favor of **data-reflex-render** and notify users with deprecation warnings in the logs.

Thoughts?

julianrubisch:
I‘d also opt for deprecating it for the moment

marcoroth:
@hopsoft In my opinion we shouldn‘t deprecate `data-reflex-root` because it would remove some of the awesome magic StimulusReflex provides. 

If I would need to provide the return value for every root in the Reflex it would be much more work then just letting StimulusReflex rerender the page through the controller and just let it morph the given DOM element(s).

marcoroth:
After I wrote that I realized that we could also name it the way **where**  the "response" is rendered. Either in the controller or in the reflex. 

This would lead to:
* `data-reflex-render="controller"` default
* `data-reflex-render="reflex"`
* `data-reflex-render="none"`

---

But apart from the naming, we would still need the `data-reflex-root` to define what part/element on the page should be replaced with the return value of the reflex method.

leastbad:
Of course I went to sleep at the exact moment everyone started saying interesting things...

Okay, thanks for thinking about this so much! I was ready the ideas and either in complete agreement or totally willing to go along with the landslide when I got to the part about potentially deprecating `data-reflex-root`. This bewilders me to the degree that I feel like I must be missing something.

If you get rid of `data-reflex-root` then how do you tell it which section to update with `data-reflex-render="partial"`? It's not hyperbolic for me to say that the reason I wrote this entire PR is *specifically* to enable people combining these two tools, as I did in the example on the PR. **render/partial + root is the money shot.** I'm very excited about none/launch, but it comes on the coat-tails of a vision for people using this targetted partial pattern *a lot*.

If we're worried about a confusing combination of names, I have two alternative suggestions. This first, respectfully, is that this concern is almost certainly overblown as we have documentation, examples and group of users who have clearly expressed a desire for this feature to exist... and they can be taught a new thing. The keywords aren't particularly confusing IMHO. The second suggestion is that we could introduce `data-reflex-target` with identical functionality to `data-reflex-root` and deprecate `data-reflex-root`, removing it in v4.

The only other idea I have at the moment is that in terms of keywords, I like "component" roughly as much as "partial" because it speaks to *what form the result will take*. I feel like partial is a means to an end, while many folks think of atomic chunks as components. This isn't a deal-breaker either way - I'd be perfectly happy with partial, too.

hopsoft:
I agree that removing the functionality is not what's desired; rather, more clarity in the API. I think what's not quite feeling right is supporting both:

1. `data-reflex-render=page` + `data-reflex-root` 
2. `data-reflex-render=partial` + `data-reflex-target` 
  I think **target** works better than **root** here semantically.

It feels like option 2 should be the preferred mechanic for partial page updates, so I'm questioning the value of keeping option 1 once this PR ships. Note that it's entirely possible that there are valid use cases, but I'd also prefer to minimize competing features in such a small library.

hopsoft:
I'll also go on record in defense of keeping **partial** for `data-reflex-render` as I think it fits better with Rails and still works semantically for pretty much any rendering preference. Be it components, partials, or something else entirely.

hopsoft:
To address @marcoroth's proposal of using **controller**, **reflex**, **none** instead of **page**, **partial**, **none**. I still prefer using the format of **what** vs **where** for `expando=value`.

leastbad:
Thank you for explaining this so clearly. I completely agree; the existance of partial+target makes page+root feel awkward and superfluous.

I feel good about have a page update be a page update. Simplifies everything. People can use `data-reflex-permanent` to mark things like Google Analytics which must be protected.

Question, is there still room for accepting a list of selectors, or should we go 1:1 for an update? It seems like if they want to do more than 1:1, they should start planning to call CableReady directly, no?

leastbad:
I'm going to update the PR with what I think is the latest decisions. Should I be changing data-reflex-root to data-reflex-target? How do we want to roll this out? v3?

hopsoft:
Let’s add the new target but keep root and emit deprecation warnings for now.

joshleblanc:
If you have a `FooReflex` with `data-reflex-render=partial`, is it expected that *every* time you call `FooReflex` you would also be using `data-reflex-render=partial`?

Is there a use case for having a FooReflex that can be called with all 3, doing different things each time?

It *seems* pretty specific. It *seems* like you're making a front end change to dictate backend behaviour. If that was the case, then wouldn't you want to define what you're rendering in the reflex itself?

```
def do_something
  render :page 
  # render :none
  # render :partial 
  
  ... 
end
```

You could actually still control it from the frontend by passing data attributes normally.

```
# FooComponent.html.erb
<button type="button" data-reflex="click->FooReflex#do_something" data-whatever="partial">Click me</button>

# FooReflex.rb

def do_something
  render element.dataset[:whatever]
  # ...
end
```

Just a thought

hopsoft:
Controlling what renders server side is interesting and feels very Railsy to me. I’d like to explore this idea a little more. i.e. implementing our own render method for Reflex to see how it feels.

leastbad:
The problem I see with this approach is that depending on the type of processing requested, `send`-ing the reflex action is just one step in a multi-step process. That process (and the steps in it) is already relevant before the code in the reflex action is running.

I suspect that serious users of the render :partial path will be calling it from Stimulus controllers.

However, yes - render :page is the default behavior and if you want to use a declarative reflex for updating just a segment of the page, you use a data-reflex-render. This is more explicit and no more characters than passing unstructed data to the server, and leaving it for every developer to come up with their own style.

I like how I did it. I did it that way for a reason, and I don't like it just because I did it.

ideasasylum:
> Controlling what renders server side is interesting and feels very Railsy to me. I’d like to explore this idea a little more. i.e. implementing our own render method for Reflex to see how it feels.

This was a more like the direction I thought things would go in. 

Last week I played around with how much of core SR would need to be modified to render a ViewComponent instead of using the original controller. [It was pretty neat](https://github.com/ideasasylum/stimulus_reflex/pull/1/files) and just required changing how the "page" was rendered and how the morphs were calculated (you wouldn't extract selections from the rendered component html, instead each selector would get the full html). I didn't need to change the javascript or data attributes at all.

But that got me thinking: this still requires modifying the SR core code each time someone wanted to render something in a different way. Ideally the reflex instance itself would be in charge of the doing the rendering and returning the morphs (selector => html maps). Then it would be simple to customise the rendering and keep the core of SR pretty clean.

I had the idea for a "renderer": a reflex would have/act as a renderer, the channel would execute the reflex as normal, then ask the reflex for the resulting morphs. The renderer would be responsible for rendering that html and calculating those selector => html mappings. By default, a reflex would have a ControllerRenderer which replicates the functionality currently in Channel but this could be overridden to change the rendering method.

Anyway, [I sketched out the idea](https://github.com/ideasasylum/stimulus_reflex/pull/2) to see how it felt. I used a module included in the Reflex but it could easily be a renderer instance.

@hopsoft is this along the lines you were thinking?

leastbad:
@ideasasylum this looks interesting, yet it’s not clear if this is being proposed as an addition to or an alternative to #211

I know that several folks are very excited about ViewComponent. My preferred order of operations is to get this PR merged before we start introducing custom renderers. The reason is simple: this PR, as written, is already a release unto itself. That’s both a lot of effort (documentation! Videos!) and a lot of new stuff for people to learn.

Most libraries have a small fraction of users that can arbitrarily map out an significant architecture change. We want SR to be a great liberator, so it’s incredibly important to ramp up the cognitive complexity in small increments.

I’m very concerned that the large percentage of Rails devs who would get cold sweats if you told them to “just implement a Renderer class” would be kind of freaked out by this approach. So far, we’ve done an extremely good job of keeping SR something you **use** instead of an API you program against.

Am I making any sense?

ideasasylum:
> this looks interesting, yet it’s not clear if this is being proposed as an addition to or an alternative to #211

I guess my ~proposal~ idea would be an alternative to this PR.

I'll admit I'm not sure what the strategic direction of SR is, or philosophically what the reflex's boundaries of responsibilities should be. I was operating within the constraints of not changing the javascript, for example, and letting the reflex control the rendering style. So it's likely I've misunderstood the direction or some other constraints.

> I’m very concerned that the large percentage of Rails devs who would get cold sweats if you told them to “just implement a Renderer class” would be kind of freaked out by this approach.

yeah, absolutely! I totally wouldn't do that either. Far too much overhead. 

In the approach I outlined, exactly nothing would change for existing users of full-page reflexes: No need to implement a renderer because the `ControllerRenderer` (naming tbd) would be the default and just extracts some of the existing functionality from `Channel`. If someone wanted a new renderer they could include a different renderer/subclass a different Reflex and no core SR code would need to be changed. The `ComponentRenderer` is one example of rendering a view component.

I see a few limitations with the mode approach outlined in this PR (but I may have misunderstood the code—please correct me!):

### The mode is specified in the view

The mode is specified in the view but this must be in sync with the way the reflex is coded.

For example, in partial mode you need to specify `data-reflex-render="partial"` in the view _and_ return the partial results from the reflex. Ideally the rendering mode/functionality would only need to be expressed in one place not in both the view and the reflex. I find it confusing to look at a reflex and not understand whether it's supposed to be used for page/partial rendering.

I'd prefer to see the rendering mode specified/implied in the reflex code—in the same place as the implementation.

### The return value from the reflex method is conditionally used

Depending on the rendering mode, the return value is either used or discarded.

In `page` and `none` modes the reflex return value is irrelevant. In partial mode, the reflex return value is assumed to be the content. This seems potentially confusing if you specify a partial rendering mode but the return value is not the content (plays back to my first point and my next…).

### The responsibilities of a reflex method changes based on the rendering mode

For a page render, the responsibility of the reflex method is simply to execute some business logic and express those result externally (i.e. updating a database record) or as instance variables. These instance can then shadow those normally set by a controller.

But for partial rendering, the reflex must perform the business logic _and_ then render the results to be returned.

I'd prefer if the reflex method implementation remained purely about the business logic and the rendering encapsulated elsewhere.

### Lost the ability to use multiple selectors

I'm not sure why it was necessary to remove this? I'll admit I slightly prefer the target-morph naming over root-selector naming but probably not enough to deprecate it. I might be missing a bigger strategy here though.

### Not entirely sure we need `none`?

`none` mode seems like is should just be a case of returning no morphs rather than a mode specified in the view

### SR core now has more logic

With modes, SR core now has more logic for handling the various alternative rendering. 

The core responsibilities of SR (server side) seems to currently be: executing a reflex (business logic), generating some output (currently re-running the controller action), sending the morphs back (selector-html mappings). I'm proposing that we delegate the output generation to a renderer class that the reflex can change/influence but doesn't need to. 

I can elaborate further on my approach, and make a full PR if it sounds promising, but this comment is already too long for a sunny Sunday… 🌞

leastbad:
Thanks for your detailed explanation, @ideasasylum! I am actually trying to explain the strategic direction, which is informed both by our vision for what's possible as well as a few hundred conversations with folks using the library over the past year - it's not coming out of nowhere.

> I'll admit I'm not sure what the strategic direction of SR is, or philosophically what the reflex's boundaries of responsibilities should be. I was operating within the constraints of not changing the javascript, for example, and letting the reflex control the rendering style. So it's likely I've misunderstood the direction or some other constraints.

... you just have to ask! Are you participating in the ongoing community and conversation on Discord? We have very involved and transparent discussions about how best to proceed and your voice is certainly welcome. For example, not changing the JS isn't really a constraint, so long as we're not making breaking API changes. This library is under vibrant development and the JS is fair game for people's better wild ideas.

One of the most powerful aspects of SR is the quality of the architectural design. It means that one of the chief benefits we enjoy as library developers is that it's relatively easy to bolt on or radically change the way certain things work. The places where the conceptual atoms fit together are quick logical and it makes it easy to experiment with concepts like modes or making the whole thing work with message_bus. It's really exciting to see alternative approaches emerge. There's no need to "correct" you and you're not in trouble for having ideas.

To address some of your points and questions:

- The mode is specified in the view: yes, and as a result it allows us to take the request through a different code pathway designed to address one of the three usage patterns we've identified; depending on how the reflex will be processed, we like reserving the opportunity to make decisions about this pathway before the reflex action method itself is executed

- The return value is conditionally used: yes, and it's one of my favourite aspects of this PR - I cannot wait to use it :grin: ... it's important to note that we take documentation very seriously and like everything else, we intend to show people exactly how this is to be used. I personally love that it expands the creative possibilities of how people can assemble the HTML that is returned

- The responsibilities of a reflex method change based on the mode: yes, and this is directly informed by the way people use the library today (ActionDispatch only) and the flexibility people regularly request wrt surgical updates; again, nobody wants to send these soldiers into the jungle without a map, so the documentation and video support is vital... but a huge part of the excitement for this PR is that we're finally giving devs a mechanism to do exactly what they want to do, and in a way that completely bypasses ActionPack, ActionDispatch and Nokogiri - so in theory, allowing for roundtrips that you just don't see on a typical Rails app. So exciting!

- Multiple selectors: we're still debating this internally, but even @marcoroth - a big multi-selector enthusiast - seems to be coming around to the notion that if they hadn't been there when he started working with the library, he would not likely have proposed them :shipit: ... that all said, it's possible that multi-selector full page reflexes could still survive... again, I'd recommend catchup up on Discord here

- Do we need none? I cannot wait to be able to use done. One of the most important things that we're going to be focusing on in our library evangelism over the coming months is the emergence of CableReady as being a power tool that can be used to solve entire classes of UI/UX concerns that are not well-covered by SR; Nate's Twitter clone video started to wake people up to this a bit, because the big secret of that video is that it's really a CableReady video that uses SR to initiate some processes. Anyhow, I'm rambling but the SR "none" -> ActiveRecord callback -> ActiveJob -> CableReady pattern is enormously potent

In conclusion, there is method to our madness. You'd be hard-pressed to find a group of maintainers that are more passionate about making sure this library succeeds, because we literally see it as the great hope for liberating the folks who are currently forced to use React in an overly complex JS ecosystem. We believe that we all deserve better, and this PR reflects us listening carefully to what people keep trying to do.

leastbad:
I wish I'd seen #223 before committing, but to me this is a pretty good example of the type of problem that can easily emerge with multiple selectors. It is fundamentally brittle, and unfortunately a little too easy to confuse Nokogiri, morphdom or both IMHO.

It's also really tricky to debug. 😿

RolandStuder:
> The awkwardly-named None reflexes

How about naming them `empty reflexes` instead of `none reflexes`

leastbad:
I woke up today realizing that @marcoroth is right - I no longer thing that this PR should disable the multi-selector functionality built into data-reflex-root in the current master.

I had good intentions but it's simply true that it does offer some utility and since I usually rally against purity tests, I can't make a strong argument for purging `data-reflex-root="[backwards],[forwards]"` when we really can just keep it.

leastbad:
As this PR has now had a month to marinate, I just wanted to regroup and focus our thoughts and energies on how best to move forward.

First, I just want to say that it's really flattering and exciting how positively this was received, considering that it's a major set of changes and I'm still fighting significant imposter syndrome when it comes to working at the library level. Overall, the feedback from people who want to be able to easily trigger events on the server and/or have more granular control over what gets delivered where has been extremely positive. With the release of broadcast_to in CableReady and the tempest-in-a-teapot that has been HEY and its Turbolinks-frame mechanism, it's clear that we're moving towards some really exciting outcomes. I'm also very excited about the progress @joshleblanc has made on his [ViewComponentReflex](https://github.com/joshleblanc/view_component_reflex) project, which just keeps getting cooler.

There's been some healthy back-and-forth on naming, deprecations and how to trigger these new modes of operation. I think we all understand that this is shaping up to be a milestone and frankly, the decisions we make around this PR are incredibly important - we will take as much time as we need to get this right as we only get one shot at it. I'm more than happy to keep doing passes on the functionality, so long as it's not back-and-forth. :wink:

@joshleblanc @hopsoft and @ideasasylum discussed potentially moving where the mode switch happens from the client to the server. The idea for implementing a SR render method was floated. Jamie suggested that it would be better to mark a reflex as being for a purpose on the server than directing a behaviour from the client. Having thought about it for a few weeks, I've changed my position and now support setting a flag in the reflex method to change the mode from `:page` to `:partial` or `:none`. I don't see a real advantage to specifying this on the client at all, anymore, because whatever happens is going to be implementation specific, right? You're going to write very different reflex action code depending on whether you want to refresh the page, tweak some elements on the page or just trigger some server processes.

My proposal is that we create a `render` method (**I'm not attached to `render` - it could be `update_mode`**) that you call in your reflex action method, defaulting to `:page` as we already do now:
```ruby
def do_the_job
  render :action_only
  LaunchJob.perform_later
end
```

I am proposing `:action_only` instead of `:none`. After a month of hand-wringing, I hate `:none` so much. Grammatically, it is awkward and confusing. I support the idea of what over where, so :action_only seems like a really good answer to the question.

Devil's advocate: maybe we support **both** client and server mode selection? Can anyone think of a case where client-side mode selection is important or enabling? My current preference is to move to server-side only, but I'm eager to be reminded of use cases.

I would also like to suggest that we at least consider `:root` over `:partial`. Because what are we updating? An element, not a partial. Plus, we're already using `data-reflex-root`.

Even though I am proposing we drop the `data-reflex-render` attribute in favor of setting the mode in the reflex action, I would advocate for keeping the optional options parameter that I defined in this PR so that we can support options in the future. This adds a few lines to the library but affords us a huge amount of room to expand in the future. Frankly, I like my implementation and think we'll need it sooner than later.

Now, I still really like (love?) returning the string that the :partial/:root mode will render from the reflex action, but hey, whatever is best for the project... The alternative would be allowing `render` to accept an optional second parameter that expects a string:
```ruby
def display_notification
  # targets whatever selector has been specified in data-reflex-root
  render :root, ApplicationController.render(partial: "layout/notification")
  # render :partial, ApplicationController.render(partial: "layout/notification")
end
```

I begrudgingly acknowledge that this syntax doesn't offend me, and it would vindicate @ideasasylum's critique that it's awkward to have one mode care about what's returned when the others don't. What would everyone prefer?

I would propose that we simply continue to use the `data-reflex-root` attribute to pass a selector for the :partial/:root mode, although we'd have to restrict this to one selector and make it clear that we accept the first valid selector and discard the rest.

One thing I've been thinking about is that technically, developers can already override the default @selectors value of ["body"] if they want to in their reflex action. What if we formalized this with a method available in the reflex action called `set_root` or `set_selectors` which would accept a string or an array? If it was a string, we would split it into an array as we do in the client. What I like about this proposal is that it seems like it would make the `data-reflex-root` syntax optional, allowing for shorter HTML markup and continuing the trend of moving decisions to the server.
```ruby
def display_notification
  set_root "#notification"
  render :root, ApplicationController.render(partial: "layout/notification")
  # or as one method with three params
  render :root, ApplicationController.render(partial: "layout/notification"), "#notification"
end
```

The last interesting proposal that has been made is that the :partial/:root update could potentially support calling `render` multiple times. This is actually pretty interesting. It could look like this:
```ruby
def update_multiple_things
  render :root, ApplicationController.render(partial: "layout/notification"), "#notification"
  render :root, ApplicationController.render(partial: "layout/count"), "#count"
  # or how about...
  render partial: {
    "#notification": ApplicationController.render(partial: "layout/notification"),
    "#count": ApplicationController.render(partial: "layout/count")
  }
end
```

The key distinction between a `:page` and a `:partial/:root` remains that a `:page` calls `ActionDispatch` and runs the controller action for the current page. `:partial/:root` does not. All three modes trigger the usual callbacks, promises and events.

Finally(!), as I said in the last comment, I no longer think that we should deprecate `data-reflex-root` and its current functionality in the context of a `:page` reflex. It serves a purpose and people are using it.

RolandStuder:
I love the multiple renders idea, something very useful when you add an item to a list. You want to append it to the list, but also render the new form again, or update a nav to make the right nav-item active and change the detail view.

I prefer `render :partial` over `render :root`, because usually you try to only have on root, so partial feels more semantically correct for what we are enabling with it.

mtomov:
Hi, thanks for the writeup!

After gathering a bit of experience with the library, I feel more confident to say that the design of the reflexes should break apart from the traditional request -> response thinking, if for nothing else, but because websocket messages do not have native support to indicate that one message is a response to another [source](https://blogs.windows.com/windowsdeveloper/2016/03/14/when-to-use-a-http-call-instead-of-a-websocket-or-http-2-0/)! 

In that case, the reflex method is actually the message coming in, and from there we can support to send 0, 1 or N messages. Those messages should be rather flexible, and CableReady already provides most of the flexibility that we're after:

#### The current behaviour:
Invoke the controller action responsible for the same url that the request has come in from, then take the html of the response and morph it at the frontend side with the body tag.

#### Future behaviours
- all related to using CableReady to broadcast any number of operations either together in a single message or in multiple messages -> so the user will get a UX change asap. 


*Examples of behaviours I'd like to trigger:*

1. produce a piece of html, say with ViewComponent, send it, and choose what operation to perform with it on the frontend - options being `morph`, `innerHTML` and friends, or call a frontend controller action to perform custom behaviour - think a `popup` to show up, or a screen out animation to play with an existing html element before playing a screen-in animation of the html coming in.

2.  Be able to send json to a frontend controller - example is I want to send both an html piece and a `url` string, and have a controller action on the frontend both insert the html in the page, and change the URL

3. Not send a message back .. like when just triggering a background job

4. The current `ActionDispatch` behaviour, but to be able to say which action exactly to call .. using a settable `url` is one option, but maybe more directly is another.

I actually think that covers it. To give examples on the implementation itself:

```ruby
def update_multiple_things
   # obviously not sure how that works, but that's the current behaviour of page updating. 
   # Can certainly be made shorter, but the long form should be supported for 
   # better understanding of what's happening "behind the scenes"
  CableReady.morph('body', ActionDispatch.call_controller(BlogPosts, :index))

  # then we broadcast just that piece
  CableReady.broadcast! 

  # now queue 3 operations together
  CableReady.morph('#notification-bar', ApplicationController.render(partial: "layout/notification"))
  CableReady.inner_html( '#topbar-counter', ApplicationController.render(partial: "layout/count"), 

  # haven't used View Components yet, not sure how to call them, 
  # but the idea is to produce an html or a json piece.
  CableReady.call_frontend_controller_action( 'PopupController#handleServerMessage', PopupComponent.render())

  # send all 3 operations together
  CableReady.broadcast! 
end
```

```js
// popup_controller.js

  // message coming back
  handleServerMessage(jsonPiece) {
    // Do something with the json, such as animate one element,
    // morph another, etc..
    CableReady.DOMOperations.innerHtml({
      element: document.querySelector()
      jsonPiece.part1,
    })

    MyAnimation.run(jsonPiece.part2)

    history.pushState({}, jsonPiece.title, jsonPiece.url)
  }

  // message forward
  doMultipleThings(event) {
    this.stimulate('MyReflex#do_multiple_things')
  }
}

```

Hopefully that takes the conversation forward. Thank you!

leastbad:
Hey @mtomov! Thanks for the really interesting notes.

For what it's worth, I'm **primarily** looking for feedback on the very-not-rhetorical questions in my last comment. I want to stress that this PR is significantly past the green field brainstorming phase; looking to firm up go-forward opinions on naming, function arguments and potential deprecations.

... but I want to address your thoughts here, too. 😁 

1. This isn't likely at this time, because the whole point of :action_only Reflexes is that you can do anything you can imagine to your page, including calling any of the CableReady methods. We want to keep the library fairly A->B->C with a clean default code path that morphs your page. Easy to learn, and frankly what most folks want.
2. This is a really clever idea, but far out of scope for StimulusReflex. Instead, you can use an :action_only Reflex to call dispatch_event, pack your event full of markup and URL strings, and then pick up that event in your Stimulus controller.
3. This is the :action_only mode Reflex in a nutshell!
4. With the introduction of the :partial Reflex mode, you can send arbitrary HTML to arbitrary DOM nodes. We will *not* be introducing arbitrary URL rendering to the default :page mode Reflex.

Based on your wishlist, it sounds like you're going to love these new modes once you've had a chance to work with what we've come up with. This design came from talking to literally hundreds of people with unique problems and charting a straight line through the feedback.

Given all that, I'd love to hear your thoughts on all of the questions in my last comment!

joshleblanc:
I'll throw my 2 cents in for the api

Since we're using `morphdom` for dom reconciliation, I think `morph` makes a good method name. 

Then for the 3 modes, we could use 

```
morph :nothing # don't do any morphing

morph :selector, selectors, html # morph the specified selector with the specific html 
                  
morph :page # Morphs the body with the controller response as we do now
```

I feel this naming stragety lends itself to easy reading.

I should note I think it makes sense that a single morph call also broadcasts immediately, rather than after the reflex is done.

julianrubisch:
I think @joshleblanc has nailed the method name, `morph` feels very natural and elegant.

I'd also like to +1 his suggestion of calling it `:selector(s)` vs `:partial` or `:root`. 

FWIW, I'm also very fond of this syntax:

```rb
  morph selectors: {
    "#notification": ApplicationController.render(partial: "layout/notification"),
    "#count": ApplicationController.render(partial: "layout/count")
  }
```

I don't think that methods with a varying number of arguments does any cognitive harm, because they're all keyword args and we're used to that anyway. The only thing that comes to mind is that the implementation might be tricky, as to what to prefer in case you'd mistakenly write `morph :page, :action_only`, or what have you...

davidalejandroaguilar:
I'm not fond of `morph` because:

* It is tied to an implementation detail. What if we decide to use another library other than `morphdom` in the future?
* This doesn't feel "Railsy" or following "The Rails Way". One of the things I liked when first learning about this library was that it adhered to the Rails philosophy. We even have that on the README: *The goal is to help small teams do big things with **familiar** tools.* `render` feels more *familiar*.

julianrubisch:
True, FWIW I don’t connect morph to morphdom, it’s just more accurately describing what is being done.

leastbad:
Render could be so familiar that it's overloaded, unfortunately.

Here's the thing: if we found that there was another library that changed the active DOM to reflect the new state of the application, we'd still be morphing the page. **The rendering has already happened.** It's not just an implementation detail - it's the best description of what this stack does to deliver updates.

Anyhow, I've asked Nate to render a final verdict on this. I don't actually have a strong opinion, believe it or not - I want to convert open questions to a specification so that I can refactor this PR and we can start planning how we'll promote it. I want this in master yesterday, because this is functionality that I want to use.

julianrubisch:
That reminds me, I wanted to bring this up earlier, but how do we version this? I suspect this is going to be a breaking change?

davidalejandroaguilar:
I have 2 ideas based on my last comment on making things more *familiar* and agreeing on @leastbad 's [last comment](https://github.com/hopsoft/stimulus_reflex/pull/211#issuecomment-647063536) proposals.

1. We allow only _one_ `render` call:
  - `render` without arguments: triggers a page update.
  - `render root:`: triggers a partial update on the `data-reflex-root` element.
  - Not calling `render`: does not trigger any kind of DOM update.

Further `render` calls are not allowed, just like controllers.

Examples:

Page render:

```rb
class NotificationReflex < ApplicationReflex
  def create
    notification = Notification.create!

    render
  end
end
```

Partial render:

```rb
class NotificationReflex < ApplicationReflex
  def create
    notification = Notification.create!

    render root: {
      '#notification': ApplicationController.render_to_string(notification),
      '#count': Notification.count
    }
  end
end
```

```rb
class NotificationReflex < ApplicationReflex
  def create
    notification = Notification.create!

    set_root '#notification'
    # Can now only provide a single string?
    render root: ApplicationController.render_to_string(notification)
  end
end
```

No render:

```rb
class NotificationReflex < ApplicationReflex
  def create
    Notification.create!
  end
end
```

Granted, this requires an explicit `render` call for a page render. If that's not desirable, then next idea:

2. We use `render :nothing` to not render anything:
  - Not calling `render`: triggers a page update.
  - `render root:`: triggers a partial update on the `data-reflex-root` element.
  - `render :nothing`: does not trigger any kind of DOM update.

`render :nothing` is what [Rails had](https://github.com/rails/rails/blob/1099329be081297c16b02141ec13ca6a05ff037c/actionpack/lib/action_controller/metal/rendering.rb#L101) before switching to `head`.

Examples:

Page render:

```rb
class NotificationReflex < ApplicationReflex
  def create
    Notification.create!
  end
end
```

Partial render:

```rb
class NotificationReflex < ApplicationReflex
  def create
    notification = Notification.create!

    render root: {
      '#notification': ApplicationController.render_to_string(notification),
      '#count': Notification.count
    } 
  end
end
```

```rb
class NotificationReflex < ApplicationReflex
  def create
    notification = Notification.create!

    set_root '#notification'
    # Can now only provide a single string?
    render root: ApplicationController.render_to_string(notification) 
  end
end
```

No render:

```rb
class NotificationReflex < ApplicationReflex
  def create
    Notification.create!

    render :nothing
  end
end
```

---

I just read the last comment on the name `render` may be overloaded. I agree that the rendering has already happened, but we're "rendering" additional content afterwards. I think this is more Railsy and beginner-friendly and less technical.

However, I'm also not attached to it and ultimately the decision should be based on what's makes the most sense and works best for developers.

leastbad:
@davidalejandroaguilar I really appreciate you advocating for your vision. I want you to know that I read it and considered it.

I also need to warn you that this version of things is not likely to be what we go forward with. It really comes down to a straw poll and right now there's a lot of momentum away from what you're proposing.

On the plus side, what we're excited about is more explicit and more flexible. Consider that by preferring morph, we're also stepping away from the possible confusion around having more than one render.

RolandStuder:
I actually agree that both `render` and `morph` are not very semantically correct.

We are not actually saying what should be rendered, but we are controlling the flow as in:

* `perform :action` -> go to the controller action and send results over the wire.
* `render :whatever` -> send this bit that I pass to you over the wire.
* `do :nothing`-> Do not do anything else, but what I specifically told you to do.

I like @davidalejandroaguilar proposal, but would slightly change it, don't use render, instead use something like wire or broadcast

````
wire :controller_action #default
wire :nothing #or nil
wire SomeViewCompontent.to_string
````
Allowing multiple wires, unless you broadcast `:controller_Action`

But you could also use `wire` or broadcast if you to use to configure what to use in the reflex with `broadcast :controller_action`, `broadcast :partial`, `broadcast :nothing`

I think broadcast or wire work better, as @davidalejandroaguilar actually display the problem with using the word render, to decide what we broadcast for example: 
`render root: ApplicationController.render_to_string(notification)`

We are saying here, semantically to render what we render!? We can render whatever we want in the Reflex, the decision is, what we send over the wire, what we broadcast.

Also as you see, I prefer something more explicit than page, by using `:controller_action`I think the verbosity makes it clearer, and I don't mind the verbosity as we will rarely have to tell a reflex explicitly to use the default.

**Comment by leastbad** [comment]
The strike against broadcast is that it's the primary method in CableReady. I'd prefer not to introduce that confusion.

I don't love wire... but that's just a poll vote. You haven't convinced me to not use morph. I find the "morph #this_id into this_string" structure of what's proposed highly grokkable and English-like. "wire #this with string" doesn't really parse as something a human would say in this context.

As previously discussed, there's *almost* no reason to call `morph :page` (or whatever we call it) as it is the default behaviour. The only reason it makes sense to expose it is in the case of branching logic determined by an external side effect like an API call eg. render the page or just an error popup. So again, Nate will decide whether it is :page, :controller_action or something else.

At least we all seem to be coalescing around :nothing instead of :none, which was giving me seizures.

Note: at this point I do not intend to pull data-reflex-root values into what I'm now calling :selector reflexes; they will be used for :page reflexes only. Current intent is for :selector reflexes to provide their own selector. Otherwise, there's no obvious way to map the selector that arrives to one of potentially several distinct morphs.

**Comment by hopsoft** [comment]
Love this PR and conversation. I think letting it simmer for a month has proven incredibly helpful for the thinking and decision making. 

### Key decisions that I very much agree with.

1. Moving the rendering concern to the server
1. Moving DOM selectors to the server
1. Keeping `data-reflex-root` behavior _as is_
1. Keeping default behavior to re-render the page _as is_

### API

I think that repurposing `render` semantics is problematic. i.e. different method signature, ok to invoke multiple times, etc... I'm liking `morph` as the method name and agree that it's descriptive and not really tied to morphdom. I'm not yet sold on the `Hash` argument given that we can call `morph` multiple times. We can always add it later if people really want it.

#### Usage

Here's how I see the API usage shaping up.

```ruby
  def page_morph
    # page will automatically re-render, this is the default
  end

  def targeted_morph
    morph "#selector", ApplicationController.render(partial: "/path/to/partial")
  end 

  def multiple_targeted_morphs
    morph "#selector", ApplicationController.render(partial: "/path/to/partial")
    morph "#selector", ApplicationController.render(partial: "/path/to/partial")
    morph "#selector", ApplicationController.render(partial: "/path/to/partial")
  end

  def no_morph
    morph :nothing
  end
end
```

hopsoft:
Also, this PR resolves #186

julianrubisch:
Once this is merged, I could pick up work on https://github.com/hopsoft/stimulus_reflex/pull/183 again, if it is still desired. 

Also, let me restate my question about semantic versioning. @hopsoft‘s API 👆🏻 has the benefit that it would not break existing implementations, IIUC?

hopsoft:
@julianrubisch I'd love to sync and collaborate on your ideas around #183 once things settle down a bit for me.

jonathan-s:
Also happy to sync on the equivalent javascript adapter once things have
settled down a bit :)

leastbad:
I love where this landed; no need to specify a morph type unless it’s :nothing.

I’m going to get to work on making this exist. Once it’s done, let’s each of us try to kick the tires really hard for a week. I’m going to try and script a video to go out for the release.

lib/stimulus_reflex/channel.rb
```diff
@@ -41,7 +42,14 @@ def receive(data)
       broadcast_message subject: "halted", data: data
     else
       begin
-        render_page_and_broadcast_morph reflex, selectors, data
+        case reflex.morph_mode
```
julianrubisch, line 45: Okay I‘m being a know it all here, but I really have a strong feeling „morph_mode“ should be its own class, or rather superclass. That way you could resort to polymorphism here (and possibly at other locations too) and do „tell don‘t ask“

leastbad: Respect. So, here's what's happening: I'm merging from master, then pushing, then updating the readme, then going to sleep. I'd be thrilled to wake up to a PR.

Truth is, that sounds great but I have no idea how to do what you're suggesting. If you want it, I'd love to learn.

lib/stimulus_reflex/broadcaster.rb
```diff
@@ -0,0 +1,57 @@
+module StimulusReflex
+  module Broadcaster
+    def self.included klass
```
hopsoft, line 3: I prefer parens on method definition arguments as a general rule. 

Also, I'd prefer to make this a bit more idiomatic Rails-like code by including `ActiveSupport::Concern` and then using `module ClassMethods` to define the class methods.

lib/stimulus_reflex/morph_mode/selector_morph_mode.rb
```diff
@@ -0,0 +1,15 @@
+module StimulusReflex
+  class SelectorMorphMode < MorphMode
+    def broadcast(reflex, selectors, data)
+      broadcast_message subject: "selector", data: data
```
hopsoft, line 4: Shouldn't this be a call to `broadcast_morphs`?

julianrubisch: This was literally copy pasted. Maybe @leastbad has some insights?

lib/stimulus_reflex/broadcaster.rb
leastbad: Fortunately or unfortunately, I am 100% reliant on @julianrubisch for this ask. The words are familiar but I don’t know how to do that.

lib/stimulus_reflex/morph_mode/selector_morph_mode.rb
leastbad: I do! 🤠

The actual changes are broadcast from reflex.rb; this triggers the callbacks.

Selector Morphs will broadcast N+1 times.

lib/stimulus_reflex/broadcaster.rb
julianrubisch: Could you write that out as a suggestion? I’m not completely sure I understand, either

leastbad:
Starts with a bang, ends with a whimper. 💯

leastbad:
Thank you to @hopsoft @julianrubisch @marcoroth @seb1441 @joshleblanc @jonathan-s @ideasasylum @RolandStuder @mtomov and @davidalejandroaguilar for your opinions, reactions, advice and patience. Go team!

PR #212, opened 2020-05-24
reflex_name restriction loosening

leastbad:
## Type of PR (feature, enhancement, bug fix, etc.)

# Enhancement

## Description

This PR would make the string fragment "Reflex" used in the `data-reflex` attributes optional, allowing for a shorter syntax.

It also allows for the use of Reflex classes that do not end in Reflex, while not changing the existing default behavior.

### Examples

This is the current behavior:

```html
<button data-reflex="click->PowerReflex#lift">Lift</button>
```
This calls PowerReflex#lift.

```html
<button data-reflex="click->Power#lift">Lift</button>
```
This first checks to see if there is a Power class, and if there is, that it implements `StimulusReflex::Reflex`. If everything looks good, Power is your Reflex class.

If there is no Power class, or Power does not inherit from `StimulusReflex::Reflex`, it falls back to PowerReflex as your class. This is transparent to the client.

## Why should this be added

First, I should say that I'm not emotionally attached to this outcome, and I'm not in love with my implementation - although I'm glad it's 100% server-side. Unfortunately, Ruby's `defined?` method handles `User` and `"User".constantize` differently:

```ruby
x = User
defined? User # true
defined? "User".constantize # "method"
defined? x # "local-variable"
```
I'm sure there are more experienced Rubyists who know how to make this better, but I didn't want to require therapy to get this done, so I went with an admittedly hacky begin..rescue approach, instead.

Anyhow, here are the FOR arguments:

- shorter syntax
- not restricted to classes ending in "Reflex"
- backwards compatible

## Heads up

I found a bug: line 31 of `channel.rb` (on current master branch; it's 32 in this PR) raises an ArgumentError if the reflex_class does not inherit from `StimulusReflex::Reflex`. However, line 35/36 - the first line of the error rescue, calls `reflex.rescue_with_handler()`. Unfortunately, if reflex_class is *not* a Reflex, it will be **nil**, causing this rescue mission to fail.

In other words, on today's master branch, if we have a power_reflex.rb that forgets to inherit, it will throw this ugly error instead of a nice friendly error.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -54,7 +54,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
   private
 
   def compute_reflex_name(reflex_name)
-    return "#{reflex_name}Reflex" unless is_reflex?(reflex_name.constantize)
+    raise unless is_reflex?(reflex_name.constantize)
     reflex_name
   rescue
     "#{reflex_name}Reflex"
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -57,7 +57,7 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
     begin
       return "#{reflex_name}Reflex" unless is_reflex?(reflex_name.constantize)
       reflex_name
-    rescue => e
+    rescue
       "#{reflex_name}Reflex"
     end
   end
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -54,12 +54,10 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
   private
 
   def compute_reflex_name(reflex_name)
-    begin
-      return "#{reflex_name}Reflex" unless is_reflex?(reflex_name.constantize)
-      reflex_name
-    rescue
-      "#{reflex_name}Reflex"
-    end
+    return "#{reflex_name}Reflex" unless is_reflex?(reflex_name.constantize)
+    reflex_name
+  rescue
+    "#{reflex_name}Reflex"
   end
 
   def is_reflex?(reflex_class)
--- a/lib/stimulus_reflex/channel.rb
+++ b/lib/stimulus_reflex/channel.rb
@@ -55,8 +55,8 @@ class StimulusReflex::Channel < ActionCable::Channel::Base
 
   def compute_reflex_name(reflex_name)
     begin
-      is_reflex?(reflex_name.constantize)
-      return reflex_name
+      return "#{reflex_name}Reflex" unless is_reflex?(reflex_name.constantize)
+      reflex_name
     rescue => e
       "#{reflex_name}Reflex"
     end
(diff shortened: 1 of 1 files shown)


lib/stimulus_reflex/channel.rb
```diff
@@ -52,6 +53,13 @@ def receive(data)
 
   private
 
+  def compute_reflex_name(reflex_name)
+    raise unless is_reflex?(reflex_name.constantize)
+    reflex_name
+  rescue
+    "#{reflex_name}Reflex"
```
foca, line 60: You can use `const_defined?` for this:

```suggestion
    Object.const_defined?(reflex_name.to_s) ? reflex_name : "#{reflex_name}Reflex"
```

Needs to be scoped to `Object` to ensure the lookup path is what you'd expect. And _needs_ to be a String, hence the `to_s` (technically, anything that responds `to_str`.)

lib/stimulus_reflex/channel.rb
```diff
@@ -22,6 +22,7 @@ def receive(data)
     target = data["target"].to_s
     reflex_name, method_name = target.split("#")
     reflex_name = reflex_name.classify
+    reflex_name = reflex_name.end_with?("Reflex") ? reflex_name : compute_reflex_name(reflex_name)
```
hopsoft, line 25: What about something like this?

```ruby
reflex_name = reflex_name.end_with?("Reflex") ? reflex_name : "#{reflex_name}Reflex"

# ...

reflex_class = reflex_name.constantize.tap { |klass| raise unless is_reflex?(klass) }
```

lib/stimulus_reflex/channel.rb
leastbad: There needs to be a #clearlydoesntknowtherightfunction tag. You are very kind. Obviously you are right!

lib/stimulus_reflex/channel.rb
leastbad: This would work, but it has less functionality than what I proposed.

My version lets you define a class with any name that inherits from `StimulusReflex::Reflex`

So you could have a totally viable `Balloon << StimulusReflex::Reflex`

hopsoft: Not sure I like that flexibility. Imagine a Rails controller not suffixed with “Controller”. I’d rather enforce convention on this.

leastbad: While I rarely find myself accessing controller consts directly, I agree with your point overall and then some. If the tables were turned, I probably wouldn't accept this logic. I do like putting sketches on the table in case something sticks. :)

leastbad:
Note that I'm pretty sure that this still explodes the rescue attempt if reflex_class is somehow not a valid SR:R, but perhaps this is as fine as it's always been. @hopsoft

lib/stimulus_reflex/channel.rb
leastbad: @foca we went in a different, simpler direction with this but I really appreciate learning about const_defined? !

foca: The resulting code is much simpler overall! I digs.

PR #213, opened 2020-05-25
Proposal: data-reflex-debounce

marcoroth:

Feature

## Description

First of all, I know it's yet another data-attribute 😉  ...and there are also ways to do this in JavaScript/Stimulus with something like the `debounce` function from `lodash-es`.

But nevertheless, it might be interesting to support something like this out of the box in StimulusReflex itself. 

But just to be honest: I'm not emotionally attached to this feature at all

## Why should this be added

This feature would make it easier to integrate a `debounce` into the StimulusReflex workflow and would allow to setup a debounce without the need of a Stimulus controller. 

### Example

This example would setup a debounce for the `<input>` element and would only fire after the given 500ms.

```html
<input data-reflex="keyup->ExampleReflex#search" data-reflex-debounce="500" />
```

I'm curious what you think about supporting something like this out of the box or if we should stick to our beloved Stimulus controllers ☺️ 

## Checklist

### DIFF
--- a/javascript/schema.js
+++ b/javascript/schema.js
@@ -1,5 +1,6 @@
 export const defaultSchema = {
   reflexAttribute: 'data-reflex',
   reflexPermanentAttribute: 'data-reflex-permanent',
-  reflexRootAttribute: 'data-reflex-root'
+  reflexRootAttribute: 'data-reflex-root',
+  reflexDebounceAttribute: 'data-reflex-debounce'
 }
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -26,6 +26,10 @@ let actionCableConsumer
 //
 const promises = {}
 
+// A dictionary to store debounces
+//
+const debounces = {}
+
 // Indicates if we should log calls to stimulate, etc...
 //
 let debugging = false
@@ -128,7 +132,17 @@ const extendStimulusController = controller => {
           }
         }
 
-        subscription.send(element.reflexData)
+        const debounceAttribute =
+          element.attributes[stimulusApplication.schema.reflexDebounceAttribute]
+        const debounceValue =
+          (debounceAttribute && debounceAttribute.value) || 0
+
+        clearTimeout(debounces[element])
+
+        debounces[element] = setTimeout(function () {
+          delete debounces[element]
+          subscription.send(element.reflexData)
+        }, debounceValue)
       })
 
       if (debugging) {


leastbad:
This is not so crazy, although I would like to see discussion of debounce vs. throttle vs. "delay".

Let me counter-propose something: 211 introduces a new options argument. What if we added debounce or throttle or delay (or some reasonable combination) as an argument for the folks calling via stimulate(), but didn't try to provide this via declarative markup?

My logic/reasoning is that if someone needs this, they should be calling an action on a Stimulus controller anyhow so that they can take proper advantage of callbacks.

I do frequently find myself wishing for a delay to let my CSS animations play before the 10ms morph update blows it away. Debounce would make beforeReflex() callbacks a lot more useful!

Note: Throttle actually makes sense in the context of #211's launch/none calls.

leastbad:
... man, imagine if it was possible to formalize a queue of received morphs that could be pulled off a stack based on their reflexId and executed when your animation completes (or really, whenever you choose; or never) instead of having to delay calling your reflex until after your beforeReflex animations complete.

Oh wait, don't we basically already do 80% of this, @hopsoft?

hopsoft:
Currently we put the onus on the developer to figure out a solution that works best for their app. I've thought about this quite a bit and agree that providing a facility to make this less painful for developers may be something worthy of adding to the library. 

I see the topics brought up here as 2 distinct concerns.

1. The ability to debounce high repeating events like `keyup` or `input`
2. The ability to introduce latency on calls to `stimuluate` to afford some UI enhancements like animations etc...

## 1 Debounce

I've considered a slightly different approach on this concern. Namely setting up a series of debounced standard DOM events. This could be created as a distinct NPM package (frankly I'm surprised that this doesn't already exist). Usage would look something like this.

```js
document.addEventListener('debounced:keyup', event => { /* logic... */ })
document.addEventListener('debounced:input', event => { /* logic... */ })
```

```html
<input type="text" data-reflex="debounced:input->ExampleReflex#do_stuff">
```

We could add something like this directly to StimulusReflex, but I think a better solution might be a separate NPM package that we pull in as a dependency. Or perhaps we simply document how to use it rather than pulling in the dependency.

## 2 Latency

Rather than adding intentional latency, I think refactoring the **before** lifecycle event to use a callback or promise might be the best solution here. i.e. We'd only `send` data to the server after the **before** lifecycle logic completes and invokes the callback/promise.

leastbad:
https://discordapp.com/channels/629472241427415060/629472241427415062/716628214335930398

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event

davidalejandroaguilar:
I like this.

I find myself trying to avoid writing custom stimulus controllers as much as possible. 

Having the `data-reflex` is already leaning on that direction, instead of only providing the `stimulate` method and having people always create Stimulus controllers.

I was just trying to think of a way of intercepting all reflexes and kinda debouncing / queueing them if they were an `input` event to avoid adding debounce everywhere. But then I remembered this PR.

What I have now is a `debounce_controller.js`. But it's a bunch of boilerplate to set up when you need it.  A quick data attribute helper would be nice. 

```erb
 <textarea
  data-controller="debounce"
  data-action="input->debounce#perform"
  data-debounce-reflex-name="SomeReflex"
  data-debounce-reflex-action="update"
  data-debounce-delay="200"
><%=title %></textarea>
```
vs

```erb
 <textarea
  data-reflex="input->SomeReflex#update"
  data-reflex-debounce="150"
><%=title %></textarea>
```

kylefox:
I love the idea of being able to debounce reflexes using only markup ❤️ 

We have a ton of Stimulus actions that exist purely to debounce reflexes. The ability to debounce with markup would allow us to eliminate a bunch of JavaScript.

```html
<input type="text" data-reflex="debounced:input->ExampleReflex#do_stuff">
```

@hopsoft ☝️ IMO this is the cleanest syntax, but it might cause issues with custom events. It's possible (and a common convention) to namespace events using `:` (like `ajax:success`) so `debounced:input` might be ambiguous.

I really like the idea of a `data-reflex-debounce="500"` attribute because it doesn't interfere with the `data-reflex` syntax, and allows us to easily specify different debounce delays:

```html
<!-- No debouncing -->
<input type="text" data-reflex="input->ExampleReflex#do_stuff">

<!-- Default debouncing (50ms?) -->
<input type="text" data-reflex="input->ExampleReflex#do_stuff" data-reflex-debounce>

<!-- Explicit debouncing (250ms) -->
<input type="text" data-reflex="input->ExampleReflex#do_stuff" data-reflex-debounce="250">
```

This syntax won't work when you have multiple reflexes on the same element that must be debounced differently from one another. It's _probably_ fine for StimulusReflex to not support this scenario — if someone requires more flexibility than ``data-reflex-debounce` provides, they can simply write their own Stimulus controller and debounce however they wish.

Regarding the implementation in this pull request: there's already a [`utils/debounce`](https://github.com/hopsoft/stimulus_reflex/blob/master/javascript/utils.js#L25-L34) function — maybe we could use that?

hopsoft:
@kylefox See: https://github.com/hopsoft/debounced I’m leaning toward this type of global solution rather than adding this type of functionality to SR directly.

kylefox:
@hopsoft Yep, your `debounced` library works tremendously well 👍

And I'm not concerned about the `debouced:<event>` syntax or lack of granular wait times after seeing how this works.

If I add a note about this library/technique to the README, could we close this issue?

hopsoft:
Closing this PR in favor of having people use the debounced JS lib. https://github.com/hopsoft/debounced

PR #214, opened 2020-05-26
Update dataset handling and some minor refactoring to better naming

hopsoft:
## Description

Update dataset API for element on the Ruby side to support access like `data_example` for combined dataset values.

Also refactored some naming around recursive `with_indifferent_access` handling of arguments.

## Checklist


lib/stimulus_reflex/element.rb
```diff
@@ -5,11 +5,9 @@ class StimulusReflex::Element < OpenStruct
 
   def initialize(data = {})
     @attributes = HashWithIndifferentAccess.new(data["attrs"] || {})
-    @data_attributes = (data["dataset"] || {}).each_with_object(HashWithIndifferentAccess.new) { |(key, value), memo|
-      memo[key.delete_prefix("data-")] = value
-    }.freeze
-
-    super attributes.merge(data_attributes).transform_keys(&:underscore)
+    @data_attributes = (data["dataset"] || {}).select { |key, _| key.start_with? "data-" }
```
marcoroth, line 8: I think this part shouldn't be necessary since `data["dataset"]` only contains `data-*` attributes.
```
.select { |key, _| key.start_with? "data-" }
```

Or does it have a different reason?

hopsoft: Hm...

```ruby
data["dataset"]
=> {"data-controller"=>"geo-selector",                                                                                                                                                               
 "data-test"=>"example",
 "data-reflex"=>"change->GeoSelectorReflex#select_region",
 "data-reflex-dataset"=>"combined",
 "data-action"=>"change->geo-selector#__perform"}
```

Issue #225, opened 2020-06-04
Lifecycle methods only called for one reflex


davidalejandroaguilar:
# Bug Report

## Describe the bug

When you setup two reflexes within the same container using the Stimulus `data-controller` attribute like [the docs](https://docs.stimulusreflex.com/lifecycle#generic-lifecycle-methods) example, only the first declared reflex lifecycle methods will get called.

In the HTML below, if you click on the `TestReflex` button, you will trigger the `OtherReflex` lifecycle method.

```html
<div data-controller="other test">
  <button data-reflex="click->TestReflex#create">TestReflex</button>
  <button data-reflex="click->OtherReflex#create">OtherReflex</button>
</div>
```

By the way, I encountered this when declaring some generic reflexes / controllers on the `body` tag, e.g. `<body data-controller="debounced ui">`. I wanted to use lifecycle methods on one of them but they were not getting called.

## To Reproduce

1. Create an HTML document like this:

```html
<div data-controller="other test">
  <button data-reflex="click->TestReflex#create">TestReflex</button>
  <button data-reflex="click->OtherReflex#create">OtherReflex</button>
</div>
```

2. Setup your client side reflexes like this:

```js
// test_controller.js
import ApplicationController from './application_controller';

export default class extends ApplicationController {
  beforeReflex() {
    alert('from TestController beforeReflex');
  }

  beforeCreate() {
    alert('from TestController beforeCreate');
  }
}
```

```js
// other_controller.js
import ApplicationController from './application_controller';

export default class extends ApplicationController {
  beforeReflex() {
    alert('from OtherController beforeReflex');
  }

  beforeCreate() {
    alert('from OtherController beforeCreate');
  }
}
```

3. Setup your `ApplicationController` like this:

```js
import { Controller } from 'stimulus';
import StimulusReflex from 'stimulus_reflex';

export default class extends Controller {
  connect() {
    StimulusReflex.register(this);
  }

  beforeReflex(element, reflex) {
    alert('from ApplicationController');
  }

  reflexSuccess(element, reflex, error) {}

  reflexError(element, reflex, error) {}

  afterReflex(element, reflex) {}
}
```

## Expected behavior

* Clicking on `OtherReflex` button triggers `OtherReflex`.
* Clicking on `TestReflex` button triggers `TestReflex`.

## Screenshots or reproduction

* Video:
https://imgur.com/a/mwYRYMB

* Repo
https://github.com/davidalejandroaguilar/reflex_bug

[...]

PR #226, opened 2020-06-04
Fix calling wrong controller lifecycle methods

davidalejandroaguilar:
# Bug fix

## Description

When selecting the controller for a specific reflex, we were always selecting the first one. This caused us to run the lifecycle methods of the wrong controller if the user specified other controllers via `data-controller`. This commit finds the correct controller and if not found, it uses the default stimulus controller as before.

Some pseudocode examples of what was happening:

Before:

```js
// Given these controllers (identifiers here)
const controllers = ['sortable', 'ui', 'autogrow', 'test']
// When reflex name was
const reflex = 'click->TestReflex#create'
// We were doing
controllers[0]
// Which selected the wrong controller. i.e. 'sortable' instead of 'test'
```

```js
// Given these controllers (identifiers here)
const controllers = ['sortable', 'ui', 'autogrow', 'test']
// When reflex name was
const reflex = 'click->TestReflex#create'
// We now do
controllers.find(controller => controller.identifier === reflexName)
// Which selects the correct controller. i.e. 'test'
```

Fixes https://github.com/hopsoft/stimulus_reflex/issues/225

## Checklist

## Notes

I'm marking this as WIP as I'm not sure which branch this should go to, or other build steps.

### DIFF
--- a/javascript/utils.js
+++ b/javascript/utils.js
@@ -32,3 +32,9 @@ export const debounce = (callback, delay = 250) => {
     }, delay)
   }
 }
+
+export const extractReflexName = reflexString => {
+  const match = reflexString.match(/(.*->)?(.*)Reflex/)
+
+  return match ? match[2] : ''
+}
--- a/javascript/utils.js
+++ b/javascript/utils.js
@@ -34,7 +34,7 @@ export const debounce = (callback, delay = 250) => {
 }
 
 export const extractReflexName = reflexString => {
-  const match = reflexString.match(/(.*->)?(.*)Reflex/)
+  const match = reflexString.match(/(?:.*->)?(.*?)(?:Reflex)?#/)
 
-  return match ? match[2] : ''
+  return match ? match[1] : ''
 }
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -14,6 +14,7 @@ import {
   extractElementDataset,
   findElement
 } from './attributes'
+import { extractReflexName } from './utils'
 
 // A reference to the Stimulus application registered with: StimulusReflex.initialize
 //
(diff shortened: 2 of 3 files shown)


javascript/utils.js
```diff
@@ -32,3 +32,9 @@ export const debounce = (callback, delay = 250) => {
     }, delay)
   }
 }
+
+export const extractReflexName = reflexString => {
+  const match = reflexString.match(/(.*->)?(.*)Reflex/)
```
leastbad, line 37: Heads-up: we've made the trailing `Reflex` string fragment optional, because it's redundant. See https://github.com/hopsoft/stimulus_reflex/blob/895ee97bcdbe8d80836f177d47a1668be041f61e/lib/stimulus_reflex/channel.rb#L25

leastbad:
@davidalejandroaguilar I'd love to get this merged this weekend, but you have to update the regex to make it accept both "click->CommentReflex#post" AND "click->Comment#post".

davidalejandroaguilar:
Hey @leastbad , I've modified the regex to account for these cases:

* click->TestReflex#create
* TestReflex#create
* click->Test#create
* Test#create

I added a basic Regex explanation within the commit as well. Let me know if I'm missing something or if you need me to reorganize /squash commits.

Sorry for the huge delay, it's been very busy weeks and I'm struggling to find time to devote to projects I like.

jasoncharnes:
We're using custom controllers that don't match the reflex name:

- **Ruby**: `app/reflexes/cms/campaigns_reflex.rb`
- **Stimulus**: `app/javascript/cms/campaigns/edit_controller.js`

When I point our libraries to `master` we lose all the lifecycle methods we defined because it doesn't see our controller now.

Would it be acceptable for me to open a PR that falls back to `controllers[0]` if a match isn't found?

leastbad:
@jasoncharnes that **sucks**. Let's get you sorted out.

So I assume that Ruby doesn't really care where the Reflex is, but Stimulus is dealing with a `campaigns--edit` controller? Is there really nothing we can do to be super smart about this?

In the meantime, PR please!

PR #230, opened 2020-06-08
adds params documentation

RolandStuder:
# Documentation for Params

## Description

Adds minimal description of `params` usage in ReflexControllers and how to change params before submitting

"Fixes" missing documentation https://github.com/hopsoft/stimulus_reflex/pull/204#issuecomment-632096254

## Where to put this?

Not sure, if this is really the right place to put it. As the javascript example does not really fit into the section of the Reflex Class description. I feel like this documentation does not fit well with the rest, should there maybe a separate page specifically on form handling?

Please let me know, how I can best help to document this, or whether you would prefer to do this yourself.


leastbad:
Thanks for getting this started, Roland! I agree that it doesn't fit super-well into the current structure.

I plan to spend most of this weekend on documentation and I'll keep you posted on how things go.

PR #234, opened 2020-06-15
__perform had a bug where it was only ever calling the first event

leastbad:
# bug fix

## Description

We accept multiple events routing to multiple reflexes, but somehow a bug slipped in that was only catching the first event

## Why should this be added

Obvious bug

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -173,8 +173,11 @@ const extendStimulusController = controller => {
         if (!reflex || !reflex.trim().length) element = element.parentElement
       }
 
-      attributeValues(reflex).forEach(reflex =>
-        this.stimulate(reflex.split('->')[1], element)
+      this.stimulate(
+        attributeValues(reflex)
+          .find(reflex => reflex.split('->')[0] === event.type)
+          .split('->')[1],
+        element
       )
     }
   })


javascript/stimulus_reflex.js
```diff
@@ -173,8 +173,11 @@ const extendStimulusController = controller => {
         if (!reflex || !reflex.trim().length) element = element.parentElement
       }
 
-      attributeValues(reflex).forEach(reflex =>
-        this.stimulate(reflex.split('->')[1], element)
+      this.stimulate(
+        attributeValues(reflex)
+          .find(reflex => reflex.split('->')[0] === event.type)
+          .split('->')[1],
```
foca, line 179: Oh, thanks so much for looking into this :heart:

One question (although I have absolutely no use case for this), but would this only call the first reflex of each event type? Is this something that should be supported?

```
data-reflex="click->SomeReflex#blah click->OtherReflex#bleh"
```

🤔

leastbad:
Yes, I think so.

I actually just added documentation for multiple reflex events per element **yesterday**. Before that, you had to look at the code or assume it worked - and by the time you got there, it’d been broken for an indeterminate amount of time.

In short, this is not something a lot of folks have been asking for and just being able to support mouseenter+mouseleave is precisely perfect. I have observed an awkward tendency for developers to fall down rabbit holes where the laws of plausible vs worth it get temporarily suspended. 😺

I am happy to tell people that want three click handlers on one element to just use javascript.

javascript/stimulus_reflex.js
leastbad: Yes, I think so.

I actually just added documentation for multiple reflex events per element **yesterday**. Before that, you had to look at the code or assume it worked - and by the time you got there, it’d been broken for an indeterminate amount of time.

In short, this is not something a lot of folks have been asking for and just being able to support mouseenter+mouseleave is precisely perfect. I have observed an awkward tendency for developers to fall down rabbit holes where the laws of plausible vs worth it get temporarily suspended. 😺

I am happy to tell people that want three click handlers on one element to just use javascript.

PR #236, opened 2020-06-16
Add a queue for `#stimulate` calls

dark-panda:
# Enhancement

## Description

This allows you to build up `#stimulate` calls until the ActionCable connection is actually made and then flush them all once the connection is established. As a side effect, you can now call `#stimulate` in `#connect` methods on Stimulus controllers.

## Why should this be added

By using a queue, you can begin making calls with `#stimulate` as soon as SR is initialized rather than waiting for the ActionCable connection to be made, which depending on latency, resources, etc. may take a little while after page load. If user interaction causes a `#stimulate` call before ActionCable is connected, we'll queue the call in a closure until the connection is made and then flush all of such calls in the order received. As an added bonus, you can now call `#stimulate` in `#connect` methods, since we don't need to wait for ActionCable to be connected, even if it isn't considered a best practice.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -34,6 +34,10 @@ let actionCableSubscriptionActive = false
 // A dictionary of promise data
 const promises = {}
 
+// A queue we can use during initialization to make sure the ActionCable
+// connection is ready to go before we start sending `#stimulate` calls.
+const stimulateQueue = []
+
 // Indicates if we should log calls to stimulate, etc...
 let debugging = false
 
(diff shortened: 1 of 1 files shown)


bricesanchez:
Hi everyone! Any update to merge this feature?

leastbad:
It's in review, sir!

We're volunteers, doing our best. Lots of moving parts.

leastbad:
Sorry to leave this one hanging for so long.

Long story short: v4 - which is a significant re-write of the client - will have a really powerful Reflex queueing system. It won't make it into v3.5, which is coming soon.

The actual implementation of the queueing is wildly different than what is proposed in this PR, so I'm going to close it with sincere thanks to @dark-panda for making it.

PR #241, opened 2020-06-19
Update install.rake to handle Typescript

iv-mexx:
# Enhancement / Bug Fix

## Description

When using typescript, you will have `index.ts` instead of `index.js` or `application.ts` instead of `application.js`. 

Prior to this change, running `bundle exec rails stimulus_reflex:install` fails because neither `index.js` nor `application.js` will be found.

Now with this change, it will work when using typescript, the contents appended to these files work in Typescript without a problem.

## Why should this be added

This reduces install friction for Typescript users.

## Checklist

### DIFF
--- a/lib/tasks/stimulus_reflex/install.rake
+++ b/lib/tasks/stimulus_reflex/install.rake
@@ -11,11 +11,16 @@ namespace :stimulus_reflex do
     FileUtils.mkdir_p Rails.root.join("app/javascript/controllers"), verbose: true
     FileUtils.mkdir_p Rails.root.join("app/reflexes"), verbose: true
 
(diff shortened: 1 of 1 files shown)


leastbad:
I support the initiative, but can you please structure your logic in a way that isn't so confusing? It's perfectly okay to be a bit verbose if necessary; this is a bad place for cleverness or ambiguity.

Even if this works, it makes it extremely brittle for future things. Like the thing that you all decide is better than TypeScript later this year. ;)

iv-mexx:
Sure, better this way?

leastbad:
F yeah, Markus! That's awesome. I approve.

iv-mexx:
Great, thanks!

PR #243, opened 2020-06-23
Support for token-based authentication

leastbad:

Enhancement

## Description

This PR proposes a small set of changes which would allow JWT (and other token-based systems) to authenticate ActionCable channels, including StimulusReflex.

Discussions with @king601 led deep down a "what if" rabbit hole to a place where developers use JWT tokens to authenticate at the channel subscription level. Thankfully a stack overflow post led to a pivotal conversation with @phikes who gave me enough code to hang myself with. Lucky for me, @ParamagicDev (with suggestions from @StephenFiser) was able to bail the effort out. It was looking really grim for a few hours.

I've done my very best to document the process at https://docs.stimulusreflex.com/authentication#token-based-authentication and you can clone a working example repo at https://github.com/leastbad/testes (although you'll need this PR checked out and running in order to test it).

## High-level details

- In order to support one authentication mechanism that could be inherited by both SR and application channels, we had to do a pretty ugly require in `channel.rb`
- This PR adds an optional `params` option to StimulusReflex.initialize()
~~- I took the opportunity to add `stimulus-reflex:connected` and `stimulus-reflex:disconnected` events alongside `stimulus-reflex:rejected` and I'm **not attached**~~
- stimulate() will throw an exception if the subscription was rejected
- I did briefly attempt to implement a connect/disconnect logic for createSubscription() before thankfully realizing I was achieving escape velocity and pulled it all out
- if ApplicationCable is not defined in the app (it comes with the ActionCable framework) then it will just skip the require

## Checklist


lib/stimulus_reflex/channel.rb
```diff
@@ -1,6 +1,18 @@
...
+    def initialize(connection, identifier, params = {})
+      super
+      application_cable = Rails.root.join("app", "channels", "application_cable", "channel.rb")
+      require application_cable if File.exist?(application_cable)
+    end
+  end
+end
+
+class StimulusReflex::Channel < ApplicationCable::Channel
+  include CableReady::Broadcaster
```
julianrubisch, line 14: Probably I’m not seeing correctly as I’ve got no computer available to try it out, but why did you re-include `CableReady::Broadcaster` here?

lib/stimulus_reflex/channel.rb
```diff
@@ -1,6 +1,18 @@
 # frozen_string_literal: true
 
-class StimulusReflex::Channel < ActionCable::Channel::Base
+module ApplicationCable
+  class Channel < ActionCable::Channel::Base
+    def initialize(connection, identifier, params = {})
+      super
+      application_cable = Rails.root.join("app", "channels", "application_cable", "channel.rb")
```
julianrubisch, line 7: I’d maybe rename this to `application_channel`, since you’re looking up `application_cable/channel.rb`, right? Might be somewhat ambiguous otherwise...

lib/stimulus_reflex/channel.rb
leastbad: This significantly predates the recent significant refactoring. It's entirely possible it's included more than it needs to be.

leastbad: You are correct - it doesn't seem to need to be in the channel anymore. Originally, it didn't need to be in the reflex. Things have changed!

PR #248, opened 2020-06-26
Update deployment.md

user073:
Improve documentation

## Description

Fix typo

Fixes # (issue)

## Why should this be added

Explain value.

## Checklist


leastbad:
That's a really interesting one. I actually don't know who's right here.

Does are become is because team implies a group of people?

What if you swap out "team" for "robot" or "woman"? Is remains the obvious choice. So what about team makes are become correct? This is not a rhetorical question - I actually do not know.

leastbad:
You are correct!

https://www.grammarly.com/blog/is-vs-are/

PR #252, opened 2020-06-29
Fix cases where plural reflexes were unresolved

joshleblanc:
## Type of PR (feature, enhancement, bug fix, etc.)

fix

## Description

If you have a reflex with a plural name like `CampaignsReflex`, and call it with `click->Campaigns#do_something`, `"Campaigns"` will incorrectly be resolved to `"Campaign"`.

From here: https://discordapp.com/channels/629472241427415060/629472241427415062/727195517351886968

## Checklist


leastbad:
Good catch. This wouldn't have happened before #212, so technically my fault for not anticipating this.

PR #256, opened 2020-07-01
Limit MutationObserver mutations

### LINKED ISSUE #251: Select with data-reflex in Firefox flickers
# Bug Report

## Describe the bug

If I have `data-reflex="change->CampaignReflex#do_something"` on a `<select>` in Firefox, I get the following functionality:

https://share.getcloudapp.com/YEuprmPB

If I change the `data-reflex` call into a `data-action"campaign#doSomething` which calls `this.stimulate` under the hood, the select functions as expected.

## To Reproduce

* Add `data-reflex="change->YourReflex#do_thing"` to a `<select>`
* Turn on your choice of uptempo music (EDM, Rave, Heavy Metal)
* Open said select in Firefox
* Watch the select ~party to the music~ not allow you to select an option, but rather flicker to the selected option continuously

## Expected behavior

A select with a `data-reflex` should function like a rather HTML select. 

https://share.getcloudapp.com/YEuprmPB

The video above is from our application. I also spun up a separate Rails app to make sure it wasn't something ~stupid~ creative we've done in our app code.

❤️ you all!

### StimulusReflex

- Gem: [3.2.3]
- Node package: [3.2.3]

### External tools

- Ruby: [2.6.6]
- Rails: [6.0.3]
- Node: [13.11.0]

### Browser

- Browser [Firefox 🔥 ]
- Version [77.0.1]

jasoncharnes:
# Bug Fix, Maybe?

## Description

**This is a theory, with a potential solution. I could be way off.**

The current `MutationObserver` fires in an endless loop, because `setupDeclarativeReflexes` (the code called from the observer) is making a mutation, hence rerunning `setupDeclarativeReflexes`.

Safari and Chrome seem to handle this ok. In Firefox, however, it causes a nasty flicker on selects. If you go to open a select dropdown, it never lets you choose another option, it continues to re-select the original value.

If you `console.log` anything inside of `setupDeclarativeReflexes` it causes Firefox (on my machine) to lock up within a second, requiring a force quit.

I believe this is the section of code causing the flicker, because if you stop the observer the flicker stops (but so does the SR functionality, obviously). If you bump the `debounce` time for `setupDeclarativeReflexes` to something higher, the flicker still happens, but at a slower pace.

The `setupDeclarativeReflexes` function seems to do one of two mutating things:

1. Adding `stimulus-reflex` to `data-controller`
2. Adding `[event]->[Reflex]#__perform` to `data-action`

I believe we can stop the endless loop by observing the mutations and not rerunning `setupDeclarativeReflexes` if either of these two mutations performed.

Fixes #251 

## Why should this be added

This feels like... not what we want? 

Unfortunately, my `MutationObserver` knowledge is laughable, as we've only used it a handful of times in our own application. But, I wanted to do my part to try and help solve this issue.

I've forked this JS code into a repo to run against our app and _**so far**_ it appears to be working. Granted, our app doesn't cover all the other ways people use Stimulus Reflex. 😅 

As **always**, I'm open to suggestions and making changes.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -244,13 +244,21 @@ const setupDeclarativeReflexes = debounce(() => {
       })
       const controllerValue = attributeValue(controllers)
       const actionValue = attributeValue(actions)
-      if (controllerValue) {
+      if (
+        controllerValue &&
+        element.getAttribute(stimulusApplication.schema.controllerAttribute) !=
+          controllerValue
+      ) {
         element.setAttribute(
           stimulusApplication.schema.controllerAttribute,
           controllerValue
         )
       }
-      if (actionValue)
+      if (
+        actionValue &&
+        element.getAttribute(stimulusApplication.schema.actionAttribute) !=
+          actionValue
+      )
         element.setAttribute(
           stimulusApplication.schema.actionAttribute,
           actionValue
@@ -327,23 +335,11 @@ if (!document.stimulusReflexInitialized) {
 
   window.addEventListener('load', () => {
     setupDeclarativeReflexes()
-
-    const observer = new MutationObserver(mutations => {
-      mutations.forEach(mutation => {
-        const shouldMutate =
-          !mutation.oldValue ||
-          (mutation.oldValue != 'stimulus-reflex' &&
-            !mutation.oldValue.includes('__perform'))
-
-        if (shouldMutate) setupDeclarativeReflexes()
-      })
-    })
-
+    const observer = new MutationObserver(setupDeclarativeReflexes)
     observer.observe(document.documentElement, {
       attributes: true,
       childList: true,
-      subtree: true,
-      attributeOldValue: true
+      subtree: true
     })
   })
 
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -327,11 +327,23 @@ if (!document.stimulusReflexInitialized) {
 
   window.addEventListener('load', () => {
     setupDeclarativeReflexes()
-    const observer = new MutationObserver(setupDeclarativeReflexes)
+
+    const observer = new MutationObserver(mutations => {
+      mutations.forEach(mutation => {
+        const shouldMutate =
+          !mutation.oldValue ||
+          (mutation.oldValue != 'stimulus-reflex' &&
+            !mutation.oldValue.includes('__perform'))
+
+        if (shouldMutate) setupDeclarativeReflexes()
+      })
+    })
+
     observer.observe(document.documentElement, {
       attributes: true,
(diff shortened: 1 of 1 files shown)


leastbad:
How the hell did you even know `oldValue` was a thing? I swear it's not even mentioned in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserverInit).

leastbad:
Alright, I think I've got my head wrapped around this.

First, thank-you for transcending "I have a problem" to "...and here's a fix for it". You are a model OSS citizen. I also really appreciate that you take the peacemaker stance, almost apologizing for fixing problems sometimes.

In this case, I think you give us too much credit and this might have led to you assuming that we had any idea we'd introduced an infinite loop. I certainly did not realize, and I have a hunch that @hopsoft didn't when he wrote this function.

I do think I therefore have to disagree with your fix (although I'm happy to be schooled if I'm missing something!) because what you're doing is introducing a sane but ultimately brittle exception to the current MutationObserver handler. Instead, wouldn't it be better to prevent the infinite loop in the first place?

So far as I can tell from looking at the `setupDeclarativeResources` function, it's lines 248 and 254 in `stimulus_reflex.js` that are setting the data-controller and data-action attributes over and over, causing the loop. Wouldn't we be better off with something like:

```js
      if (controllerValue && element[stimulusApplication.schema.controllerAttribute] != controllerValue) {
        element.setAttribute(
          stimulusApplication.schema.controllerAttribute,
          controllerValue
        )
      }
      if (actionValue && element[stimulusApplication.schema.actionAttribute] != actionValue)
        element.setAttribute(
          stimulusApplication.schema.actionAttribute,
          actionValue
        )
```

Also: @hopsoft I could be tired but doesn't the current line 248 and 254 of `stimulus_reflex.js` actually replace any other controllers present on the element at the time the `stimulus-reflex` attribute being placed? This would seem like it should break a huge number of things, especially in a world where there's a healthy selection of composable controllers people might be using on an element.

jasoncharnes:
> what you're doing is introducing a sane but ultimately brittle exception to the current MutationObserver handler. Instead, wouldn't it be better to prevent the infinite loop in the first place?

💯 I was grasping at straws and couldn't seem to grasp enough context to figure out _how_ to prevent the loop.

> Wouldn't we be better off with something like: (code that actually fixes the root problem)

YES! This is exactly what I wanted to do. 🎉 I had to make a slight modification, but it works like a dream. ❤️ 

Thanks a ton @leastbad, I'm _really_ happy to get this sorted out! 🎉

hopsoft:
@leastbad Lines 248 and 254 do replace the current value but `setupDeclarativeReflexes` also preserves the original values and simply appends the changes to existing values for these attributes.

PR #257, opened 2020-07-01
Fallback to first Stimulus controller in array

jasoncharnes:
# Bug Fix

## Description

#226 does the great work of automatically using a matching Stimulus controller name with a matching Reflex. 👏🎉 

Unfortunate for us, is that we do some aerobatics with our naming and namespaces. Our Stimulus controllers use a custom name that doesn't match the reflex. This means that `findControllerByReflexString` returns `null` and we lose all our JS callback functionality.

I think there's a middle ground until we can [figure out a way around this](https://github.com/hopsoft/stimulus_reflex/pull/226#issuecomment-652549495). We can default to looking for a matching controller name, and falling back to `controllers[0]` (as done before) if a match isn't found.

## Why should this be added

This was a pretty unexpected, surprising change in functionality for us. At most, I'd like to continue using custom Stimulus controller names. At minimum, I'd like to put off a requirement for matching controller/reflex names for just a bit if possible. 🙏 

I think this change gives priority to those who have matching Stimulus Reflex controller/reflex names, while still giving the rest of us room to make messes. 😉 

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -263,7 +263,7 @@ const setupDeclarativeReflexes = debounce(() => {
 // identifier. e.g. Given these controller identifiers ['foo', 'bar', 'test'],
 // it would select the 'test' controller.
 const findControllerByReflexString = (reflexString, controllers) => {
-  return controllers.find(controller => {
+  const controller = controllers.find(controller => {
     if (!controller.identifier) return
 
     return (
@@ -271,6 +271,8 @@ const findControllerByReflexString = (reflexString, controllers) => {
       controller.identifier.toLowerCase()
     )
   })
+
+  return controller || controllers[0]
 }
 
 // compute the DOM element(s) which will be the morph root


leastbad:
Apologies for the switch-a-roo. It just goes to show that any codebase with enough complexity slowly becomes Jenga.

I think that this is a perfectly reasonable compromise.

For what it's worth, I'd still love to know the identifier that comes with `app/javascript/cms/campaigns/edit_controller.js` - is it `campaigns--edit` or something else? Ideally I'd love to land on a solution that just works, regardless of folder structure and naming.

jasoncharnes:
> For what it's worth, I'd still love to know the identifier that comes with app/javascript/cms/campaigns/edit_controller.js - is it campaigns--edit or something else? Ideally I'd love to land on a solution that just works, regardless of folder structure and naming.

The identifier is `campaigns--edit` 👍

leastbad:
So here's what I'm thinking; maybe this is a hack. But if you like the idea, PR please:

What if we did something like `controller.identifier.toLowerCase().split('--')[-1]` except in valid JS because I don't think JS supports -1 indexes? eg. you might have to
```js
    const jizz = controller.identifier.toLowerCase().split('--')
    return (extractReflexName(reflexString).toLowerCase() === jizz[jizz.length - 1])
```

PR #259, opened 2020-07-04
Prep for pre release of 3.3.0

hopsoft:


lib/stimulus_reflex/broadcaster.rb
```diff
@@ -1,10 +1,6 @@
 module StimulusReflex
   module Broadcaster
-    def self.included klass
-      klass.class_eval do
-        include CableReady::Broadcaster
-      end
-    end
+    include CableReady::Broadcaster
```
leastbad, line 3: @julianrubisch I'm trying to figure out how we were confused about this (and I do mean "we")

julianrubisch: Partially because I couldn’t test it and believed stackoverflow about including modules in modules 🤷‍♂️

PR #260, opened 2020-07-06
conditionally set children_only in morph

joshleblanc:
This is related to issue 1 from discord. 

> It's starting to feel like :selector morphs should be performed as children_only: false which in real terms is like shifting from modifying innerHTML to outerHTML - unlike :page morphs, the container element would be re-written. This is because we want re-rendering partials and components to follow the principle of least surprise; right now if you re-render a partial inside a container you'll end up with a 2nd copy of the container element in the container element. Note: morphdom gets weird about replacing the body tag, so if you morph "body" it will still be children only.

I didn't think just setting children_only to false for selector morphs was really the right solution because that's not strictly the problem. The *problem* if if you have some selector, say `#foo`, and you set the html of this component to `<div id="foo">..</div>`, then you now have conflicting elements in your DOM that is going to confuse all of your selectors. In this specific case, you want children_only to be false. 

So that's what this PR tries to do. 

There's a related PR in for `cable_ready` that I had to make to even get this to work with `children_only: false`. https://github.com/hopsoft/cable_ready/pull/58

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -108,10 +108,11 @@ class StimulusReflex::Reflex
   end
 
   def enqueue_selector_broadcast(selector, html)
+    document = Nokogiri::HTML(html)
     cable_ready[channel.stream_name].morph(
       selector: selector,
       html: html,
-      children_only: true,
+      children_only: !document.css(selector).present?,
       permanent_attribute_name: permanent_attribute_name
     )
   end


leastbad:
I'd really hoped to not have to bring Nokogiri into this method. It will slow things down.

Also, this solution doesn't factor in the outlier case where body is concerned. children_only needs to be true if the selector is "body"

I'm still trying to wrap my head around this. Forgive my pedantry but I believe you're confusing *problem* and *emergent behaviour* in this scenario. We have this black box and it behaves in nuanced ways depending on which levers you pull in what order.

We need to do a better job of thinking through all permutations of what could happen. For example, what if your html fragment doesn't have one top-level element? What if it does, but it contains a different id or selector?

```html
<div id="sam">content</div>
```
```rb
morph "#sam", '<div id="jim">better content</div>`
```

leastbad:
Also, I have a bad feeling about treating things differently based on the presence of a selector. Why should these fake partials have different code paths?

```html
<div id="sam">content</div>
```
```rb
morph "#sam", '<div id="sam">better sam content</div>'
morph "#sam", '<div id="jim">better jim content</div>'
morph "#sam", "I am just anonymous and better"
```

leastbad:
Or how about:

```html
<div id="sam">content</div>
```
```rb
morph "#sam", 'I want to kick things off with some text!<div id="sam">better sam content</div>And end on the same note.'
```
It's a real party.

joshleblanc:
Might be a better option to just pass children_only to the method, optionally.

As an aside: are selectors expected to be unique?

leastbad:
> Might be a better option to just pass children_only to the method, optionally.

At this point, I'm definitely open to morph having an optional third parameter that controls children_only. However, the burden of selecting the most obvious default still falls on us, or we doom people to writing verbose code more often than not.

leastbad:
There's only supposed to be one selector; it's going to be explained in the [eventual] documentation that only the first match will be relevant.

leastbad:
This can probably get closed, now...

leastbad:
Thanks for your help with this!

PR #261, opened 2020-07-07
potential fix

leastbad:

Bug fix

## Description

Decide between morph with children, morph without children (partials) or inner_html based on the following pseudo-Ruby:

```rb
return morph children_only: true if selector == "body"
if html.empty? or html.start_with("<") == false
  return inner_html.
else
  doc = nokogiri(html)
  return morph children_only: false if doc.top_level_tag.selector == selector
  return morph children_only: true
end
```

In the end, I didn't attempt to provide a 3rd parameter argument because there really is only one condition where you'd want to use `children_only: false` and if you're that one weirdo in a thousand that has an exception, you can always call CableReady directly.

Fixes #211 

## Why should this be added

Put me out of my misery

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -114,7 +114,7 @@ class StimulusReflex::Reflex
       children_only: true,
       permanent_attribute_name: permanent_attribute_name
     }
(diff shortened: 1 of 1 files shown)


hopsoft:
This seems reasonable to me. I like that you're still optimizing for performance and only using Nokogiri parsing when necessary.

PR #262, opened 2020-07-07
alternate solution

leastbad:

bug fix

## Description

Once I finished the previous implementation, of course the refactoring started. This solution is actually inspired by what @joshleblanc proposed for CableReady.

First, I am not keen on using inner_html because it doesn't respect `data-reflex-permanent` in any case. So I decided to just go with it and work out a version that uses Nokogirl for all calls.

The breakthough came from realizing that it's not replacing the container that's important, but modifying the content depending on the context. If you're updating a partial/component, you want fragment.inner_html, otherwise fragment.to_html. Therefore, **this version does not require children_only: false**

The next thing that I've figured out is that the data-reflex-permanent bug has been there all along! We were just never stumbling into it.

In order for the `data-reflex-permanent` to work, two things have to be true: first, **it just does not work on text nodes**. You have to wrap all text in at least a `span` or else it will just overwrite even if it's marked as permanent. The second detail seems to be that it only respects `permanent` if the selector is the same.

Note: I'm basically falling asleep on my keyboard so I intend to finish my research tomorrow, but I wanted to share now so people don't waste time.

There's no getting around it: we're going to have to give people some rules/best practices for calling morphs. Either that or we can detect text and wrap it in a `span` which might not be a welcome addition.

Fixes #211

## Why should this be added

For future generations

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -108,9 +108,11 @@ class StimulusReflex::Reflex
   end
 
   def enqueue_selector_broadcast(selector, html)
+    fragment = Nokogiri::HTML(html)
+    parent = fragment.at_css(selector)
     cable_ready[channel.stream_name].morph(
       selector: selector,
-      html: html,
+      html: parent.present? ? parent.inner_html : fragment.to_html,
       children_only: true,
       permanent_attribute_name: permanent_attribute_name
     )


hopsoft:
I'm really liking this solution and prefer it over #261. Would love to get some feedback from others in the community.

leastbad:
There were three apparent serious issues with v3.3.0.pre0 for selector morphs:

1. Updating partials and ViewComponents would create duplicate elements under the parent
2. `data-reflex-permanent` was being ignored
3. It was reported that post-Reflex callbacks (after/success/error) were not being emitted

This PR adds some logic to test whether you are attempting to re-render a partial/VC with new state vs just changing its content. As with full-page Reflexes with root targets, is not possible to use the `morph` helper to change the container element indicated in the selector. **If you need to change a container, either target the element's parent or use `CableReady#outer_html`.** Note that `outer_html` does not respect `data-reflex-permanent`.

I really went to the mat to figure out the precise behaviour of morphdom so that I could relay to you the specific rules that pertain to `data-reflex-permanent`. The interesting outcome of my research was the realization that the "bug" has always been present but the way StimulusReflex worked previously, we never hit it. There was a case where someone *could have* but it's never been previously reported.

Finally, I have verified that callbacks are working as **expected**. Please note that if you morph the originating element - usually the element with the Stimulus controller corresponding to your Reflex on it, but not always - out of existance, it cannot emit events. As DHH always says, this library gives you sharp knives and you have the responsibility to design your DOM hierarchy in such a way that you don't send the elements you need to `/dev/null`.

-----------------

What are the `data-reflex-permanent` rules? I'm glad that you asked, because I worked hard to get you this information.

It turns out that there are **three** requirements that must be met or else your elements will be over-written:

1. There has to be a `data-reflex-permanent` attribute on the top-level element you want to protect.
2. The top level of your update needs to be a container element that has to have the same selector as the original; this means that `<div id="bob">` has to be updated by `<div id="bob">` **or** `<span id="bob">` - I'll let you puzzle that one out. `<div id="jim">` will not respect the *permanent* attribute.
3. Once you've got the container in place, **the top-level child of your container has to be an element node, not a text node**. This means `<div id="bob"><span>good</span></div>` and `<div id="bob">bad</div>`. Stupid, right?

[onBeforeElUpdated](https://github.com/patrick-steele-idem/morphdom#faq) is simply never called unless all three conditions are met. In fairness, the **El stands for "Element"**.

To recap, if you want to keep your elements permanent, it has to look something like this:

`<div id="bob" data-reflex-permanent><span>Permanent</span></div>`

------------------

How did I arrive at conclusions after all of the speculation and conjecture? Well, first I created a decent but not great fix in PR #261. Then to fully test this PR, I created a **morphs branch** of my `stimulus_reflex_harness` app which you can clone here: https://github.com/leastbad/stimulus_reflex_harness/tree/morphs

The important and relevant files are the [Reflex class itself](https://github.com/leastbad/stimulus_reflex_harness/blob/morphs/app/reflexes/example_reflex.rb), which contains 40 assertions for 10 different fuzzing scenarios. Now that it's pointing at SR master branch, it should be quite approachable. The view which is generating `afterReflex` callbacks for all 10 morphs is [here](https://github.com/leastbad/stimulus_reflex_harness/blob/morphs/app/views/home/index.html.erb).

hopsoft:
Wow. Thanks for your thorough work on this. The trick now is to document the happy paths without scaring users away from this feature.

leastbad:
Who needs RSpec?!

So the good news here is that this fuzzing can be quickly explained away by focusing on what developers **should** do instead of what they shouldn’t. 

If they stick to some rules, they won’t have to think about this unholy miscarriage.

It’s not exactly a 2x2 grid but we can keep this as a troubleshooting appendage, deep in back.

leastbad:
Note to all of our future selves: somehow, we all forgot `morph "#text", 5`

Hopefully #268 will take care of this.

Sadly, we now have #269 to contend with...

PR #267, opened 2020-07-15
Update deployment docs after the official AnyCable 1.0 release

rmacklin:
### Description

AnyCable 1.0 has been officially released:
https://evilmartians.com/chronicles/anycable-1-0-four-years-of-real-time-web-with-ruby-and-go

In light of that, I've updated the docs to stop recommending installation of the release candidate version. Instead, the docs now suggest "1.0"; it didn't seem necessary to specify a patch version given that we probably don't want to have to update these docs every time a new patch version is released.


leastbad:
Thanks for this!

Issue #278, opened 2020-07-21
First argument of type "object" in this.stimulate() will always be assigned to options.


shawnleong:
# Bug Report

## Describe the bug

The first argument of type "object" being passed into `this.stimulate()` will always be assigned to `const options`.
The check to see if an object is an options object will always return true... even if the option keys are not present in the object.

This results in data["args"] being an empty array because the first argument of type object is shifted and assigned to `options`. 

The corresponding server side reflex expects an argument but is not passed any will raise the following error
`wrong number of arguments (given [], expected [[:req, :params]], optional [])`

## To Reproduce

Steps to reproduce the behavior

1. Create a stimulus controller based reflex
2. Pass in an object as the first argument with any key that's not used for options
3. Stimulate

**For example**
Client side stimulus controller
```
// app/javascript/controllers/example_controller.js
var params = { example: "object" }
this.stimulate("ExampleReflex#action", params)
```

Specific issue
```
// stimulus_reflex/javascript/stimulus_reflex.js
...
const options = {}
if (
  args[0] &&                                            // true
  typeof args[0] == 'object' &&                         // true
  Object.keys(args[0]).filter(key =>
    ['attrs', 'selectors', 'reflexId'].includes(key)
  )                                                     // This returns an array even if key is not present, which in turn allows the conditional to eval true
) {
  const opts = args.shift()
  Object.keys(opts).forEach(o => (options[o] = opts[o]))
}
...
```

## Expected behavior

Argument of type 'object' should not be assigned to `options` if it does not have any of the keys `['attrs', 'selectors', 'reflexId']` 
```
// stimulus_reflex/javascript/stimulus_reflex.js
...
const options = {}
if (
  args[0] &&                                         
  typeof args[0] == 'object' &&                       
  Object.keys(args[0]).filter(key =>
    ['attrs', 'selectors', 'reflexId'].includes(key)
  )                                                       
  .includes(true)                // quick fix
) {
...
```

Update: Added a Pull Request #279 

## Screenshots or reproduction

![Screenshot from 2020-07-21 15-02-05](https://user-images.githubusercontent.com/17049613/88023287-5f579e80-cb63-11ea-9155-56e4d82b44f2.png)

### StimulusReflex

- Gem: [3.3.0.pre2]
- Node package: [3.3.0.pre2]

### External tools

- Ruby: [2.6.3]
- Rails: [6.0.0]

### Browser

- Browser [chrome, firefox]

[...]

PR #279, opened 2020-07-21
Fix argument of type object always being assigned to options

shawnleong:
Bug Fix

## Description
The first argument of type "object" being passed into this.stimulate() will always be assigned to `const options`.

The check to see if an object is an options object will always return true... even if the option keys are not present in the object.

Added an additional check on the 3rd expression to check if there were any `true` values in the array returned by `Object.keys(args[0]).filter( ... )`.

Fixes issue #278 

## Why should this be added

Because arguments of type "object" end up being assigned to options when they are not supposed to.

(I'm assuming this is unintended based on the the documentation of ["Receiving arguments"](https://docs.stimulusreflex.com/reflexes#receiving-arguments) where a reflex can receive an object as the first argument.)

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -100,9 +100,9 @@ const extendStimulusController = controller => {
       if (
         args[0] &&
         typeof args[0] == 'object' &&
-        Object.keys(args[0])
-          .filter(key => ['attrs', 'selectors', 'reflexId'].includes(key))
-          .includes(true)
+        Object.keys(args[0]).filter(key =>
+          ['attrs', 'selectors', 'reflexId'].includes(key)
+        ).length
       ) {
         const opts = args.shift()
         Object.keys(opts).forEach(o => (options[o] = opts[o]))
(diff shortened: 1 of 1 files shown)


leastbad:
First, great catch. I'm honestly not sure how this snuck past me.

I'd like to counter-propose what I suspect will be a simpler fix:

```js
// stimulus_reflex/javascript/stimulus_reflex.js
...
const options = {}
if (
  args[0] &&
  typeof args[0] == 'object' &&
  Object.keys(args[0]).filter(key =>
    ['attrs', 'selectors', 'reflexId'].includes(key)
  ).length
) {
  const opts = args.shift()
  Object.keys(opts).forEach(o => (options[o] = opts[o]))
}
...
```
JS evaluates `0` as falsy, so throwing a .length on there seems like it would fix it, right?

shawnleong:
@leastbad  No worries, happy to help.

Ah yes, that's a much more elegant fix rather than having to iterate through a possible array of falsy values.

Wasn't thinking that deep past my own quick fix. Should have done better.

I've updated the pull request with the suggested fix. Thank you!

PR #291, opened 2020-08-15
Add instructions for existing projects

gerrywastaken:
# Docs

## Description

The sentence above this change mentions easily setting up StimulusReflex for new and existing rails projects, but then proceeded to only give instructions for new projects.

> You can easily install StimulusReflex to new and existing Rails projects.

This commit fixes that.

## Why should this be added

This helps people with adding Stimulus Reflux to existing projects

## Checklist


leastbad:
Thanks, Gerry!

Issue #294, opened 2020-08-31
Fail loudly if there's a version mismatch between the gem and the npm package.


noise-machines:
# Feature Request

Hi folks! Thanks for the great work you've done on Stimulus Reflex, and for the friendly and welcoming support I received in the Discord server.  

## Is your feature request related to a problem?

I recently got bitten by having a mismatched version of the gem (3.2.3) and the npm package (3.3.0.pre2).

## Describe the solution you'd like

I know you mention that mismatch in your very handy troubleshooting guide, but it'd be even more helpful if Stimulus Reflex checked for this scenario and complained loudly about it.

I thought I had confirmed they were at the same version, but I was mistaken. If SR complained loudly in this scenario, it would've saved me some time and helped me avoid some very strange error messages.

leastbad:
I'm sure that we'd be happy to implement this functionality if there's a reasonably approachable way to do it.

Could you do some research on how other projects with Ruby+JS components handle this and report back?

RolandStuder:
The PR #297 (so far) addresses half the issue. Running `rails stimulus_reflex:install` will currently install mismatched versions as `yarn` command used in that task will always install the latest version.

But mismatches between gem and npm package can come up, when you `bundle update` or `yarn upgrade` or otherwise adjust one of the versions.

I suggest we do a separate PR to address this.

---

I am not in love of doing a check in every reflex call or cable connection. I would prefer doing the checks in some form during the initialization of the rails app and then be more or less aggressive about informing the user.

Suggestion:

* use some form of initialization of SR to do the check.
* either raise an error (and provide a config option, to opt out of these errors?)
* or warn client and/or server side about the mismatch (using some class variable to store the result of the check).

My questions:

* Do you agree that the check should happen when initializing the app? Or would a memoized check on a reflex trigger be find as well?
* If you prefer check on initialization, would you raise and error and warn then, or upon execution of reflexes?
* If we warn when executing reflexes would you do it client or server side (I think client side, as this is usually where one is looking when troubleshooting)

references on how others do checks:

* `spring` does a check on `run`: https://github.com/rails/spring/blob/309ec0defa2b5c48d51cdb7d19251f6027a1b31b/lib/spring/client/run.rb#L116
[...]

PR #297, opened 2020-09-05
Avoid mismatching client and server versions

piotrwodz:
# Enhancement

Me and @rolandstuder paired up together on this enhancement.

## Description

This fix improves installation task of Stimulus Reflex.

Following the [setup](https://docs.stimulusreflex.com/setup#command-line-install):
```
bundle add stimulus_reflex
bundle exec rails stimulus_reflex:install
```
latest **stable** version of `stimulus_reflex` gem (`3.2`) gets added to `Gemfile` with bundler and the exact **latest** version of `stimulus_reflex` package (`3.3.0-pre3`) gets added to `package.json` with yarn. They do mismatch!

Fixes https://github.com/hopsoft/stimulus_reflex/issues/294

## Why should this be added

This PR avoids StimulusReflex client and server having mismatched versions. It is a good practice to keep the gem and the npm package versions in sync.

## Checklist

## Questions

1) Does `stimulus_reflex` [version](https://github.com/hopsoft/stimulus_reflex/blob/master/javascript/package.json#L3) in `package.json` get updated automatically when gem [version](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/stimulus_reflex/version.rb#L4) is increased?
2) It would be nice to run some check such as `verify_server_version` in order to display a warning about mismatched client and server versions and act accordingly. The question is: where should it happen? On [initialize](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/stimulus_reflex/reflex.rb#L52) / [process](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/stimulus_reflex/reflex.rb#L119) of `Reflex`?

### DIFF
--- a/lib/tasks/stimulus_reflex/install.rake
+++ b/lib/tasks/stimulus_reflex/install.rake
@@ -1,12 +1,14 @@
 # frozen_string_literal: true
 
 require "fileutils"
+require_relative "../../stimulus_reflex/version"
 
 namespace :stimulus_reflex do
   desc "Install StimulusReflex in this application"
   task install: :environment do
     system "bundle exec rails webpacker:install:stimulus"
-    system "yarn add stimulus_reflex"
+    gem_version = StimulusReflex::VERSION.gsub(".pre", "-pre")
+    system "yarn add stimulus_reflex@#{gem_version}"
 
     FileUtils.mkdir_p Rails.root.join("app/javascript/controllers"), verbose: true
     FileUtils.mkdir_p Rails.root.join("app/reflexes"), verbose: true


leastbad:
I **love** where you two are going with this PR.

1. If I understand the question, it is a manual process but we're pretty careful about keeping them synced.
2. I would start this journey in `lib/stimulus_reflex.rb`. After everything gets required, scan the project's `package.json` and find the line with `stimulus_reflex` on it. If the version is mismatched, I would use `puts` to output a giant, multi-line warning that maybe even builds an ASCII box alerting people and suggesting they update. This stops short of actually updating the `package.json`, which I just know would break something somewhere.

The issue that I don't currently have an easy fix for is: what if either the `Gemfile` or `package.json` don't have version numbers for `stimulus_reflex`? For example, if I have a local bundle, my `Gemfile` looks like:
```rb
gem "stimulus_reflex", github: "hopsoft/stimulus_reflex", branch: "master"
```
and because I use `yarn link` there literally is no `stimulus_reflex` line in my `package.json`.

RolandStuder:
We have two separate problems:

* `rails stimulus_reflex:install ` may install different versions
* updating `stimulus_reflex` or `yarn upgrage` might lead to differing versions, users do not get warned about not matching versions

#294 is conflating the two. We should 1) make sure on install we match versions 2) warn loudly if they don't match.

I think we should make two separate PRs to not conflate the two.  This PR solves number 1, and we could solve number 2 independently.

RolandStuder:
A warning, the current implementation will install a specific version of the npm package, so `yarn upgrade` will not automatically upgrade (which is what we want, as otherwise again it would automatically upgrade to prerelease versions, unless there is another way to go about this)

So the package json will have `"stimulus_reflex": "3.2.3",`, rather than `"stimulus_reflex": "^3.2.3",` (no caret)

piotrwodz:
> The issue that I don't currently have an easy fix for is: what if either the `Gemfile` or `package.json` don't have version numbers for `stimulus_reflex`? For example, if I have a local bundle, my `Gemfile` looks like:

@leastbad Then we would need to read the versions from `Gemfile.lock` and `yarn.lock` - search for the line with `stimulus_reflex` and fetch a version number.

leastbad:
That sounds... hard.

lib/tasks/stimulus_reflex/install.rake
```diff
@@ -1,12 +1,14 @@
 # frozen_string_literal: true
 
 require "fileutils"
+require_relative "../../stimulus_reflex/version"
```
RolandStuder, line 4: @piotrwodz I think you should change this to `require "stimulus_reflex/version"`, a relative path might mess things up. Not sure, but it's probably better that way.

leastbad: I'm going to do this locally before I cut the gem.

leastbad:
I'm going to roll with this! Congrats on your first PR, @piotrwodz 💯

PR #300, opened 2020-09-08
Enhance controller templates docs

### LINKED ISSUE #298: Update client controller template comments
# Feature Request

Both of the templates in `lib/generators/templates/app/javascript/controllers/` have some minor issues that could be easily addressed.

- on line 29 of the `application_controller.js.tt` it shows `ExampleReflex#demo` but this could be simplified to `Example#demo`
- same thing happens on line 20 of the generic template
- in the generic template, we should demonstrate that if they want to append and not just override methods present in controller, you have to call `super`. This is especially important for `connect`. It's a small bit of documentation that will save many headaches! So...
- add a `connect` example to the generic file eg.
```js
connect () {
  super.connect()
  // add your code here
}
```
- in the application controller template, re-word so that it's more clear and explicit: "these are event handlers for every reflex, for every controller that inherits from it"
- in the generic template, let's change reflex name from `update` to something more obviously specific like `dance` so that we can minimize the number of folks learning the framework who assume that these conceptually map to restful verbs
- let's point out that the controller has to be put on an elements with `data-reflex` for the callbacks to fire

There could be more! But these templates are an important part of our API and an early step in how people learn. If they are ambiguous, they will confuse people.

pinzonjulian:
# Enhancement

## Description

This PR enhances the comments included in the `application_controller.js.tt` and `%file_name%_controller.js.tt` that explain how stimulus controllers work in the context of StimulusReflex.

Fixes #298 

## Why should this be added

This PR solves a few doubts and gotchas I experienced myself as a first time user of SR. By enhancing these comments, I expect new developers using the gem to encounter less unexpected errors and be able to use the gem out of the box with less problems.

## Checklist

### DIFF
--- a/lib/generators/templates/app/javascript/controllers/%file_name%_controller.js.tt
+++ b/lib/generators/templates/app/javascript/controllers/%file_name%_controller.js.tt
@@ -22,12 +22,17 @@ export default class extends ApplicationController {
   // }
 
   /* Reflex specific lifecycle methods.
-   * Use methods similar to this example to handle lifecycle concerns for a specific Reflex method.
-   * Using the lifecycle is optional, so feel free to delete these stubs if you don't need them.
+   * For every method defined in your Reflex's ruby object, a set of lifecycle methods become available
+   * in this javascript controller. These are optional so feel free to delete these stubs if you don't
+   * need them.
+   *
+   * Important:
+   * Make sure to add data-controller="<%= class_name %>" to your markup alongside
+   * data-reflex="<%= class_name %>#example" for the lifecycle methods to fire properly.
    *
    * Example:
    *
-   *   <a href="#" data-reflex="<%= class_name %>Reflex#example">Example</a>
+   *   <a href="#" data-reflex="<%= class_name %>#example" data-controller="<%= class_name %>">Example</a>
    *
    * Arguments:
    *
--- a/lib/generators/templates/app/javascript/controllers/%file_name%_controller.js.tt
+++ b/lib/generators/templates/app/javascript/controllers/%file_name%_controller.js.tt
@@ -17,7 +17,7 @@ export default class extends ApplicationController {
    *   element - the element that triggered the reflex
    *             may be different than the Stimulus controller's this.element
    *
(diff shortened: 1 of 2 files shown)


pinzonjulian:
@leastbad I didn't quite understand what you meant by:
> in the application controller template, re-word so that it's more clear and explicit: "these are event handlers for every reflex, for every controller that inherits from it"

By event handlers do you mean the lifecycle methods/callbacks?

pinzonjulian:
Hey @leastbad 

I have a failing test I couldn't get how to fix.
https://github.com/hopsoft/stimulus_reflex/pull/300/checks?check_run_id=1083901859

It's comparing the output of the generated file to something but I'm not sure what's it comparing it to. 

These are the test lines responsible:
`assert_file "app/javascript/controllers/demo_controller.js", /DemoReflex/`
`assert_file "app/javascript/controllers/posts_controller.js", /PostsReflex/`

Could you guide through it? I haven't written these kinds of tests before.

marcoroth:
Hey @pinzonjulian! 

Thank you for the PR! This is looking good so far.

> By event handlers do you mean the lifecycle methods/callbacks?

Yes, he is talking about the four lifecycle methods in the ApplicationController template. (`beforeReflex`, `reflexSuccess`, `reflexError` and `afterReflex`).

See:
https://github.com/hopsoft/stimulus_reflex/blob/b8c5e973ea6b8186d76ec117dda96b0d9427a2d7/lib/generators/templates/app/javascript/controllers/application_controller.js.tt#L34-L44

Concerning the tests:

`assert_file` is checking if a given file is present and in this case it is also checking if the word `DemoReflex` is present in that file. Since you changed it to just the name without the `Reflex` you should be able to change the test to something like:

`assert_file "app/javascript/controllers/demo_controller.js", /Demo/`

Maybe it also make sense to check if the action is also present since `Demo` could also be pretty ambiguous.

I hope this helps! 😊

leastbad:
I believe `reflexHalted` is a thing, now, too!

leastbad:
We can add more to this, later! For now, these changes are great. Thanks @pinzonjulian!

PR #309, opened 2020-09-16
Print warning and exit if caching is disabled or npm/gem versions are mismatched

### LINKED ISSUE #258: Development caching is confusing
There was a [discussion](https://github.com/hopsoft/chatter/issues/2) on the Chatter repo about how things would act strangely if caching is not enabled in development. I believe that I have fully [cracked the code](https://github.com/hopsoft/chatter/issues/2#issuecomment-653757438) but I feel the scenario should be revisited.

Long story short: we don't (currently?) support cookie-based sessions. Instead, in the StimulusReflex install rake task we actually set the session storage to be cache_store and then enable caching in development as part of the task.

There are two very real issues with this:

First, some people might have legitimate reasons to not want caching enabled in development. Honestly, I don't understand the logic of developing with caching off, considering that if you're going to have bugs in production, development seems like a pretty good time to catch them. I personally believe that development caching should be **on by default** for this reason - but I don't get to make this call. If they need to have caching off and use StimulusReflex, they can use ActiveRecord session storage.

Second, the bigger problem is that a lot of problems emerge when people clone repos that use StimulusReflex and they don't know to explicitly turn caching on. It's done for you when you install SR, but you're on your own when you're cloning. **A significant and growing percentage of support concerns on Discord are resolved by getting developers to enable caching.**

I have several possible solutions to suggest:

1. We could [expand on our current opinionated defaults](https://github.com/hopsoft/chatter/blob/master/config/environments/development.rb) and instead of just adding `config.session_store :cache_store` to `config/environments/development.rb` on line 3 as we do now, we could actually **remove** lines 19 and 27-31 from development.rb, essentially removing the `rails cache:dev` locked in the *ON* position.
2. We could create `tmp/caching-dev.txt` if it doesn't already exist as part of the rake task.
3. We could install a git hook to create `tmp/caching-dev.txt` if it doesn't exist at checkout.
4. We could do:
```
config.cache_store = :memory_store
config.action_controller.perform_caching = false
```

None of these are perfect, but what's clear to me is that every outcome has compromises attached to it, including the status quo.

excid3:

Fix 

## Description

As stated in #258, caching being required trips up a lot of people. Typically, they'll see a seemingly random `InvalidAuthenticityToken` and no idea that it was related to the Stimulus Reflex requirements around caching. Usually they see this trying to login or sign up in their apps and find they get this error instead.

I've gotten a bunch of questions about this from Jumpstart Pro users who want to use Stimulus Reflex and run into this problem and it leads to a lot of confusion.

Fixes #258

This also addresses the other half of #297 and will verify the NPM package version matches the Rubygem version on boot. 👍 

## Why should this be added

This configures an initializer to verify that caching is enabled. If not, it will print a message and exit if caching is disabled. Same goes for mismatched client and server versions.

## Checklist

### DIFF
--- a/lib/stimulus_reflex.rb
+++ b/lib/stimulus_reflex.rb
@@ -20,12 +20,12 @@ require "generators/stimulus_reflex_generator"
 
 module StimulusReflex
   class Engine < Rails::Engine
-    initializer "verify_caching" do
+    initializer "stimulus_reflex.verify_caching_enabled" do
       unless caching_enabled?
         puts <<~WARN
           Stimulus Reflex requires caching to be enabled. Caching allows the session to be modified during ActionCable requests.
           To enable caching in development, run:
-          
+
             rails dev:cache
         WARN
         exit
--- a/lib/stimulus_reflex.rb
+++ b/lib/stimulus_reflex.rb
@@ -22,11 +22,11 @@ module StimulusReflex
   class Engine < Rails::Engine
     initializer "verify_caching" do
       unless caching_enabled?
-        puts <<-WARN
-Stimulus Reflex requires caching to be enabled. Caching allows the session to be modified during ActionCable requests.
-To enable caching in development, run:
-
-  rails dev:cache
+        puts <<~WARN
+          Stimulus Reflex requires caching to be enabled. Caching allows the session to be modified during ActionCable requests.
+          To enable caching in development, run:
+          
+            rails dev:cache
         WARN
         exit
       end
--- a/lib/stimulus_reflex.rb
+++ b/lib/stimulus_reflex.rb
@@ -20,6 +20,8 @@ require "generators/stimulus_reflex_generator"
 
 module StimulusReflex
   class Engine < Rails::Engine
+    NODE_VERSION_FORMAT = /(\d\.\d\.\d.*):/
+
     initializer "stimulus_reflex.verify_caching_enabled" do
       unless caching_enabled?
         puts <<~WARN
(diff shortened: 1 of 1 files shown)


RolandStuder:
I tried it on my machine and it works nicely!

A word of warning though: If you try it with `stimulus_reflex_harness` you might be confused, as the harness does enable caching by default and `rails dev:cache` won't turn it ON/OFF.

excid3:
This also addresses the other half of #297 and will verify the NPM package version matches the Rubygem version on boot. 👍

lib/stimulus_reflex.rb
```diff
@@ -20,5 +20,53 @@
...
+
+            yarn upgrade stimulus_reflex@#{gem_version}
+        WARN
+        exit
+      end
+    end
+
+    private
+
+    def caching_enabled?
```
julianrubisch, line 51: I noticed you took careful steps to align the different „pre“ version formats between node and ruby. Just checking: you did test this, correct?

RolandStuder: @julianrubisch I did some tests. It looks good, we also did the same substitution when adjusting the install command, see https://github.com/hopsoft/stimulus_reflex/pull/297/commits/53643b5af16e3c9aa674cd26e7177aeedf5fbb95#diff-91900d59e187ebdd5ac9ec1ada32fe44L9

excid3: Yep, it's on purpose to not confuse anyone (and it's the actual versions we compare, so a sanity check at the same time).

RolandStuder:
@excid3 the title and description should be updated, as this PR covers both #258 and #294.

lib/stimulus_reflex.rb
```diff
@@ -20,5 +20,53 @@
 
 module StimulusReflex
   class Engine < Rails::Engine
+    NODE_VERSION_FORMAT = /(\d\.\d\.\d.*):/
```
ParamagicDev, line 23: Should this be `/(\d{1,2}\.\d{1,2})\.\d{1,2}.*):/`

Or if you ever think you're going to go above 2 digits you could use `\d+`

Currently, `\d` will only match the first number, but will not match if there is a 2nd digit.

excid3: Good call, that should be `\d+` for each. 👍

PR #310, opened 2020-09-17
Support IE11 msCrypto (#308)

### LINKED ISSUE #308: IE11 Support: "crypto" is undefined
# Bug Report

Window.crypto is undefined in Internet Explorer 11.

```
ReferenceError: "crypto" is undefined [object Object]
```
## Describe the bug

https://github.com/hopsoft/stimulus_reflex/blob/691dca45eb6d119a467d3efeb4fac11e0e230321/javascript/utils.js#L7  

from:
https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto
```
var cryptoObj = window.crypto || window.msCrypto; // for IE 11
```
is the possible fix.

something like the following would work
image

open your app with IE11

### StimulusReflex

- Gem: 3.3.0.pre3
- Node package: 3.3.0-pre3

### Browser

- Browser Internet Explorer
- Version 11.1082.18362.0

chooselife22:
# IE11 Browser Support

## Description

To use the crypto object in Internet Explorer 11 we need to call it with a different name (msCrypto instead of crypto).
Described in: https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto

Fixes #308

## Why should this be added

IE11 Support

## Checklist

### DIFF
--- a/javascript/utils.js
+++ b/javascript/utils.js
@@ -1,6 +1,7 @@
 // uuid4 function taken from stackoverflow
 // https://stackoverflow.com/a/2117523/554903
 export const uuidv4 = () => {
+  const crypto = window.crypto || window.msCrypto
   return ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, c =>
     (
       c ^


leastbad:
Thanks, Andy!

PR #311, opened 2020-09-18
Refactor sanity checks on boot

excid3:

Bug fix

## Description

Cleaned up and refactored the sanity checker into a class that's run in an initializer and changed a few things:

* No longer exit, just print warnings. Exiting is too heavy handed.
* Check if yarn.lock and package.json exist before reading them.
* Move warning when NPM version can't be found.
* Memoize versions after lookup so we're not reading files multiple times.
* Fix version match to support multiple digits for each part (i.e. `3.10.11`)

## Why should this be added

Bug fixes

## Checklist


leastbad:
Great work on this. Thanks, Chris!

PR #322, opened 2020-10-07
Move StimulusReflex::Channel to app/ and allow for a configurable parent channel

dark-panda:

Feature/enhancement and bug fix

### Feature/Enhancement

Allows the parent class of `StimulusReflex::Channel` to be configurable via `StimulusReflex.parent_channel`. This can be done in an initializer...

```ruby
StimulusReflex.setup do |config|
  config.parent_channel = 'SomeChannelThatsNotStandard'
end
```

It defaults to `ApplicationCable::Channel` as it currently does.

### Bug Fix

This came about while attempting an upgrade because for some reason my app's own `ApplicationCable::Channel` implementation was not being loaded correctly, but only in certain situations. For instance, passenger in would not load the file, but puma would. There are mysterious forces at work in the Rails loading mechanisms and all of the various pieces that touch it, and they sometimes have different behaviour.

In putting together this PR, a second issue was discovered that is not specific to my app. Because SR is overriding `ApplicationCable::Channel#initialize` and then doing a `require` on the user-supplied implementation, the user-supplied implementation gets skipped on the first call to a channel's `#initialize` method. Thus, if the user supplies their own `ApplicationCable::Channel#initialize` method, it will only start to be used after at least one ActionCable connection is made, as that will cause the SR implementation to be called first, which itself then overwrites its own implementation via the `require`. Because this PR avoids overriding `ApplicationCable::Channel` in any way, we avoid this oddity. As to why anyone would override `ApplicationCable::Channel#initialize` is another matter, but regardless, this PR avoids this behaviour.

## Why should this be added

* It is possible that someone would want to provide a different parent class to `StimulusReflex::Channel`, or otherwise name their base channel something else entirely. This allows them to configure this behaviour, with the default being as it was before.

* It avoids a small issue discovered when calling a user-supplied `ApplicationCable::Channel#initialize`.

* Reduces code complexity by avoiding having to override `ApplicationCable::Channel` in SR.

## Checklist

### DIFF
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -1,16 +1,6 @@
 # frozen_string_literal: true
 
-module ApplicationCable
-  class Channel < ActionCable::Channel::Base
-    def initialize(connection, identifier, params = {})
-      super
-      application_channel = Rails.root.join("app", "channels", "application_cable", "channel.rb")
-      require application_channel if File.exist?(application_channel)
-    end
-  end
-end
-
-class StimulusReflex::Channel < ApplicationCable::Channel
+class StimulusReflex::Channel < StimulusReflex.parent_channel.constantize
   def stream_name
     ids = connection.identifiers.map { |identifier| send(identifier).try(:id) || send(identifier) }
     [
--- a/lib/stimulus_reflex.rb
+++ b/lib/stimulus_reflex.rb
@@ -11,7 +11,6 @@ require "cable_ready"
 require "stimulus_reflex/version"
 require "stimulus_reflex/reflex"
 require "stimulus_reflex/element"
-require "stimulus_reflex/channel"
 require "stimulus_reflex/sanity_checker"
 require "stimulus_reflex/broadcasters/broadcaster"
 require "stimulus_reflex/broadcasters/nothing_broadcaster"
@@ -20,6 +19,14 @@ require "stimulus_reflex/broadcasters/selector_broadcaster"
 require "generators/stimulus_reflex_generator"
 
 module StimulusReflex
+  # Parent channel class for StimulusReflex channels.
+  mattr_accessor :parent_channel
+  @@parent_channel = "ApplicationCable::Channel"
+
+  def self.setup
+    yield self
+  end
+
   class Engine < Rails::Engine
     initializer "stimulus_reflex.sanity_check" do
       SanityChecker.check!


RolandStuder:
Nice work!

As @leastbad mentions we have two PRs now that introduce configuration with initializers. Personally I prefer that we put the configuration into a configuration object. I feel like methods like `StimulusReflex.parent_channel` should give access the actual parent_channel, not the config for it, 
Here is how I did it: https://github.com/hopsoft/stimulus_reflex/pull/318/files#diff-c6c0f62499a9b75d52acab7ff5795f93R24

And a minor thing: Is the rails convention not `configure do` instead of `setup do` (for example see https://api.rubyonrails.org/classes/Rails/Application.html#method-i-config_for)

What do you people think?

leastbad:
`configure` is definitely the correct option. Thankfully easy to fix!

As for the configuration object pointing to the actual parent_channel... are you suggesting that instead of 
`config.parent_channel = "Foo::Channel"` 
it should be 
`config.parent_channel = Foo::Channel`

RolandStuder:
@leastbad I am suggesting `config.parent_channel = "Foo::Channel"`, so we could have `StimulusReflex.parent_channel` return `Foo::Channel`` if we wanted.

The current implementation on this PR returns the string `"Foo::Channel"` for  `StimulusReflex.parent_channel` , there is no `config` or `configuration` in this PR, there is just a `setup` method you can use set the class variable `@@parent_channel`.

ps: "Would `.config` be better than `.configuration` to access/alter the configuration? Rails is not consistent on this one: `Rails.configuration == Rails.application.config # returns true`

dark-panda:
I’d be fine with all that, I didn’t want to get into a bigger PR with a whole configuration mechanism as I figured that would require more discussion and I felt the bugs I discovered needed more immediate attention. The strategy I used here was obviously very simplistic, and was effectively taken directly from the Devise gem, which many Rails devs are likely familiar with. The config mechanism is kind of secondary to the purpose of this PR specifically though I suppose, so however that configuration gets set and retrieved is a somewhat separate but related matter. 

As for setting the value as a string specifically rather than a class constant, this is done due to the Rails initialization process sometimes acting a bit wonky when you’re autoloading constants, and sometimes loading constants in this part of the initialization process causes some weirdness when you’re in dev mode and constants need to be removed and re-added to the running dev app. If it were a constant, it may set up a hard reference and that causes havoc with that reloading mechanism. With a string, the constant can effectively be resolved just in time when needed and that can help alleviate the issue.

RolandStuder:
@dark-panda I was unclear: I like the config only containing a String. My suggestion is just to put all the configuration in its own place under `StimulusReflex.configuration`

So we can have
`StimulusReflex.configuration.exit_for_failed_sanity_checks  # -> true`
`StimulusReflex.configuration.exit_for_failed_sanity_checks # -> "Foo::Channel"

I put in a (untested) PR to add my config implementation to your thing (which is also pretty simple basically this: https://thoughtbot.com/blog/mygem-configure-block).

leastbad:
@dark-panda hey Jay! Great news... #339 got merged earlier today.

If you can rebase and merge with master, this should be approved very quickly.

leastbad:
Closed by #346. Thank you again, @dark-panda, and sorry for the PR schenanigans.

PR #323, opened 2020-10-08
Fix form autosave docs

julianrubisch:
# Documentation

## Description

Found out that the autosave form example will only work with `data-reflex-dataset="combined"`, thanks @RolandStuder 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -469,6 +469,10 @@
 
 [Full Changelog](https://github.com/hopsoft/stimulus_reflex/compare/v2.1.9...v2.2.0)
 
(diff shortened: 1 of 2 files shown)


leastbad:
Nice catch!

I'm pretty sure that the changelog is regenerated automatically, changing it directly seems like it won't stick?

QQ: does `data: {reflex_dataset: "combined"}` *actually* output `data-reflex-dataset`? If so, I had **no idea**. 😸

julianrubisch:
Oops, I have no idea how that changelog change slipped in 🤯

I copied this straight from my source, so I guess yes?!

PR #324, opened 2020-10-09
introduce tab isolation mode

leastbad:

Enhancement

## Description

Pass `isolate: true` to the SR initialization to put the library in tab isolation mode. SR-initiated CableReady events will only fire in the tab that created them.

`StimulusReflex.initialize(application, { isolate: true })`

## Why should this be added

Some developers have users with multiple tabs open at once. This will ensure that those developers have a mechanism to restrict Reflex execution to only the tab which initiated them.

## Checklist


hopsoft (approved):

Shigawire:
This is so good. Thank you so much for this feature!

Issue #321, opened 2020-10-07
Wrong input name parsing


strelok1111:
# Bug Report

When form input name contents integer value as key, StimulusReflex inserts nil in params as many times as the key number is. And if the number is big enough, the browser hangs.

## To Reproduce

```
<form>
  <input name="foo[100][bar]" />
  <a data-reflex="click->TestReflex#test" data-reflex-root='#bar'>click me</>
  <div id="bar"></div>
</form>
```
TestReflex
```
def test
  morph '#bar', params.inspect
end
```
When I click `<a>` element, I get this in `<div id="bar"></div>` element:

`<:parameters>"page", "action"=>"index", "foo"=>[nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, {"bar"=>"123"}]} permitted: 1>`

If I change field name to `foo[q_100][bar]` everything works as expected:

`<:parameters>"page", "action"=>"index", "foo"=>{"q_100"=>{"bar"=>"123"}}} permitted: 1>`

## Expected behavior

`<:parameters>"page", "action"=>"index", "foo"=>{"100"=>{"bar"=>"123"}}} permitted: 1>`

### StimulusReflex

- Gem: 3.3.0
- Node package: 3.3.0

### External tools

- Ruby: 2.6.3
- Rails: 5.2.4.3
- Node: 10.21.0

### Browser

- Browser chrome, firefox
- Version 80.0.3987.149, 81.0

marcoroth:
Hey @strelok1111! Thanks for reporting this bug. I was able to reproduce your issue in my app too. 

For me it breaks at `<input name="foo[10000][bar]" />`.

The issue is caused when we serialize the form data with the `serializeForm()` function here:

[...]

PR #325, opened 2020-10-09
Opt-in form serialization & params overriding

s-s:

Feature

## Description

First of all I would like to thank you for your great project, like a breath of fresh air! 

This PR introduces new option `skipFormSerialization: true|false` for `stimulate` method to disable form serialization into `params`. Also it changes the order final `params` are produced, by applying  `params` set in client `before` lifecycle callback over serialized form. 

Might be a sort of workaround for #321 by providing a way for custom form serialization.

## Why should this be added
As a developer I want to have a control over what is passed to server and have least surprises there. #208 had changed the way `params` are produced by giving priority for general case (form serialization) over specific case (passing custom params or their overwriting in client’s `before` callback). This steals control from developer. This PR tries to provide this control back.

Also some of my use-cases don’t require form serialization at all. I.e. I’m just call `stimulate` in my Stimulus controller with all args/parameters I need and produce some sprinkles server-side using `nothing` morphs and CableReady methods. IMO having an option to opt out of (quite large in my case) form serialization would be great in such cases.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -114,7 +114,7 @@ const extendStimulusController = controller => {
     //
     // - target - the reflex target (full name of the server side reflex) i.e. 'ReflexClassName#method'
     // - element - [optional] the element that triggered the reflex, defaults to this.element
-    // - options - [optional] an object that contains at least one of attrs, reflexId, selectors
+    // - options - [optional] an object that contains at least one of attrs, reflexId, selectors, resolveLate, skipFormSerialization
     // - *args - remaining arguments are forwarded to the server side reflex method
     //
     stimulate () {
@@ -137,7 +137,7 @@ const extendStimulusController = controller => {
         args[0] &&
         typeof args[0] == 'object' &&
         Object.keys(args[0]).filter(key =>
-          ['attrs', 'selectors', 'reflexId', 'resolveLate'].includes(key)
+          ['attrs', 'selectors', 'reflexId', 'resolveLate', 'skipFormSerialization'].includes(key)
         ).length
       ) {
         const opts = args.shift()
@@ -178,14 +178,15 @@ const extendStimulusController = controller => {
 
       setTimeout(() => {
         const { params } = element.reflexData || {}
+        const skipFormSerialization = options['skipFormSerialization'] || false
         element.reflexData = {
           ...data,
           params: {
-            ...params,
-            ...serializeForm(element.closest('form'), {
+            ...!skipFormSerialization && serializeForm(element.closest('form'), {
               hash: true,
               empty: true
-            })
+            }),
+            ...params
           }
         }
 
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -184,12 +184,13 @@ const extendStimulusController = controller => {
 
       setTimeout(() => {
         const { params } = element.reflexData || {}
-        const formData = options['serialize'] == false
-          ? {}
-          : serializeForm(element.closest('form'), {
-              hash: true,
-              empty: true
-            })
+        const formData =
+          options['serialize'] == false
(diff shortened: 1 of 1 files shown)


leastbad:
Workaround, sure... but it's not a *solution* to #321 in any way, but I appreciate that you're trying to make a strong case. 😃 

Help me understand... what is the current default preventing you from doing? Is this about wire size or functionality?

marcoroth:
I also thought about "opt-in form serialization" while I was thinking about #321 and asked myself why it is turned on by default.

If we would want to be consistent we would need a `data-reflex-dataset="form"` to also collect the form values. Similar as we do with the [data attribute inheritation](https://docs.stimulusreflex.com/reflexes#inheriting-data-attributes-from-parent-elements) with `data-reflex-dataset="combined"`.

s-s:
> Workaround, sure... but it's not a _solution_ to #321 in any way, but I appreciate that you're trying to make a strong case. 
You are right, it is not a solution, just a sort of _workaround_.
> Help me understand... what is the current default preventing you from doing? Is this about wire size or functionality?
Actually I have stucked with #321 as well by creating a form with nested attributes for `has_many` association. I'm using SR to populate this form with templates for new objects (several "Add new" links). I have experienced performance regression (visual lag) while calling reflex due to broken form serialization as described in #321. Opting out serialization has restored performance. Haven't profiled it yet...

leastbad:
@marcoroth I agree; I don't want to dodge responsibility but I wasn't directly involved in merging the `form-serialize` stuff originally, and if I could go back in time, I would make the functionality opt-in, for sure.

The problem is that now it is out in the wild, and we could change the default with a major SemVer bump, or we could do the slightly less pretty thing and adopt an override to turn it off, as Sergey proposes.

I honestly don't know what is the best course. Perhap a major version bump isn't so bad anyhow.

Question: if we added `data-reflex-dataset="form"` as you suggest, is there a scenario where people might want to use **both** `combined` and `form` at once? I suppose we could just split on a space...

marcoroth:
Of course a major version bump for something like this isn't nice. I don't know if it's really worth it.
I'm just saying from a consistence point of view it would make sense, but I don't have a problem when it's turned on by default and you have to opt-out.

> Question: if we added data-reflex-dataset="form" as you suggest, is there a scenario where people might want to use both combined and form at once? I suppose we could just split on a space...

I can imagine that there is a use case where you would want to have both. I haven't run into the need for that (yet) but I can see where that could be useful. On the other hand you could also just include the data-attribute as an input in your form:

```html
<div data-post-id="1">
  <form data-reflex="submit->Post#create" data-reflex-dataset="combined form">
    <input name="name">
  </form>
</div>

<!-- vs -->

<form data-reflex="submit->Post#create" data-reflex-dataset="form">
  <input name="id" type="hidden" value="1">
  <input name="name">
</form>
```

leastbad:
I'm not sure how I managed to piss off Prettier-Standard, but I just swapped `skipFormSerialization` (negative) for `serialize` (positive). I figure that by making it work as though it's an option that's true by default, we get around the awkwardness of making a negative the default.

Does that work for you?

leastbad:
@marcoroth here's what I think: for v3, let's go with the `serialize` (defaulting to `true`) approach.

When we do ultimately save up enough grievences for a v4, whether form serialization or isolation mode, we can consider flipping some defaults and then building out the `data-reflex-dataset="form"` as you suggested.

Does that seem like a reasonable compromise for v3?

marcoroth:
@leastbad Sure, I'm fine with this. This sound like a good and reasonable approach to me!

hopsoft (changes requested):
I like this but think we should change the config key to `serializeForm`.

s-s:
@hopsoft, done! :)

hopsoft (approved):

PR #315, opened 2020-09-23
Proposal: Reduce bundle size and add a bundler for Stimulus Reflex javascript

ParamagicDev:
# Type of PR (feature, enhancement, bug fix, etc.)

Proposal

## Description

Move `@rails/actioncable`, `stimulus`, and `cable_ready` to `peerDependencies`

Add `cable_ready` to the install task.

Add [microbundle](https://github.com/developit/microbundle) to bundle javascript for publishing.

## Why should this be added

By moving these `dependencies` to `peerDependencies`, it will reduce the size of StimulusReflex's package and remove any possible version issues between user defined packages and the packages SR ships.

Adding a Bundler helps SR run in different environments.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] Checks (StandardRB & Prettier-Standard) are passing

PR #327, opened 2020-10-10
Move form-data merge logic to the server-side

### LINKED ISSUE #321: Wrong input name parsing

marcoroth:

Bug Fix

## Description

This PR changes **where** the form data gets merged into the `params`.

The `<form>` still gets serialized in the frontend. Instead of merging it in the frontend the data gets passed separately via the websocket connection to the backend in the attribute `form_data`. After that the form-data gets merged into the params and is then available in a reflex as it was before with `params` (see #204).

This approach has the benefit, that the serialized data gets interpreted the same was as it would if the form was submitted via a HTTP post request. There seems to be no simple algorithm or library that is able to easily serialize the form to a matching "Rails-hash". This is why we convert the form-data on the server. And with that we can also get rid of the `form-serialize` dependency which isn't maintained anymore.

We are making use of the browser-native `FormData` object which seems to have the same browser-compatibility as websockets, so this should be fine: https://caniuse.com/mdn-api_formdata.

## Why should this be added

This PR fixes #321 and also prepares for a future where we would want to serialize form data from a shadow-dom (like shoelaces). The only thing we need to pass to the backend via the websocket connection is a url query string (urlencoded or not), like:

```
person[address][23][city]=Paris&person[address][45][city]=London
or
person%5Baddress%5D%5B23%5D%5Bcity%5D%3DParis%26person%5Baddress%5D%5B45%5D%5Bcity%5D%3DLondon
``` 

The backend-part converts both of the above to:

```ruby
{'person' => {'name' => 'Bob', 'address' => {'23' => {'city' => 'Paris'}, '45' => {'city' => 'London'}}}}
``` 

Perviously the same form data was available on the backend like this:

```ruby
{'person' => {'name' => 'Bob', 'address' => [nil, nil, nil, [...], {'city' => 'Paris'}, [...], nil, nil, nil, [...], {'city' => 'London'} ] }}
```

/cc @strelok1111

## Checklist


leastbad:
Interesting!

Just so I'm following along (I know it's still a draft) couldn't we swap out `form-serialize` for the native `formData` API even if we don't move serialization to the server? Or am I missing the whole point?

marcoroth:
With the native `formData` API we are only able to get a query string out of the `<form>` but not a nested object like we want to have. 
To get the query string we could use the `form-serialize` library, but since it's not maintained anymore and we can achieve the same thing with the native API I figured we can swap that out.

leastbad:
This might be a master of the obvious question(s) but:

- does it matter if the library isn't maintained anymore? the basic concept of the library is not going to be changing, or am I wrong about that?
- what are people using instead of `form-serialize`? or is that a 🦗 🦗 🦗 ?

marcoroth:
* In my opinion it doesn't matter if the library is maintained anymore or not - the concept will stay the same. But we would either need to fork the package or somehow get a patch in there to make it behave like we want/need it. That's why I thought it's easier/more logical to use the same method that Rails uses by moving it to the server-side so we can match the http behavior. 

* To be honest I don't really know if the others really need to serialize form-data like we have to, since there are a plenty of different form-libraries which are going to abstract the standard html form in a way or another and miss-use to original intend of the html form 😅. 
It feels like we have a very edge-case Rails-specific problem here, that no library will solve. But if there is one I'm not aware of I'd happy to use it. 
The other option would be that we try to write an algorithm which builds the right looking hash on the client-side. This just didn't seemed like the way to go, but I'm open for that. 

I can't see any point to would speak for/against "do it on the client" and I can't see a point that we speak for/against "do it on the server". The only point is that the Ruby method already exists and matches how we want it to be. I hope that makes sense in a way 🙃

leastbad:
That's a great explanation. It hadn't occurred to me that the serialization was a Rails-y near-spec, but of course that's self-evident if you think for a moment. 🥇 

Nate has historically been against forking morphdom, which I pushed hard for but ultimately I'm very glad he talked sense into me.

That said, it could be that a fork of a library like `form-serialize` that is so purpose-built might be okay. It's worth having the conversation when he swings by.

lib/stimulus_reflex/channel.rb
```diff
@@ -35,7 +35,9 @@ def receive(data)
     arguments = (data["args"] || []).map { |arg| object_with_indifferent_access arg }
     element = StimulusReflex::Element.new(data)
     permanent_attribute_name = data["permanent_attribute_name"]
+    form_data = Rack::Utils.parse_nested_query(data["form_data"] || "")
     params = data["params"] || {}
+    params.deep_merge!(form_data)
```
s-s, line 40: I would suggest changing the order of merging to allow params overwriting:
```suggestion
    params = Rack::Utils.parse_nested_query(data["form_data"] || "")
    params.deep_merge!(data["params"] || {})
```

Here is focused example where param `foo` is explicitly set in stimulus controller:

view:
```html
<form>
  <input name="foo" type="text" value="chicken">
  <div id="result"></div>
  <a data-action="chicken-egg#poke" data-controller="chicken-egg" href="#">Poke!</a>
</form>
```
stimulus controller:
```javascript

  connect() {
    StimulusReflex.register(this)
  }

  beforeReflex(element) {
    const { params } = element.reflexData
    element.reflexData.params = { ...params, foo: 'egg' }
  }

  poke(event) {
    event.preventDefault()
    this.stimulate('ChickenEgg#poke')
  }
}
```
reflex:
```ruby
class ChickenEggReflex < StimulusReflex::Reflex
  def poke
    morph '#result', params[:foo]
  end
end
```

s-s:
I can confirm this PR fixes #321, have checked it on a real scenario with nested form.

lib/stimulus_reflex/channel.rb
marcoroth: This makes sense - as you already described this in #325. 
I changed it accordingly.

s-s: Great! Thank you!

RolandStuder:
I like this approach! A dependency less and it will rely on the same mechanism as Rails does, so no surprises with differences in how a form gets serialized.

Interesting note: Rails on remote forms adds `button-name: button-value`  to the form params, also they provide a method to serialize any element even if it is not a form, see: https://github.com/rails/rails/blob/c9e52d028c0347e377177a271486edfb3dac3975/actionview/app/assets/javascripts/rails-ujs/features/remote.coffee#L36

marcoroth:
@RolandStuder good call! I adjusted it. Now it also includes the button which submitted the form.

From my side this PR is ready as is.

leastbad:
Sorry for the commit spam. Some day I will understand git.

I wanted to try and help by resolving the inevitable merge conflict with #325. I also changed the JS-side `form_data` to `formData`.

If everything still looks good to you, Nate will probably have a chance to check this out today.

marcoroth:
@leastbad looks good to me! Didn't know that you are able to push to the branch in my fork 🙃 

I just updated the content of `package.json` with the content of `javascript/package.json` to reflect the new changes - most notably #315.

leastbad:
> @leastbad looks good to me! Didn't know that you are able to push to the branch in my fork 🙃

I thought that was mighty fishy, TBH. Is there any chance that there's a serious security hole hiding in plain sight on GH and it took me trying to do something obnoxious to notice?

marcoroth:
I just realized that this is what the "Allow edits by maintainers" options means. Pretty cool!

See: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

leastbad:
Hey @marcoroth, I just noticed that your new `serializeForm` function returns an array, which could be empty.

However, [this line](https://github.com/hopsoft/stimulus_reflex/blob/376a05107f26475920b92bbc7d57400d0c352577/javascript/stimulus_reflex.js#L186) suggests that the current implementation will send `{}` if the option if `options['serializeForm']` is false.

Finally, [this line](https://github.com/hopsoft/stimulus_reflex/blob/376a05107f26475920b92bbc7d57400d0c352577/lib/stimulus_reflex/channel.rb#L39) suggests that it will default to an empty string.

Clearly they can't all be right! 😁 

Can you please make sure that all three places have the correct defaults? Note that it seems likely the `channel.rb` instance doesn't need an empty fallback since presumably the client is going to send something, regardless.

leastbad:
Much better. 👍

marcoroth:
Hey @leastbad, you are right! 

> However, [this line](https://github.com/hopsoft/stimulus_reflex/blob/376a05107f26475920b92bbc7d57400d0c352577/javascript/stimulus_reflex.js#L186) suggests that the current implementation will send `{}` if the option if `options['serializeForm']` is `false`.

This true, I changed it. Now it defaults to`''` as it should. I guess that's a left-over from the merge with #325.

> Note that it seems likely the channel.rb instance doesn't need an empty fallback since presumably the client is going to send something, regardless.

That's right. I wanted to be extra safe here, but there is no need for it:
```ruby
irb(main):011:0> Rack::Utils.parse_nested_query(nil)
=> {}
irb(main):012:0> Rack::Utils.parse_nested_query("")
=> {}
```

Quick note: The new `serializeForm` function does **not** return an Array. It returns an url query string, like: 
```
person[address][23][city]=Paris&person[address][45][city]=London
```

PR #330, opened 2020-10-16
add signed/unsigned accessors to element

joshleblanc:
# Feature

## Description

This is an expiremental idea that we're hashing out in discord. Provide accessors on `element` that allow implicitly locating signed and unsigned global IDs.

With this change, you can use `element.signed[:foo]` or `element.unsigned[:foo]` to map the dataset attribute `foo` to a model.

```ruby
class MyReflex < ApplicationReflex
  def add_comment
    comment = element.signed[:comment]
    comment.update(content: params[:content] if comment
  end
end
```

## Why should this be added

Developer happiness 

## Checklist


hopsoft (approved):
This is really cool. Very nice work!

PR #331, opened 2020-10-16
Add enable/disableDebug function and expose them on a global object

julianrubisch:
# Enhancement

## Description

Set a `sessionStorage` key for debugging and use that. As it stands, https://github.com/hopsoft/stimulus_reflex/compare/master...julianrubisch:enable-disable-debugging?expand=1#diff-ceab174211ef2723122defb53635f4dc25e684f9c68a3d1fd3cda0046214892fR396 will actually wipe out the setting on reload, but I didn't want to break the existing API.

## Why should this be added

This idea came when playing around with LiveView. I realized they expose convenient `enable/disableDebug` functions globally.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -34,8 +34,14 @@ let actionCableSubscriptionActive = false
 // A dictionary of all active Reflex operations, indexed by reflexId
 window.reflexes = {}
 
-// Indicates if we should log calls to stimulate, etc...
-let debugging = false
+// Debug state convenience functions
+const getDebug = () => {
+  return sessionStorage.getItem('stimulus-reflex-debugging') === 'true' || false
+}
+
+const setDebug = debug => {
+  sessionStorage.setItem('stimulus-reflex-debugging', debug)
+}
 
 // Subscribes a StimulusReflex controller to an ActionCable channel.
 // controller - the StimulusReflex controller to subscribe
@@ -82,7 +88,7 @@ const createSubscription = controller => {
       rejected: () => {
         actionCableSubscriptionActive = false
         emitEvent('stimulus-reflex:rejected')
-        if (debugging) console.warn('Channel subscription was rejected.')
+        if (getDebug()) console.warn('Channel subscription was rejected.')
       },
       disconnected: willAttemptReconnect => {
         actionCableSubscriptionActive = false
@@ -190,7 +196,7 @@ const extendStimulusController = controller => {
 
       reflexes[reflexId] = { finalStage: 'finalize' }
 
-      if (debugging) {
+      if (getDebug()) {
         Log.request(
           reflexId,
           target,
@@ -210,7 +216,7 @@ const extendStimulusController = controller => {
 
       promise.reflexId = reflexId
 
-      if (debugging) promise.catch(NOOP)
+      if (getDebug()) promise.catch(NOOP)
       return promise
     },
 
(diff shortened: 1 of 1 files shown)


javascript/stimulus_reflex.js
```diff
@@ -397,7 +403,7 @@ const initialize = (application, initializeOptions = {}) => {
     'stimulus-reflex',
     controller || StimulusReflexController
   )
-  debugging = !!debug
+  setDebug(!!debug)
```
fractaledmind, line 406: Geniunely just curious about your thinking and preferences here. My default thought would be to coerce the value in the `setDebug` method, instead of at the various calling sites.

julianrubisch: Coincidence... didn’t pay enough attention when refactoring 😉 thanks

javascript/stimulus_reflex.js
```diff
@@ -34,8 +34,14 @@ let actionCableSubscriptionActive = false
 // A dictionary of all active Reflex operations, indexed by reflexId
 window.reflexes = {}
 
-// Indicates if we should log calls to stimulate, etc...
-let debugging
+// Debug state convenience functions
+const getDebug = () => {
+  return sessionStorage.getItem('stimulus-reflex-debugging') === 'true' || false
```
hopsoft, line 39: Do we need the explicit `|| false` here?

hopsoft (changes requested):
I like where you're going with this but I think I'd rather see an abstraction on top of session storage. Something like this perhaps.

```js
const Session = {
  set debugging (value) {
    return sessionStorage.setItem('stimulus-reflex-debugging', !!value)
  },
  get debugging () {
    return sessionStorage.getItem('stimulus-reflex-debugging') === 'true'
  }
}
```

We could save this in its own file and import where needed. I expect we'll find more use cases for it down the road.

javascript/stimulus_reflex.js
julianrubisch: Probably not. Being overly cautious again 😁

leastbad:
I'm having one of those moments. Could you explain to me the why of this PR like I'm 🤤 

Why do we need to store the debugging flag in sessionStorage? What's the advantage of having it be in global scope?

Note: not a challenge in any way, I just actually don't (yet) get it. My experience has been really binary: either I am in development and want debugging, or I'm in production and do not. What am I currently missing?

julianrubisch:
SessionStorage survives page reloads, while a global object does not. I just found myself flipping the debugging flag in the initialization method manually a couple of times, because console got too noisy. 

Figured I’d like to stay in the console to flip it on/off

leastbad:
Okay, thanks for the explanation!

julianrubisch:
True, to actually make use of sessionStorage we’d have to drop the initializer option. That would mean a (minor/major?) version bump. Still, as @hopsoft noted, an abstraction would probably be useful still. I can see other use cases for a sessionStorage in the future...

leastbad:
I was 100% fine with the option existing; just because I don't plan to use it doesn't mean it's not useful for folks with other usage patterns (and I don't have to have an opinion on everything).

However, I do have a rather strong negative *reaction* to changing the current behavior, which is serving quite well. After all, you could make a strong argument that you might not want people randomly turning debug mode on in production.

Given that you can easily tell your inspector to (temporarily) filter messages from a script by right-clicking on them, I worry that we might be getting ahead of ourselves on this.

julianrubisch:
True. It’s just a suggestion, after all.

PR #334, opened 2020-10-20
Delegate flash to the request

hopsoft:
# Enhancement

## Description

Expose the Rails flash directly to the reflex by delegating to the request.

## Why should this be added

Makes using the flash more intuitive and cleans up a little verbosity in reflexes.

## Checklist


leastbad:
I'm surprised that 1. it was that easy and 2. this didn't come up AGES ago!

PR #337, opened 2020-10-25
Server-side logging per reflex action

### LINKED ISSUE #264: Improve server-side logging options
# Feature Request

It would be nice if StimulusReflex offered better logging tooling that was comfortably in-between "way too verbose" and "totally suppressed".

@marcoroth recently added amazing client-side logging, but we offer no SR-specific feedback unless there's literally an excpetion. This leads to lots of `puts` statements in Reflexes while debugging. We can do better!

The ideal solution emits **one tastefully colored line per Reflex action**.

Some possible columns include:

ReflexClass#action eg. Example#hello
Timestamp: eg. Time.now
Round-trip execution time eg 8ms
Session Identifier: eg. [215a1c85]
Connection Identifier from AC channel: eg. leastbad
Mode: eg. :selector
Selector: eg. #notification
Operation: eg. (selector: #notification)

## Dream output

`[215a1c85] Example#hello in 8ms (selector: #notification) for "leastbad" at 2020-07-07 15:04:13`

... but in color.

### Bonus points for being configurable via an initializer:

```rb
# config/initializers/stimulus_reflex.rb
  # config.logging = :nil
  config.logging = [:session, :reflex, :benchmark, :operation, :identifier, :timestamp]
end
```

## Additional context

The current options are two extremes: https://docs.stimulusreflex.com/troubleshooting#server-side

Adding an initializer is long overdue! Here's how:

```rb
  class << self
    mattr_accessor :logging
    self.logging = [] # this is the default, even if there is no initializer
  end

  def self.configure(&block)
    yield self
  end
```

## Colors!

puts "\e[32mThis is green!"

puts "\e[1;32mThis is bright green!"

9UVnC

piotrwodz:

Feature

## Description

Server-side logging per reflex action.

This PR adds:
- logger class responsible for printing a colorized log per reflex action (green - success / yellow - halted / red - error)
- stimulus reflex config initializer template
- generator for that template :point_up (usage: `rails generate stimulus_reflex::config`)
- config class handling specific user's configuration found in `config/initializer/stimulus_reflex.rb` (this class is initializing `debugging` and `logging` and defining accessors for them)
- logger attribute to reflex class
- finally invoking of logger after commiting session

Log message does not include a round-trip duration. Indeed, I am not sure how to know the exact exuction time.  Any suggestions would be appreciated. Other than that, I covered all columns mentioned in the description of the issue.

Fixes #264

## Why should this be added

Currently there is no server-side logging per reflex action. It would be nice to have one 👍 

It's also a good compliment to shutting off the highly verbose ActionCable logs. This gives you the information you need, not a core memory dump.

## Usage

This PR will install a default logging format that looks like this:

WindowsTerminal_8jogYfMI55

The developer can modify this Reflex log message format to their own taste. The message format can be redefined in the SR initializer by returning a String from a passed proc.

Available tokens:
`session_id`, `session_id_full`, `reflex_info`, `operation`, `reflex_id`, `reflex_id_full`, `mode`, `selector`, `operation_counter`, `connection_id`, `connection_id_full`, `timestamp`

Available colors:
`red`, `green`, `yellow`, `blue`, `magenta`, `cyan`, `white`.

Some examples of how a developer can define colorized Reflex log formats:

```ruby
config.logging = proc { "[#{session_id}] #{operation_counter.magenta} #{reflex_info.green} -> #{selector.cyan} via #{mode} Morph (#{operation.yellow})" }
```

```ruby
config.logging = proc { "[#{session_id}] #{operation_counter.magenta} #{reflex_info.green} -> #{selector.cyan} via #{mode} Morph" }
```

```ruby
config.logging = proc { "[#{session_id_full}] Reflex with id #{reflex_id.blue} #{reflex_info.green} via (#{mode.yellow}) for #{connection.cyan} at #{timestamp.magenta}" }
```

Advanced usage allows specifying ActiveRecord attributes for objects in the list of ActionCable Connection identifiers. For example, let's say your Connection class has `identified_by :current_user` and this points to your applications `User` model. If your user is logged in and the User model has an `email` attribute, you could specify a logging format such as:

```ruby
config.logging = proc { "#{reflex_info.green} -> #{selector.cyan} for #{email.red}" }
```

This will display the logged-in user's email address in red. If there is no user logged in, the logger will output `-`.

## Checklist


marcoroth (commented):
Thanks @piotrwodz for your contribution. I just left some comments.

The other thing I noticed: ActionCable logs two pretty long messages in between the server-side log entry you added. This makes it really hard to spot the reflex action log.

```
StimulusReflex::Channel#receive({"target"=>"FolderReflex#edit", "args"=>[], "url"=>"http://localhost:3000/", "attrs"=>{"class"=>"fa fa-pen btn btn-light", "data-reflex"=>"click->FolderReflex#edit", "data-reflex-dataset"=>"combined", "data-action"=>"click->folders#__perform", "checked"=>false, "selected"=>false, "tag_name"=>"I"}, "dataset"=>{"data-controller"=>"folders", "data-folder-id"=>"64", "data-reflex-root"=>"#folder_64", "data-reflex"=>"click->FolderReflex#edit", "data-reflex-dataset"=>"combined", "data-action"=>"click->folders#__perform"}, "selectors"=>["#folder_64"], "reflexId"=>"4425aa17-b097-4c22-b3a1-9f323e06beb4", "permanent_attribute_name"=>"data-reflex-permanent", "params"=>{}})

Folder Load (0.2ms)  SELECT "folders".* FROM "folders" WHERE "folders"."id" = $1 ORDER BY "folders"."name" ASC LIMIT $2  [["id", 64], ["LIMIT", 1]]
  ↳ app/reflexes/folder_reflex.rb:20:in `edit'
  Rendered application/_actions.html.erb (Duration: 0.0ms | Allocations: 9)
  Rendered application/_layout_switch.html.erb (Duration: 0.0ms | Allocations: 8)
  Rendered folders/_parent_folder.html.erb (Duration: 0.0ms | Allocations: 5)
  Rendered collection of folders/_item.html.erb [0 times] (Duration: 0.0ms | Allocations: 3)
  Rendered collection of documents/_item.html.erb [0 times] (Duration: 0.0ms | Allocations: 3)
  Rendered folders/_empty.html.erb (Duration: 0.0ms | Allocations: 5)
  Rendered folders/_folder.html.erb (Duration: 2.3ms | Allocations: 1353)
  Rendered application/_actions.html.erb (Duration: 0.0ms | Allocations: 9)


[ActionCable] Broadcasting to StimulusReflex::Channel:b7fe26c2a9926a4ae566789057e761c3: {"cableReady"=>true, "operations"=>{"morph"=>[{"selector"=>"#folder", "html"=>"[... a lot of html in here]", [...], "morph"=>:page}}]}}

StimulusReflex::Logger logging success for reflex action:
[b7fe26c2a9926a4ae566789057e761c3] LayoutReflex#set_layout (page: ##folder) for b7fe26c2a9926a4ae566789057e761c3 at 2020-10-26 00:08:57 +0100

StimulusReflex::Channel transmitting {"cableReady"=>true, "operations"=>{"morph"=>[{"selector"=>"#layout-switch", "html"=>"\n  <div class=\"btn-group\" role=\"group\" aria-label=\"Basic example\">\n    <button type=\"button\" data-reflex=\"click-&gt;LayoutReflex#set_layout\" data-layout=\"list\" id=\"layout-list\" class=\"btn btn-ou... (via streamed from StimulusReflex::Channel:b7fe26c2a9926a4ae566789057e761c3)

```

Do we have an option to manipulate that output and/or the order of them?

leastbad:
Regarding the issue of initializers, I've asked @RolandStuder if he'd be willing to address this: https://discordapp.com/channels/629472241427415060/733725826411135107/770050840111087659

lib/generators/config_generator.rb
```diff
@@ -0,0 +1,16 @@
...
+require "rails/generators"
+
+module StimulusReflex
+  module Generators
+    class ConfigGenerator < Rails::Generators::Base
+      source_root File.expand_path("templates", __dir__)
+
+      # Generates default StimulusReflex configuration file into your application config/initializers directory.
+      def copy_config_file
+        copy_file "stimulus_reflex_config.rb", "config/initializers/stimulus_reflex.rb"
```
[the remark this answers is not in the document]
piotrwodz: Thanks for that. Since config and its generator were extracted from here to #339 it could be handled there I believe.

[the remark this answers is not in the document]
marcoroth: Yes. @RolandStuder already took care of that as it seems 👍

piotrwodz:
> Do we have an option to manipulate that output and/or the order of them?

@marcoroth The output of `[ActionCable] Broadcasting` can be manipulated. I am not sure about the output of `StimulusReflex::Channel transmitting` though.

Re changing the order: it just does not seem right to me. There are several operations happening on a `reflex` before its action is being logged. I would prefer not to change their order. In fact, some of these operations build reflex action log.

Personally, what I find verbose is the output of `StimulusReflex::Channel#receive` but I do not think it is a scope of this PR.

piotrwodz:
@marcoroth On a second thought, all these long messages appearing around my reflex action log are verbose ☝️ Their output gets duplicated.

@leastbad What is your opinion on that? I would prefer improving these ActionCable messages in a separate PR.

leastbad:
Hey @piotrwodz, I'm a bit fuzzy on the context in which you're referring to changing the order.

You could be talking about:
- which debug elements are displayed, in which order
- when the debug message is emitted, in relationship to the ActionCable message
- at what stage of the server-side Reflex processing do you want to emit the message

Help me drill down and I'll give you a better answer. :)

piotrwodz:
@leastbad This is the [context](https://github.com/hopsoft/stimulus_reflex/pull/337#pullrequestreview-516399777) I was referring to. In particular whether it is possible to manipulate the output and/or the order of ActionCable messages.

leastbad:
Hi @piotrwodz, I appreciate you allowing me the opportunity to express myself more clearly.

While I might not always have time/capacity to act (or even respond) to every conversation, the one thing I *do* do here is read everything that transpires, as carefully as I am able. The intention of my question, then, is that I'm feeling a bit lost when the topic of "order" comes up. Thanks to the ambiguity of technical language, the intentions expressed in this thread simply aren't as nakedly obvious as you believe that they are... hence me asking for clarification and why I suggested three possible interpretations. Unfortunately, pointing me to a message I'd already read doesn't fix this for me; it's like speaking slowly and loudly to someone who doesn't speak the same language. 😁 

When I wrote the original issue, I had a vision for a solution where developers could craft their own debug message based on the information that they wanted to see. Front and center in this hope was that you could change the order in which the information was displayed by changing the the order in which the keys were specified. In the example cited:

`[215a1c85] Example#hello in 8ms (selector: #notification) for "leastbad" at 2020-07-07 15:04:13`

might be generated by:

`config.logging = [:session, :reflex, :benchmark, :operation, :identifier, :timestamp]`

while

`2020-07-07 15:04:13 Example#hello for "leastbad"`

would be generated by:

`config.logging = [:timestamp, :reflex, :identifier]`

Now, the subtle difference here is that in this example, there's no "at " before the timestamp... so that means we can adapt our little DSL to understand strings as well as symbols, perhaps:

`config.logging = [:session, :reflex, :benchmark, :operation, :identifier, "at ", :timestamp]`

or perhaps it makes more sense to express this as a string? I honestly don't know, but I was excited to have the conversation:

`config.logging = ->(context) { "#{context.session} for #{context.identifier}" }`

Awesome! Now, I also brought up the notion of using color. I see that you've run with setting a color for the whole line, which is a good start, but the main reason I'm excited for color is to break up the individual elements of a log line, making it far easier to parse in a log analyzer.

I can imagine defining color using both the array and lambda approaches:

`config.logging = ->(context) { "#{context.session} for \e[1;32m#{context.identifier}" }`

`config.logging = [:timestamp, :reflex, "\e[1;32m", :identifier]`

As for the relationship to the standard Rails ActionCable broadcast log line... it's [currently] my view that once you have AC/SR/CR setup and working properly in your application, you really don't need the noise of seeing a huge multi-line blob of text showing all of the attributes coming in OR the huge multi-line blob of text showing the CableReady operation hash data being sent back to the client. It was my hope that the single line debug we're discussing in this PR would become the preferred log output that you run with in production, and in development when nothing is going horribly wrong. That if you were actively debugging a problem with data not arriving as expected, you'd turn on the full debug and go over every detail.

So, with the above all said, the order in which this debug line appears in relationship to the standard ActionCable log blobs seems somewhat moot because they are really designed for different phases of the application development lifecycle.

Does that make sense?

Curious for your reaction to any/all of this.

RolandStuder:
@piotrwodz #339 is merged, so you can now use the `StimulusReflex.config`, maybe you want rebase or merge and handle the conflicts, so @leastbad does not have to do it.

leastbad:
💆‍♂️

piotrwodz:
Hey @leastbad 👋 Thank you for your feedback. I find it valuable and I can see clearly where you are going with it. Let me attack mentioned aspects one by one:
1) Crafting debug message:
Seems reasonable to me. Your examples provide more flexibility of a printed output. Indeed, developers might have different needs when it comes to _what_ and _how_ they want to see that output 👍
I prefer to aim at:
`config.logging = %w[reflex session for identifier operation at timestamp]`
This is a mix of reflex-related strings (e.g. session, operation etc.) and custom strings in between them that can formulate a sentence together.
2) Colors:
Indeed, I took a different approach here. My intention was to emphasize whether a reflex was successful, halted or with some errors. Hence, three colors used.
Your suggestion of having a different color per log element (e.g. green for reflex, yellow for operation etc.) is achievable 👍 I like that array-based approach.
3) Relationship to the standard Rails ActionCable broadcast log line:
Now you are talking! Muting standard verbose messaging completely and printing only the reflex log is an improvement. It can be controlled through debug configurable variable. I would like to give it a go 👍

leastbad:
Awesome! I love it when a plan comes together.

Pretty much the only thing I'm a bit fuzzy on is your use of `%w`... what if I want to include the word "reflex" in my output? What if I want to insert non-alphanumeric characters like `-` or `[`braces`]`? I figured that an array of symbols and strings would immediately remove any ambiguity.

Can you think of a different approach that is similarly flexible?

leastbad:
Hey @piotrwodz, I had originally hoped to get this into v3.4.0 but from my perspective, it makes more sense to spend some time on this over the next weeks and get it perfect. Do you agree?

piotrwodz:
Hi @leastbad Sure thing! I am working towards the completion as we outlined above. It should be released once we are happy with the final implementation 👍

piotrwodz:
@leastbad @marcoroth I addressed your remarks 👍
More control over customizing reflex log message (using brackets, spaces, words etc.) ✅
Colorizing log elements depending on developer's taste ✅
Opting out of ActionCable verbose log messages on `debug: false` ✅

app/channels/stimulus_reflex/channel.rb
```diff
@@ -113,4 +117,10 @@ def fix_environment!
       controller.renderer.instance_variable_set(:@env, connection.env.merge(controller.renderer.instance_variable_get(:@env)))
     end
   end
+
+  def disable_full_debug
+    if StimulusReflex.config.debug == false && StimulusReflex.config.parent_channel == "ApplicationCable::Channel"
+      ActionCable.server.config.logger = Logger.new(nil)
```
piotrwodz, line 123: This line stops `ActionCable` from producing any verbose log messages. Note: this only works if a parent channel class is `ApplicationCable::Channel` of course.

leastbad: Actually, why is this important? I honestly don't follow. :)

piotrwodz: Well, since https://github.com/hopsoft/stimulus_reflex/pull/346 was merged we gained a possibility of using other channel class than `ApplicationCable::Channel`, didn't we? This if statement opts out of printing AC log messages only. Can we be sure that provided custom class is built on top of AC? If so, then we do not this check: `StimulusReflex.config.parent_channel == "ApplicationCable::Channel"`.

leastbad: At this juncture - in advance of potential future modular transports - SR cannot work if channel doesn't ultimately inherit from ActionCable. I think we should drop that guard.

I also admit that I didn't initially intend for SR to explicitly shut off the built-in ActionCable logging, just that it would be encouraged/possible alongside the functionality you're building.

See, the thing is that turning off AC logging also happens to make the whole system work faster, too! What's not to love, right?

Well, the problem is that the opposite of `StimulusReflex.config.debug == false` shouldn't be turning on AC debugging. Instead, I think we should just make strong recommendations to people that they should enable SR debugging and disable AC debugging, and show them how (as we do already, [here](https://docs.stimulusreflex.com/troubleshooting#server-side)).

leastbad:
Hey, what's the practical difference between `puts` and `Kernel.print`?

app/channels/stimulus_reflex/channel.rb
leastbad: So, here's an interesting development. According to my testing, setting `ActionCable.server.config.logger = Logger.new(nil)` inside the channel doesn't have any impact until the 2nd connection is established... which makes sense because the channel is already live.

Anyhow, having sat on this for some hours, I think I'm going to pull out the `config.debug` stuff. The correct thing to do is provide good instructions, not impose our logic on how the developer wants to configure their Rails app. They could want both!

leastbad:
Alright, now I think we're cooking with gas! :fire:

I removed the ActionCable logger stuff as I said I would, and then kept going... I've added a fair number of additional symbols which can be passed, including the colors themselves. One thing I am quite proud of is that if you pass a symbol, it will attempt to access that attribute on your connection's identifiers. For example, I use Devise and my User table has a `login` attribute. Since my connection is `identified_by current_user`, I can put `:login` in my debugger array and it will pull out "leastbad" if I'm logged in, "-" if I'm not.

To get this working, I had to wire up the broadcaster classes to track their own CableReady operations in an array of arrays that takes the form `[["body", :inner_html], ["#user", :morph]]`.

In my application, I use
```rb
  config.logging = ["[", :session_id, "] ", :cyan, :login, " ", :magenta, :operation_counter, " ", :green, :reflex_info, " -> ", :white, :selector, " ", :yellow, :operation, :white, " via ", :red, :mode]
end
```
and this gives me

WindowsTerminal_4xUBZz38jp

while the template included in the PR is:
```rb
  config.logging = [:timestamp, :red, " [", :session_id, "] ", :magenta, :operation_counter, " ", :green, :reflex_info, " -> ", :white, :selector, " ", :yellow, :operation, :white, " via ", :blue, :mode, " Morph", :cyan, " to ", :connection_id]
end
```

WindowsTerminal_htBSOA3yN4

So you see it is super flexible.

Now, my major outstanding gripe is that the configuration array format I suggested is way ugly! I wonder if we could adapt it to take a lambda where we pass in the Reflex and parse it somehow. Just thinking out loud...

Anyhow, what do you think!?

app/channels/stimulus_reflex/channel.rb
piotrwodz: _"The correct thing to do is provide good instructions, not impose our logic on how the developer wants to configure their Rails app. They could want both!" ❤️_

100% agree with it @leastbad

piotrwodz:
> So you see it is super flexible.
> 
> Now, my major outstanding gripe is that the configuration array format I suggested is way ugly! I wonder if we could adapt it to take a lambda where we pass in the Reflex and parse it somehow. Just thinking out loud...
> 
> Anyhow, what do you think!?

Of course it is flexible 🎸 That's what we are aiming for.

I aprreciate your contribution.

Let me see how it could work out with the lambda. If it turns out to be the right way to go, then I would get rid of an array by all means.

leastbad:
https://www.ruby3.dev/string-theory/2020/11/05/gsub-blocks-partitions-string-scanners-oh-my/

Hey @piotrwodz, @jaredcwhite wrote a great post on different clever ways to parse a string. I wonder if we could come up with our own micro-markup interpolation syntax.

piotrwodz:
> https://www.ruby3.dev/string-theory/2020/11/05/gsub-blocks-partitions-string-scanners-oh-my/
> 
> Hey @piotrwodz, @jaredcwhite wrote a great post on different clever ways to parse a string. I wonder if we could come up with our own micro-markup interpolation syntax.

Hey, an interesting one. In the beginning I had a similar idea but eventually I found a smarter solution. I am on it now!

piotrwodz:
Done 🎉 I am really proud of the final result!

leastbad:
I am super excited to check it out!

Can you throw some examples in the description? No need to go crazy :)

piotrwodz:
For now, one example is already included in our config template: 1a90e33

leastbad:
BOOM! I love it.

I suspect we'll `s/context/r|reflex` but this makes me super happy. Great work, Piotr.

piotrwodz:
> BOOM! I love it.
> 
> I suspect we'll `s/context/r|reflex` but this makes me super happy. Great work, Piotr.

Thank you. I love it too 💯

lib/stimulus_reflex/configuration.rb
```diff
@@ -14,11 +14,12 @@ def configuration
   end
 
   class Configuration
-    attr_accessor :on_failed_sanity_checks, :parent_channel
+    attr_accessor :on_failed_sanity_checks, :parent_channel, :logging
 
     def initialize
       @on_failed_sanity_checks = :exit
       @parent_channel = "ApplicationCable::Channel"
+      @logging = []
```
RolandStuder, line 22: @piotrwodz Should we maybe provide a default logging? People upgrading will not have the initializer with your new configuration.

leastbad: 100%

It's likely that once we have the syntax nailed down, we'll copy what's in the template so it's the actual default; reinforcing the educational potential. It likely will use fewer colors 😀

piotrwodz: Definitely.

lib/stimulus_reflex/logger.rb
```diff
@@ -0,0 +1,119 @@
...
+    def connection_id
+      connection_id_full[0..7]
+    end
+
+    def timestamp
+      Time.now.strftime("%Y-%m-%d %H:%M:%S")
+    end
+
+    def method_missing method
+      return "\e[#{COLORS[method]}m" if COLORS.key?(method)
```
RolandStuder, line 94: I wouldn't use method missing to create the color methods, if we know which ones we create, I would use `def_method`

```ruby
COLORS.each do |name, code|
  define_method(name) { "\e[#{code}m" }
end
```

piotrwodz: Metaprogramming! I like this suggestion, Roland 👍

lib/stimulus_reflex/logger.rb
```diff
@@ -20,26 +20,16 @@ def initialize(reflex)
     end
 
     def print
-      return unless config_logging.lambda?
-
       puts
       reflex.broadcaster.operations.each do
-        puts log + "\e[0m"
+        puts StimulusReflex.config.logging.call(self) + "\e[0m"
```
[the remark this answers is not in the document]
leastbad: That's smart. Good catch. :)

lib/stimulus_reflex/logger.rb
```diff
@@ -0,0 +1,105 @@
...
+        puts config_logging.call(self) + "\e[0m"
+        @current_operation += 1
+      end
+      puts
+    end
+
+    private
+
+    def config_logging
+      return @config_logging if @config_logging
```
piotrwodz, line 25: @RolandStuder Where does this instance variable come from?

piotrwodz: Is this the one defined in line 28? I am having difficulties with grasping what's happening here.

lib/stimulus_reflex/logger.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      StimulusReflex.config.logging.binding.eval("using StimulusReflex::Utils::Colorize")
+      @config_logging = StimulusReflex.config.logging
```
piotrwodz, line 28: I liked `StimulusReflex.config.logging` being a method. Especially if you take into account the fact we call it twice.

lib/stimulus_reflex/logger.rb
leastbad: Yes, line 28! It's memoizing to @config_logging

lib/stimulus_reflex/logger.rb
leastbad: We call it twice in the same file, but don't mutate it or need it memoized... I dunno, seems like overkill, no?

lib/stimulus_reflex/logger.rb
piotrwodz: Ah `@config_logging` is nil in a first run but then it gets assigned an actual value. I see now. Thanks.

lib/stimulus_reflex/logger.rb
piotrwodz: Well, we actually do memoize it as you explained [here](https://github.com/hopsoft/stimulus_reflex/pull/337/files#r523287037), don't we? Personally, I find this method harder to read.

RolandStuder: We need the lambda to be  `using Colorize`, so we need to do the `.binding.eval` somewhere. 

We could put it in the `configuration.rb` and have a special setter that does the `.binding.eval`. I shortly went that route, but then thought the context is clearer if do the binding stuff is where the logger calls it. It is more of a concern of the logger than the configuration, that this config should be colorizable.

I then memoized it, because not every call to the logger should again do the `binding.eval("using …")`, and it's probably not great for performance to do the `binding.eval` every time a reflex gets logged. We mutate its binding, so yes I think we need to memoize it or have a setter that mutates the binding!

I would be fine to move the binding stuff into the `configuration.rb`. Would have the advantage that I can change `config.logging` at runtime (though I would not know why you would want to do that, well actually you might want to play around on the console, while finding your perfect log formatting :-) )

RolandStuder: btw: By setter I mean a writer, something like this:
```ruby
class Configuration
  def logging= (lambda)
    @logging = lambda.binding.eval("using StimulusReflex::Utils::Colorize")
  end
end

**Comment by RolandStuder** [comment]
I made a force push, if you want to work on this branch you should reset to the remote repo state:
https://www.scivision.dev/git-pull-after-force-push/

File: lib/stimulus_reflex/logger.rb (hunk 7 above)
Comment (RolandStuder) [reply lib/stimulus_reflex/logger.rb:28]: Sleeping on it, I think it would be more ergonomic to do it with the writter in the config. I myself found a bit annoying having to restart the server, when changing the logging lambda, to see the changes reflected.
@piotrwodz Do you want to do that?

Comment (piotrwodz) [reply lib/stimulus_reflex/logger.rb:28]: @RolandStuder Tho minds think alike 👍

Yes! It is definitely going to improve things. Not only the fact of changing lambda without a restart but also currently `config_logger` method is responsible for two things and I would like to change it. It should be memoizing `StimulusReflex.config.logging` lambda only.

I would like to do that! Thanks!

Comment (RolandStuder) [reply lib/stimulus_reflex/logger.rb:28]: @piotrwodz Memoizing will no longer be necessary in `Logger` if you adjust the writer. Just make sure, that in the `def initialize` you also do the binding, or better yet you also use the setter/writer. `@logging = -> {}` does not call `def logging=`

**Comment by leastbad** [comment]
I have just run `git reset piotrwodz/logger-server-side --hard` and nothing is on fire.

What is the status of the `config_logger` stuff you guys were discussing? I have had my eyes and brain elsewhere, but I'd like to be able to tell Nate that this branch could be merged...

**Comment by piotrwodz** [comment]
@leastbad I am going to incorporate what Roland and me discussed yesterday but I am out half a day today. Will take care of it in later hours. Hope you can wait for it!

File: lib/stimulus_reflex/logger.rb (hunk 7 above)
Comment (piotrwodz) [reply lib/stimulus_reflex/logger.rb]: @RolandStuder I spent some time on it ☝️

Unfortunately, this ain't going to work out. We cannot define a setter like proposed above. Why? Because it does not refer to our lambda anymore. It points to its binding instead which as a result causes `print` to crash.

What I decided to do then was separation of `StimulusReflex.config.logging` memoization and setting config logging binding.

**Comment by piotrwodz** [comment]
@leastbad @RolandStuder I think it is ready to go! 🚗 🚀

File: lib/stimulus_reflex/logger.rb [hunk 8]
```diff
@@ -12,6 +12,7 @@ def initialize(reflex)
     def print
       return unless config_logging.lambda?
       puts
+      set_config_lambda_binding
```
RolandStuder, line 16: Hm, not quite what I meant. This will still do the `.eval` on every reflex. The memoization only matters for multiple operations in the reflex.

piotrwodz: Ah, wait. Does it mean the purpose of memoization was to do `eval` once?

piotrwodz: @RolandStuder My apologies. I misunderstood the intentions there. I thought what was memoized was `StimulusReflex.config.logging` but there was no `||=` which I missed. The trick was to avoid `eval` there. Restored the previous version!

Surely, we can improve on the `eval` and move it to `Configuration` but as I mentioned [here](https://github.com/hopsoft/stimulus_reflex/pull/337#discussion_r523794571) definig a setter did not really work out.

I am all good to merge this PR in a current shape ✅ Any improvements can be done afterwards. What do you think?

RolandStuder:
I am doing some further learning. Im at my depths here. But I think we could get the API even down do
`{"session_id".red}`by using `instance_eval` we could then actually get rid of the method_missing stuff, and allow the lambda or prod to just access whatever is available in the Logger instance.

So I would propose to wait still a couple of days. I'll try to get it working today.

RolandStuder:
I am doing some further learning. Im at my depths here. But I think we could get the API even down do
`{"session_id".red}`by using `instance_eval` we could then actually get rid of the method_missing stuff, and allow the lambda or prod to just access whatever is available in the Logger instance.

So I would propose to wait still a couple of days. I'll try to get it working today.

leastbad:
Getting rid of the method_missing would undermine why I wrote it, wouldn't it? That is, I want to be able to access AR attributes on objects in the connection's identifiers.

leastbad:
Also, just a meta observation: you guys might be taking something that was as good as done and thinking about it too much. I am going to re-read all of the comments again, but it's not at all clear to me what is being improved right now. I say this with kindness, not frustration; i just know it all seemed done on Friday, but now it somehow isn't anymore.

piotrwodz:
@leastbad I agree.

FYI: the current state of this PR is the same as what we had on Friday. I am good to call it: done!

leastbad:
Great work, @piotrwodz and @RolandStuder. I just tested it again and all looks good. I also added some words to the PR description. Hopefully @hopsoft will have some time to check it out soon.

hopsoft:
Looks like we have a conflict. Happy to approve once resolved.

PR #338, opened 2020-10-26
Aliases method_name to action_name

obie:
Aliases method_name to action_name for better compatibility with controller libraries like Pundit

Enhancement

## Description

One-line change, aliases method_name to action_name

Fixes # (issue)

## Why should this be added

Provides compatibility with Pundit and potentially other libraries that enhance the controller layer.

## Checklist

- [x ] My code follows the style guidelines of this project
- [x ] Checks (StandardRB & Prettier-Standard) are passing

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -45,6 +45,8 @@ class StimulusReflex::Reflex
 
   attr_reader :channel, :url, :element, :selectors, :method_name, :broadcaster, :permanent_attribute_name
 
+  alias action_name method_name # for compatibility with controller libraries like Pundit that expect an action name
+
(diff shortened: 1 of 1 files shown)


leastbad:
LET'S GOOO

hopsoft:
My only reservation with this is that it further blurs the line between Reflex and Controller. We've had some users get confused about the distinction at times because they are so similar in nature. Let's just be very clear in the docs that a Reflex is similar to, but is definitely not a Controller.

leastbad:
My intention was to just not say anything at all; to me, this doesn't feel like a feature enhancement so much as a concession to make tools like Pundit work better. We don't really tell people about the @method_name accessor anyhow, right?

PR #339, opened 2020-10-26
Make StimulusReflex configurable and add an initializer

RolandStuder:
# Enhancement
## Description

As several PullRequest wanted to introduce some sort of configuration / initialization file, we need a shared solution. Affected PullRequests are: https://github.com/hopsoft/stimulus_reflex/pull/322 / https://github.com/hopsoft/stimulus_reflex/pull/337 / https://github.com/hopsoft/stimulus_reflex/pull/318 

So this PR provides a common implementation. It changes:
* Adds configuration object on `StimulusReflex.configuration`
* Adds generator for Config file `/config/initializers/stimulus_reflex.rb`
* Generates config file on install task.

## Usage

In an `initializer` or your `application.rb`

### You can configure using a block with:

```ruby
   config.example = :some_value
end
```
### Or direct assignment
```ruby
StimulusReflex.configuration.example = :other_value
# or with the alias
StimulusReflex.config.example = :other_value
```

### Generate a config file

`bundle exec rails generate stimulus_reflex:config`

## Why should this be added

This way we can configure in a consistent way and acomodate different needs, where it makes sense.

## Acknowledgements

* Generator code is mostly _stolen_ from @piotrwodz
* Config implementation mostly inspired by https://thoughtbot.com/blog/mygem-configure-block

## How to implement new config vars:

In `lib/stimulus_reflex/configuration.rb`
* Add `attr_accessor :var` to `Configuration`
* Add a default value in `initialize`

In `lib/generators/templates/config/initializers/stimulus_reflex.rb`
* Describe what the config variable does and explain the options
* Give an example _or_ repeat the assignment of the default (whatever makes more sense for that case)

## Checklist


RolandStuder:
@dark-panda @piotrwodz This PR has a implementation for making SR configurable. Please comment if you agree with the implementation. The idea would then probably be to adjust our PRs to make this all compatible.

piotrwodz:
@RolandStuder Thanks. Totally, this solution is very similar to what you can find in #337 so I agree.

lib/stimulus_reflex/configuration.rb
```diff
@@ -0,0 +1,23 @@
...
+      @configuration ||= Configuration.new
+    end
+
+    alias config configuration
+  end
+
+  class Configuration
+    attr_accessor :example # put your attr_accessors here, like `attr_accessor: debug`, remove this line when you do.
+
+    def initialize
```
piotrwodz, line 19:
```suggestion
    # Assign a default to a configurable variable
    # def initialize(example)
    #     @example = true
    # end
```

piotrwodz (commented):
Hey @RolandStuder thanks for this.

I left two remarks in a configuration class. Suggesting some code to be behind a comment so that it is more an _indication_ on how to do it rather than actual code implementation.

lib/stimulus_reflex.rb
```diff
@@ -18,6 +19,7 @@
 require "stimulus_reflex/broadcasters/page_broadcaster"
 require "stimulus_reflex/broadcasters/selector_broadcaster"
 require "generators/stimulus_reflex_generator"
+require "generators/config_generator"
```
dark-panda, line 22: This line and the line above it don't actually need to be here and doesn't need to be required into the application if you restructure the directories a little bit. They will be autoloaded if you place them in `lib/generators/stimulus_reflex/` instead of `lib/generators/`. You could then delete these two `require` statements. This would let Rails do its thing when it comes to finding generators on its own as well as avoid unnecessarily adding more classes and such into the actual running Rails app.

RolandStuder: Oh, nice, thx, I will incorporate that.

lib/stimulus_reflex/configuration.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def configuration
+      @configuration ||= Configuration.new
+    end
+
+    alias config configuration
+  end
+
+  class Configuration
+    attr_accessor :example # put your attr_accessors here, like `attr_accessor: debug`, remove this line when you do.
```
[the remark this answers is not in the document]
RolandStuder: True, I just thought it would be slightly clearer that this should be removed when the first person actually adds a real variable.

lib/stimulus_reflex/configuration.rb
RolandStuder: dito, but I'll do it, it is a bit cleaner not having an actual dummy config var implemented. Though on second thought, without an implemented variable, you cannot test this out with your app.

piotrwodz: The problem is: you have nothing to be configured/set up yet. This PR is a groundwork for future configurable variables like `debug`, `logging` etc.

RolandStuder: I understand that. My point was, if somebody wants to check out this branch, they can test it with the example config variable. But that point will be moot, once this PR gets merged. 

I mean basically it shouldn't matter anyway, as there should be no release with this PR merged, if we don't also merge one of the PRs that actually needs this :-)

lib/stimulus_reflex.rb
RolandStuder: @dark-panda I tried to do it but I ran into a limitation with that approach: As we the commands to be
* `rails generate stimulus_reflex example` and
* `rails generate stimulus_reflex:config`

For the magic to work, the reflex generator has to be `StimulusReflexGenerator`
and the config generator has to be `StimulusReflex::ConfigGenerator` (as far as I understand)

So to make it work, it expects the corresponding template files to be corresponding folders, which means some template files will be in `lib/generators/stimulus_reflex/config` but the others in `lib/generators/` (one module-namespaced under `/stimulus_reflex` the other not)

I find that somewhat confusing, to have to search for the templates in two places.

* Do you agree that having two template folders is confusing, or is that a better practice, because Rails wants it that way?
* Or is there another way to do this? Any suggestions on what would be best to do here?

RolandStuder: Idea from @ParamagicDev just create some rake tasks, then the commands can be independent of the internal structure. 

> Is the goal just to add rake tasks keep the CLI the same? Because you could easily create a tasks folder that holds both and you can define the commands there's in a stimulus_reflex.rake file. 

On the other hand, are we bikeshedding here? (Genuine question)

dark-panda: @RolandStuder It seemed to work okay for me without any special renaming. Here's my directory structure:

```
lib/generators/stimulus_reflex/config_generator.rb
lib/generators/stimulus_reflex/templates/app/reflexes/application_reflex.rb.tt
lib/generators/stimulus_reflex/templates/app/reflexes/%file_name%_reflex.rb.tt
lib/generators/stimulus_reflex/templates/app/javascript/controllers/%file_name%_controller.js.tt
lib/generators/stimulus_reflex/templates/app/javascript/controllers/application_controller.js.tt
lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
lib/generators/stimulus_reflex/stimulus_reflex_generator.rb
```

All of the templates generated fine and I didn't touch the code in any files at all other than to remove the `requires`.

RolandStuder: @dark-panda Yeah, you are right, must have messed up something. Now everything is in autoloaded paths and I was able to remove the requires. Thx for your help.

RolandStuder: @dark-panda Hm, but now I had to require the Generator in the Testfile, I tried to move the testfile as well into `test/generators/stimulus_reflex/stimulus_reflex_generator_test.rb` but with no success (I get `NameError: uninitialized constant StimulusReflex::StimulusReflexGenerator` or if I adjust `test ::StimulusReflexGenerator` I get `NameError: uninitialized constant StimulusReflexGenerator` unless I require the generator in the test file.

Autoloading is somewhat obscure to me. Happy to accept a PR towards this branch ;-)

dark-panda: Not a problem, I’ll write one up tomorrow.

leastbad: Hey gang, I'm doing my best to follow along but autoloading is black magic to me.

I'm hovering around the merge button because this PR unlocks us moving forward with several others (including yours, @dark-panda!) but it's not clear if the Testfile/Generator stuff is actually blocking. If I merge this, is the PR you're discussing something that could be made against the master branch instead?

I don't want to rush anyone, either. If you tell me to chill out, I will. 😁

leastbad:
Thanks for putting this together, @RolandStuder and taking great care to make sure that it's the best of all takes.

I'm really proud of this contribution to the library. Like Roland, it is an enabler of great things.

RolandStuder:
@dark-panda The config PR is merged, but you can of course still do a PR on master to improve the Autoloading stuff.

PR #341, opened 2020-10-28
Get package version from second line in the yarn.lock

RolandStuder:
# Bugfix

## Description

People sometimes get their NPM package directly from Github, by putting something like the following in their `package.json`:
```
    "stimulus_reflex": "leastbad/stimulus_reflex#isolation_optional"
```

In this case the sanity checker does not manage read the version (as is looked at only the first line of the `yarn.lock` entry like `stimulus_reflex@3.3.0`

So in this case people mistakenly get a version misleading info of not being able to locate the stimulus_reflex package.

## What changes

This PR makes it so the version get read from the second line of a `yarn.lock` entry (see an example entry here:)
```
stimulus_reflex@leastbad/stimulus_reflex#isolation_optional:
  version "3.3.0"
  resolved "https://codeload.github.com/leastbad/stimulus_reflex/tar.gz/543fb0f1c543e7679ddb9720cc6ddfeefd40f898"
  dependencies:
    form-serialize ">= 0.7.2"
```

Fixes the red herring we had here: https://discord.com/channels/629472241427415060/733725826411135107/770995207918845954

## Why should this be added

Value is small, but will make it easier to have have confusing information on startup in such cases.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/sanity_checker.rb
+++ b/lib/stimulus_reflex/sanity_checker.rb
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
 class StimulusReflex::SanityChecker
-  NODE_VERSION_FORMAT = /(\d+\.\d+\.\d+.*):/
-  JSON_VERSION_FORMAT = /(\d+\.\d+\.\d+.*)"/
-
   def self.check!
     instance = new
     instance.check_caching_enabled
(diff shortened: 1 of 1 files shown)


RolandStuder:
I actually wonder, if we should not just always read the version from the `package.json` in the `node_modules/stimulus_reflex` folder. Is there an advantage to read it from `yarn.lock`?

"Historically" we just used the lockfile and @leastbad, wanted it to work with `yarn link`, but his implementation should cover all cases, should it not?

RolandStuder:
This is already covered in #318, only checking the `package.json`
see: https://github.com/hopsoft/stimulus_reflex/pull/318/files#diff-b4d8b4e02fbd526078f034c3dcc69d6b0ff91d1c684fecc396c7d738cb3d1dd9R68

Therefore closing this

PR #343, opened 2020-10-30
Use Webpacker folder if available

coorasse:
# Type of PR: Enhancement

## Description

My installation of stimulus_reflex was simply failing because non of the `filepath` were found.

I have a different folder than app/javascript in my Rails App: this is already defined in the webpacker configuration. 
[Of course, people might not have Webpacker installed...but maybe we could use it if it is available.](https://www.youtube.com/watch?v=XLGzFQg_1xc) :) 

## Checklist


leastbad:
Sure, I'll bite. 😀 

Nice catch!

Issue #333, opened 2020-10-17
Proposal to separate the javascript into two parts; core and rails specific.


jonathan-s:
# Feature Request

Right now the stimulus reflex javascript library is in a single file molded to work perfectly with actioncable. I would like to see a separation into two javascript packages; `stimulusreflex-core` and `stimulus_reflex`.

`stimulus_reflex` would depend on `stimulusreflex-core` and this package would be tailored to work with rails and thus have a dependency on actioncable. 

`stimulusreflex-core` would contain all the logic for making stimulusreflex work the way it does today except for the part where it assumes that actioncable is the websocket connection. 

This feature request can only be dealt with after we've done satisfactory work on #182. #345 Does some excellent groundwork as well, and if that gets merged that would also set us up to split the library into a core and 'rails' specific implementation.

## Is your feature request related to a problem?

As a django user I'm able to use the javascript version of stimulus reflex. Right now it depends on the javascript actioncable library, as a django user I wouldn't want that to be included.

leastbad:
Hey @jonathan-s, as you know @julianrubisch has been doing some high-quality thinking about how to make SR more modular with an adapter approach, especially around transport mechanisms.

It seems like in order to support other transports, we'd have to compartmentalize ActionCable anyhow. This would seem to (somewhat) implicitly address the spirit of this issue, because ActionCable would become just another adapter.

This would all very definitely require a major SemVer bump, so we do have some latitude to make big changes. However, the other side of the coin here is that this isn't a trivial adaptation and we are currently chasing other priorities so I want to be realistic about the degree to which this is on an unknown timeline. Hopefully Q2 2021? I'm spitballing.

One thing that could be really helpful is a triage of the API surface to identify which chunks of code would ultimately be moved into an adapter. I don't expect you to own this on the server, but on the client, there's some good conversation to be had around, for example, what I call "library-level" events for connectivity. Eg. does an adapter have to support these events? What does it mean to be connected or disconnected when you're using Ajax? Is there a heartbeat involved?

jonathan-s:
Thanks @leastbad, I definitely understand that this isn't a trivial update and needs some thinking around it. I added this issue so we can start some thinking around it for this very reason. 

You also pose some great questions around what an adapter should support. I'll have a think about this and provide some input here!

julianrubisch:
[...]

PR #345, opened 2020-10-31
Introduce `@stimulus_reflex/polyfills` and prepare for a multiple npm package future

marcoroth:
# Type of PR 

Feature

## Description

This Pull Request addresses two parts:

1. The `@stimulus_reflex/polyfills` package 
2. and a skeleton to organize our JavaScript code in multiple packages with [lerna](https://github.com/lerna/lerna)

### 1. Polyfills for StimulusReflex

This Pull Request contains the code for the `@stimulus_reflex/polyfills` package. The polyfills-package provides support for StimulusReflex and CableReady in older browsers like Internet Explorer 11. 

#### Usage

To include the polyfills you just have to import the package and you are set! Typically you want to import it in `app/javascript/packs/application.js`.

```javascript
// app/javascript/packs/application.js

import '@stimulus_reflex/polyfills'
import 'controllers'

// ...
```

#### Motivation

Splitting up the StimulusReflex JavaScript code into multiple packages gives us few benefits. While the usage for the developer still stays the same (just include the `stimulus_reflex` package) we are really flexible to manage our code in different packages behind the scences. 

The `@stimulus_reflex/polyfills` is the first package we want to introduce alongside the current JavaScript package. The pollyfills shouldn't be included in the "core" package if you don't need IE11 support. With this we can make it optional to use this package while still maintaining all the JavaScript package in the same repository instead of maintaining them in independent repositories. 

Plus it would provide the architecture to support different transport adapters and other alike packages in the future. For example it would enable #182, #183, #196 and #333 to become a reality. 

I could see something like this in the future:

```javascript
import { Application } from 'stimulus'
import { definitionsFromContext } from 'stimulus/webpack-helpers'

import { SomeAdapter } from '@stimulus_reflex/some-adapter'
import { SomePlugin } from '@stimulus_reflex/some-plugin'

import consumer from '../channels/consumer'

const application = Application.start()
const context = require.context('controllers', true, /_controller\.js$/)
application.load(definitionsFromContext(context))

StimulusReflex.initialize(application, { consumer, adapter: SomeAdapter, plugins: [ SomePlugin ] })
```

#### Current setup

Lerna allows to setup multiple packages in one repository, typically under the `packages/` directory. Every folder under the `packages/` directory is seen as it's own package. Directories beginning with an `@` are namespaced packages.

Currently it contains three packages:

* `stimulus_reflex`
* `@stimulus_reflex/core`
* `@stimulus_refelx/polyfills`

##### The `stimulus_reflex` package

The `stimulus_reflex` packages is meant to "compose" all of our packages into a single one with all the "default" packages the user expects to get. In our case it currently just exports everything from the `@stimuls_reflex/core` package. But in a future where we would have different adapters and other optional packages we can compose the defaults in this package. This package is also here for backwards compatibility. The user still just needs to `yarn add stimulus_reflex` and is set.

##### The `@stimulus_reflex/core` package

Before we split up our current JavScript package into multiple single packages the `@stimulus_reflex/core` is meant to contain all the code from our current JavaScript package under `javascript/`. So we don't have the effort to split up the current packges just yet. We just need to move it here, so we have it easier in the future when we are about to split it up. I just didn't move all the stuff in there yet, because we have a lot of moving parts and open PRs. So I figured to wait until everything setteled a little bit to prevent conflicts. 

##### The `@stimulus_reflex/polyfills` package

As described in the first part of this PR this package is here to add support for older browsers and can be included optionally.

#### Local development setup

Once you checked out this branch you can run `yarn bootstrap`. This will set up all the packages on your local machine including the build to the `dist/` directory in all packages.

While every package is an independent npm package you can still `cd` into the correspoding folder and do whatever you would do in a regular npm package. Lerna is just here to glue everything together.

Lerna provides also a few nice "helpers" to run commands in all packages. In the root of the project you can run `yarn build` for example which runs `lerna run build` in the background. The `lerna run <command>` command will execute the given command as an npm-script in every package. Pretty handy!

Lerna also helps with releasing the npm packages. There are commands to increment the package version for every package "globally" and to release the packages pretty easily in an single batch.

---

Let me know what you think about all of this. I'm pretty exciting about this change and I think it will help us to push the framework even further!

## Checklist


jonathan-s:
This is very exciting. It touches upon #333 in some ways.

packages/@stimulus_reflex/polyfills/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "dependencies": {
+    "@stimulus/polyfills": "^1.1.1",
+    "@webcomponents/template": "^1.4.4",
+    "core-js": "^3.4.0",
+    "formdata-polyfill": "^3.0.20"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
```
ParamagicDev, line 31: Should we perhaps add a sideEffects declaration as well?

Something like: 

```json
sideEffects: [
  "./index.js",
  "./polyfills/**/*.js"
]
```

This helps Webpack optimize treeshaking. Here's the link to the docs.

https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

marcoroth: Thank you @ParamagicDev, good catch! I think it make sense to declare them as side effects. I'll update the PR!

Do you know if there is a way to test that the optimization really kicked in?

ParamagicDev: Appears you may just have to link and build and see what happens.

This was the best I could find:

https://github.com/webpack/webpack/tree/master/examples/side-effects

marcoroth:
I just extracted the `cable_ready` specific polyfills to it's own package. See hopsoft/cable_ready#77

jonathan-s:
Every time I see an update to this I wait in anticipation!

marcoroth:
I'm sorry @jonathan-s! The current purpose of this PR is to have the `@stimulus_reflex/polyfills` package somewhere 😅 
I guess the other aspect(s) of this PR needs to be addressed separately.

marcoroth:
I'm going to close this since `@stimulus_reflex/polyfills` has been extracted into https://github.com/stimulusreflex/polyfills

PR #347, opened 2020-11-02
StimulusReflex.debug= on left hand side

drnic:
I think this change -- setting `StimulusReflex.debug` -- reads a little cleaner; and sits nicely underneath `StimulusReflex.initialize`.

```
StimulusReflex.debug = process.env.RAILS_ENV === "development"
```


drnic:
Ahh damnit. My Prettier prefers double quote; so my github edit has double quotes.

PR #349, opened 2020-11-03
Fix serializeForm initialization

marcoroth:
# Type of PR

Bug fix

## Description

@thylacinelol (and probably @acetinick) noted on Discord, that their reflexes using forms stopped working while upgrading to `3.4.0.pre1`.

The default option for `options['serializeForm']` is `undefined`. With the value of `undefined` or `true` the  client should serialize the form. With the value `false` you opt-out of the form initialization and it should not send any from params to the backend.

Currently you had to opt-in with `serializeForm = true` to get it working again. It should be the other way around.

This PR fixes this logic. 

## Why should this be added

The current logic would break backwards compatibility.

## Checklist


leastbad:
Ugh, this is almost certainly my doing. Thanks for saving the day, again, @marcoroth

PR #351, opened 2020-11-05
move `cable_ready` to development dependencies

marcoroth:
# Type of PR

Enhancement

## Description

This PR enforces users to add `cable_ready` as an explicit dependency to their app's `Gemfile`.

## Why should this be added

This is to match the behavior introduced in #315. Moving `cable_ready` to the `peerDependencies` resulted that users now need to add `cable_ready` as a dependecy to the app's `package.json`.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,6 @@ PATH
   remote: .
   specs:
     stimulus_reflex (3.4.0.pre1)
-      cable_ready (>= 4.3.0)
       nokogiri
       rack
       rails (>= 5.2)
@@ -177,6 +176,7 @@ PLATFORMS
 
 DEPENDENCIES
   bundler (~> 2.0)
+  cable_ready (>= 4.3.0)
   pry
   pry-nav
   rake
--- a/lib/tasks/stimulus_reflex/install.rake
+++ b/lib/tasks/stimulus_reflex/install.rake
@@ -6,6 +6,7 @@ require "stimulus_reflex/version"
 namespace :stimulus_reflex do
   desc "Install StimulusReflex in this application"
   task install: :environment do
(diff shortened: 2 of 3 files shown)


leastbad:
Wait, why do we want to move the `cable_ready` gem dependency to `development_dependency`?

marcoroth:
Users now need to explicitly add `cable_ready` as a dependency to their `package.json` in StimulusReflex `3.4.x`. From a consistency point of view it would make sense to do the same on the gem side. So that you explicity need to add the `cable_ready` gem to your `Gemfile`.

New apps are already covered by the installer. Existing apps upgrading from `3.3.x` to `3.4.0` just require to run `yarn add cable_ready` and `bundle add cable_ready`.

julianrubisch:
this is already merged but... shouldn't that be a `runtime_dependency` ? what's the difference? 🤯

leastbad:
If I merged prematurely, that's on me. Luckily, we're not running out of numbers for PRs. :)

julianrubisch:
I have little experience, but it seems a `runtime_dependency` would be the equivalent of a `peerDependency`... maybe @marcoroth can weigh in?

marcoroth:
@julianrubisch As I understood `add_runtime_dependency` is more like `add_dependency`  than `add_development_dependency`.

Looking at [this](https://guides.rubygems.org/specification-reference/#add_development_dependency) seems like that `add_development_dependency` is the way to go to match the `peerDepedencies` bahavior:
> Development dependencies aren't installed by default and aren't activated when a gem is required.

Also:
https://guides.rubygems.org/patterns/#declaring-dependencies

If you think this isn't right, please let me know!

julianrubisch:
Great, thanks for clarifying!

PR #352, opened 2020-11-06
pass reflex id to reflex

joshleblanc:
# Enhancement

## Description

A reflex ID is generated on the client side and is used to uniquely identify a reflex. 

It seems to reason that this property should belong to the reflex class as well, instead of floating in the ether. Specifically, I need this when extending the `Reflex` class, and manually calling `cableready.morph`, since the SR is expecting to receive it on the client side.

Just as a little explanation, right now the `data` object that the channel receives is passed around to the broadcasters, so the `reflexId` is never actually available outside of the channel. This makes it available.

## Checklist


leastbad:
Hey Josh, could you rename the server-side variable to `reflex_id` (not `id`) and make sure it passes muster with StandardRB, please?

PR #354, opened 2020-11-06
Add guard clause to return valid empty form data

julianrubisch:
# Bug fix

## Description

After a lot of headache @marcoroth and me found out that `new FormData(null)` throws an error in Chrome, but does not in safari.

The `closest('form')` call in `stimulate` would return `null` in case no form is present up the DOM hierarchy, blowing up `serializeForm`, but just on Chrome 😂 

## Checklist

### DIFF
--- a/javascript/test/utils.serializeForm.test.js
+++ b/javascript/test/utils.serializeForm.test.js
@@ -4,6 +4,14 @@ import { serializeForm } from '../utils'
 
 describe('formSerialize', () => {
   context('basic', () => {
+    it('should output an empty string if no form is present', () => {
+      const dom = new JSDOM('<div></div>')
+      const form = dom.window.document.querySelector('form')
+      const actual = serializeForm(form, { w: dom.window })
+      const expected = ''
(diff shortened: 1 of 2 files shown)


jonathan-s:
Throws an error in firefox as well. So fixes it for firefox too.

marcoroth:
Yeah, also throws an error in Opera. Seems like Safari is just a little bit more forgiving 🙈

marcoroth:
@julianrubisch maybe we can add another test case for this in https://github.com/hopsoft/stimulus_reflex/blob/888c2635e4cc65efaca1d92d3b2d6949c3d91758/javascript/test/utils.serializeForm.test.js#L5

jonathan-s:
I'd be happy to volunteer some help to set up cypress. It [could have helped](https://docs.cypress.io/guides/guides/launching-browsers.html#Browsers) in a situation like this. I'd need some help on the rails part to get started though.

julianrubisch:
@ParamagicDev is our man for Cypress 😎

PR #357, opened 2020-11-10
serializeForm: only append given input if element is submit button

marcoroth:
# Type of PR
Bug Fix

## Description

@thylacinelol on Discord noticed an edge case in the `serializeForm` implementation. It turned out, that the `seralizeForm` function in combination with an `input[type=checkbox]` can cause some weird behavior, where the input always stays checked. 

It just happened when the `change` event was on the checkbox itself like:

```erb
<%= form_with model: @post, data: { reflex: "submit->PostReflex#submit", signed_id: @post.to_sgid.to_s } do |form| %>
  <div class="form-group">
    <%= form.label :is_public %>
    <%= form.check_box :is_public, data: { reflex: "change->PostReflex#submit", reflex_dataset: "combined" } %>
  </div>
<% end %>
```

If you move the `change` event to the form itself it worked as it should, like: 

```erb
<%= form_with model: @post, data: { reflex: "submit->PostReflex#submit change->PostReflex#submit" } do |form| %>
  <div class="form-group">
    <%= form.label :is_public %>
    <%= form.check_box :is_public, data: { reflex_dataset: "combined" } %>
    <%= form.submit %>
  </div>
<% end %>
```

That's because of `L24` here:
https://github.com/hopsoft/stimulus_reflex/blob/aaf81be396558798b98b8e469037f663ae8dbb61/javascript/utils.js#L23-L25

Here we just append the value of the given element to the FormData, no matter what type of input it is. This lead to the situation that the function serialized the `is_public` input twice: `is_public=0&is_public=1`, so the input `is_public` was always set.

Screenshot 2020-11-10 at 21 08 13

This PR fixes that this behavior. Now we just serialize the input if it really is a submit button, so the checkbox won't be serialized twice anymore.

Also appends the first submit button to the FormData, if the element triggering the event was no submit button. This is how the Rails Form via HTTP works.

## Why should this be added

Makes the `serializeForm` function behave like it is supposed to.

## Checklist

### DIFF
--- a/javascript/test/utils.serializeForm.test.js
+++ b/javascript/test/utils.serializeForm.test.js
@@ -9,7 +9,7 @@ describe('formSerialize', () => {
       const form = dom.window.document.querySelector('form')
       const actual = serializeForm(form, { w: dom.window })
       const expected = ''
-      assert.deepStrictEqual(actual, expected)
+      assert.equal(actual, expected)
     })
 
     it('should serialize empty form', () => {
@@ -17,7 +17,7 @@ describe('formSerialize', () => {
       const form = dom.window.document.querySelector('form')
       const actual = serializeForm(form, { w: dom.window })
       const expected = ''
-      assert.deepStrictEqual(actual, expected)
+      assert.equal(actual, expected)
     })
 
     it('should serialize basic form with single input', () => {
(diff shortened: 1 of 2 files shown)


leastbad:
Just to play devil's advocate, what if the form included a `<button type="submit">`?

marcoroth:
That's a legitimate question @leastbad. I double checked how Rails handles this. You can submit the form with a `<button>` or even a `<button type="submit">`. This also works with our current setup. But Rails and also our implementation doesn't add the button itself to the FormData.

PR #359, opened 2020-11-11
Allow to supress warnings for sanity checks

RolandStuder:
# Enhancement

## Description

Changes `confg.exit_on_failed_sanity_checks` that uses a boolean, to `config.on_failed_sanity_checks` with three options `:exit`, `:warn`, `:ignore`

Some people were seeing the warning on test runs, which can be annoying especially if you are working with TDD. So it would be better to provide an opt out possibility.

## Checklist

### DIFF
--- a/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
+++ b/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 StimulusReflex.configure do |config|
-  # Enable/disable whether startup should be aborted when the sanity checks fail
-  # config.exit_on_failed_sanity_checks = true
+  # Enable/disable exiting / warning when the sanity checks fail options:
+  # `:exit` or `:warn` or `:ignore`
+  # config.on_failed_sanity_checks = true
 
   # Override the parent class that the StimulusReflex ActionCable channel inherits from
   # config.parent_channel = "ApplicationCable::Channel"
--- a/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
+++ b/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
@@ -3,7 +3,7 @@
 StimulusReflex.configure do |config|
   # Enable/disable exiting / warning when the sanity checks fail options:
   # `:exit` or `:warn` or `:ignore`
-  # config.on_failed_sanity_checks = true
+  # config.on_failed_sanity_checks = :exit
 
   # Override the parent class that the StimulusReflex ActionCable channel inherits from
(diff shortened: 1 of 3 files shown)


RolandStuder:
Here this was reported as an issue: https://discord.com/channels/629472241427415060/733725826411135107/773114004666318858

leastbad:
One thing that is self-evident from the conversations with both AbuMareBear and Dave Sanders is that there absolutely needs to be a section in the documentation on setting up the Rails test environment so that it can work with the same general configuration used in development. Any actual testing, whether with minitest, RSpec or "does the witch float" could follow.

I am concerned that we are optimizing to minimize the annoyance of folks who (innocently and benevolently) haven't set up their test environments. It stands to reason that if they follow some basic, one-time steps, they won't be plagued with errors. Right?

1. Make sure you're running Redis
2. Configure `Gemfile` with `redis` and `hiredis`
3. Configure `test` environment of `cable.yml` similar to `development`
4. Update `test.rb` environment to set `cache_store` and enable action controller caching
5. Make sure redis server connection details are available via env or credentials

ANYHOW. 😀 

I really like the implementation you've come up with. I could be crazy but what you call `:warn` could be anything that isn't `:ignore` or `:exit`, right? Like literally any other value?

I'm glad that we're still defaulting to `:exit`, which I read as "warn, then exit" - right?

I am going to make a tiny non-functional tweak to the error message to point to the correct location.

RolandStuder:
> I really like the implementation you've come up with. I could be crazy but what you call :warn could be anything that isn't :ignore or :exit, right? Like literally any other value?

Would be fine for me, but you never know what the future holds. I almost thought to treat everything but `:exit` and `:warn` as `:ignore`, as this would have allowed for this: `.conig.on_sanity_checks = :crickets` :-)

leastbad:
What do you think? Too much?

RolandStuder:
@leastbad Nah, that's nice to offer such clear information. I like it.

lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 StimulusReflex.configure do |config|
-  # Enable/disable whether startup should be aborted when the sanity checks fail
-  # config.exit_on_failed_sanity_checks = true
+  # Enable/disable exiting / warning when the sanity checks fail options:
+  # `:exit` or `:warn` or `:ignore`
+  # config.on_failed_sanity_checks = true
```
hopsoft, line 6: We should update this example comment to use one of the new values.

```ruby
# config.on_failed_sanity_checks = :exit
```

hopsoft (changes requested):
This looks great. I had one minor nitpick.

hopsoft (approved):

PR #361, opened 2020-11-12
fix: revert CR and @rails/actioncable to dependencies

ParamagicDev:

:bug: fix

This fix revert @rails/actioncable and cable_ready to regular dependencies

This also reverts the `.gemspec` file to reflect the same.

## Description

After chatting with the guys in Discord and doing some further research, it appears CR and ActionCable are simply implementation details and not exposable to the outside world. As a result, it feels safe to revert CR and ActionCable back to regular dependencies and keep Stimulus as a peerDependency since it is extendable.

Heres a longer read on NPM package management that better explains the NPM "contract" and how "isolated" dependencies work.

https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/

## Why should this be added

This fixes issues related to users upgrading. Only people using prerelease will be affected. They simply should remove their current `cable_ready` dependency via `yarn remove cable_ready`. Users should also be safe to remove `cable_ready` from their Gemfile if they did this as well.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     stimulus_reflex (3.4.0.pre3)
+      cable_ready (>= 4.3.0)
       nokogiri
       rack
       rails (>= 5.2)
@@ -176,7 +177,6 @@ PLATFORMS
 
 DEPENDENCIES
   bundler (~> 2.0)
-  cable_ready (>= 4.3.0)
   pry
   pry-nav
   rake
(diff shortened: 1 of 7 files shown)


leastbad:
I agree with moving CableReady back to dependency, and I suspect that CR depends on ActionCable, so maybe it doesn't matter... but it seems like if ActionCable was a peerDependency it would help the Rails 5.2 folks. But wait, they changed the name of the package, so never mind... looks good to me, Konnor! 😁

marcoroth (approved):
Thanks @ParamagicDev, from the revert side of things this looks good to me!

PR #362, opened 2020-11-13
Do not run sanity check on `rails generate stimulus_reflex:config`

RolandStuder:
If a sanity check fails, the user gets a hint to create the configuration file with
`rails generate stimulus_reflex:config`, however this command will also fail because it
will also run the sanity check. Therefore we have to bypass the sanity check if
the caller is a `generate_command`.

## Why should this be added

We should add this, as the idea of the whole sanity check is to help people. But if the command we recommend fails, we might be just causing more grief.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/sanity_checker.rb
+++ b/lib/stimulus_reflex/sanity_checker.rb
@@ -3,12 +3,21 @@
 class StimulusReflex::SanityChecker
   JSON_VERSION_FORMAT = /(\d+\.\d+\.\d+.*)"/
 
-  def self.check!
-    return if StimulusReflex.config.on_failed_sanity_checks == :ignore
+  class << self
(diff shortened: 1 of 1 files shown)


lib/stimulus_reflex/sanity_checker.rb
```diff
@@ -3,12 +3,22 @@
...
+
+      instance = new
+      instance.check_caching_enabled
+      instance.check_javascript_package_version
+    end
+
+    private
+
+    def called_by_generate_command?
+      caller.any? { |call| call.match("generate_command") }
```
marcoroth, line 19: Maybe we could even simplify the check:
```suggestion
      ARGV.include? "stimulus_reflex:config"
```

PR #364, opened 2020-11-13
Add tests for broadcasters

julianrubisch:
# Tests Tests

## Description
Add test coverage for broadcasters

## Checklist


lib/stimulus_reflex/broadcasters/broadcaster.rb
```diff
@@ -9,7 +9,7 @@ class Broadcaster
 
     def initialize(reflex)
       @reflex = reflex
-      @logger = Rails.logger
+      @logger = Rails.logger if defined?(Rails.logger)
```
hopsoft, line 12: Would safe navigation work here instead of a defined check?

julianrubisch: Possibly, I will try it out!

julianrubisch: Alas, doesn't work :-(

julianrubisch: The problem is, I can't find a way to stub **module** methods with minitest... gives me `NameError: undefined method `logger' for class `Rails'` when I try ... well, because it's no class method, but a module method...

lib/stimulus_reflex/broadcasters/page_broadcaster.rb
```diff
@@ -8,7 +8,7 @@ def broadcast(selectors, data)
 
       return unless page_html.present?
 
-      document = Nokogiri::HTML(page_html)
+      document = Nokogiri::HTML.parse(page_html)
```
julianrubisch, line 11: I changed this because it's easier to stub - I think it should be just an alias?

hopsoft (approved):

PR #370, opened 2020-11-17
Pass additional reflex-related data to reflex from data

joshleblanc:
# Enhancement

## Description

xpath, cXpath, and reflexController are all needed if you want to simulate an SR call. Currently they are only available in the data object in the channel. This PR adds them to the reflex class.

## Checklist


julianrubisch:
Standardize please 😉

leastbad:
I'd like to make a counter-proposal, and I'm curious what folks think. I'm not hung up on this, but in the past we have made a reasonable effort to keep from polluting the Reflex class with every attribute passed to the channel.

A week or two ago, we added the `reflex_id`. On a high level, it makes sense to me that someone might want to be able access the `reflex_id` inside a Reflex, even though it's very much a client-side concept.

Now we're looking at adding three more client-side attributes to the server-side Reflex class, and we're getting into that slippery slope territory again.

So, what if we split the difference and create a hash with all four keys in it (`reflex_id`, `xpath`, `c_xpath`, `reflex_controller`) and call it something like `@client_routing` - a name that is highly unlikely to conflict with anything. This way, you could have access to all four attributes without polluting the Reflex class.

Does that work for you? And Julian/Nate, is what I'm saying desired and necessary, or am I overthinking this?

julianrubisch:
I concur, but I want to register my concern that we should start testing these attributes...

leastbad:
I actually wonder if we can go back in time and put `@permanent_attribute_name` in this hash, too. There's no reason for it to be exposed to the Reflex class; it only gets used in the Broadcasters and they could easily be modified to pull this value from the proposed hash.

leastbad:
@julianrubisch I have nothing but enthusiasm for testing the attributes. How do we start?

julianrubisch:
Let me add tests for the broadcasters first... let's do this "inside out"

lib/stimulus_reflex/reflex.rb
```diff
@@ -60,7 +61,8 @@ def initialize(channel, url: nil, element: nil, selectors: [], method_name: nil,
     @params = params
     @permanent_attribute_name = permanent_attribute_name
     @broadcaster = StimulusReflex::PageBroadcaster.new(self)
-    @reflex_id = reflex_id
+    @client_attributes = OpenStruct.new(client_attributes)
```
julianrubisch, line 64: I'm splitting hairs now here, but could we make that an _actual_ struct so there's a clear definition of what can go in there?

hopsoft:
I like the idea of moving `permanent_attribute_name ` into this struct as well. I vote we go for the land grab and get this all cleaned up under this PR.

leastbad:
So far, so good! Now, we just have to update the broadcasters themselves.

In `lib/stimulus_reflex/broadcasters/broadcaster.rb` on line 8, we'll want to change `permanent_attribute_name` to `client_attributes`.

In `lib/stimulus_reflex/broadcasters/page_broadcaster.rb` on line 19, we'll want to change the value to `client_attributes.permanent_attribute_name`.

In `lib/stimulus_reflex/broadcasters/selector_broadcaster.rb` on line 18, we'll want to change the value to `client_attributes.permanent_attribute_name`.

joshleblanc:
@leastbad I delegate those accessors, so no changes should be necessary there.

leastbad:
I don't understand where you delegated. How would the broadcaster classes know that we moved the permanent attribute into the new accessor?

joshleblanc:
https://github.com/hopsoft/stimulus_reflex/pull/370/files#diff-87b7a54fdec6e28f9dee9180a91acf18c74c073940340e9e6d56b5fcb9679c40R55

leastbad:
I swear that wasn't there the last time I looked. 😀

PR #375, opened 2020-11-21
Exit with nonzero status code

julianrubisch:
# Enhancement

## Description

The sanity checker should exit with a nonzero status code to indicate failure.

## Why should this be added

While performing system tests in CI, I saw the pipeline go green in a few seconds, because the sanity checker exited with code 0. Ideally, this should raise a red flag instead. I think a nonzero exit code is the correct behavior here.

## Checklist

- [ ] This is not a documentation update


leastbad:
Good catch!

PR #377, opened 2020-11-21
Handles to mitigate race conditions when running reflexes in quick succession on the same element

existentialmutt:

enhancement

## Description

This PR adds two options applications can use to get finer-grained control of when promises resolve and add custom params to the reflex.

1.  Allow `stimulate` calls to set `resolveLate: "afterFinalize"` to run promise resolution after the finalize callback and its associated reset of `element.reflexData`.

2. Allow stimulus controllers to specify `guaranteedReflexParams` that will be added to the params payload just before the stimulus leaves the client.

## Why should this be added

The finalize callback stage in the prerelease introduces a race condition when running reflexes in quick succession on the same element.  If a second reflex is initiated before the first one finalizes, the first reflex can wind up clearing data set by the second reflex's beforeReflex.  As currently implemented, reflexes resolve their promise either before or after the DOM update (depending on resolveLate) but always before finalize clears the reflexData on the element, so with the current implementation there's not a good way to know when we're past that point.

The particular way this was biting me was my app uses a beforeReflex to send a unique page_id in the params, which is used to scope client state in the Rails cache.  I'm also debouncing reflexes and running them through a client-side queue to ensure only one is out at a time.  With the current pre-release, timing issues randomly lead to the params being cleared after the page id is set but before the reflex sends.

I don't have enough experience with the code to eliminate the race condition, so this PR implements a couple handles to mitigate specific effects.  I've included these in my app and together they eliminate any adverse effects from the timing issues.

## Checklist


leastbad:
Thank-you **SO MUCH** for catching this [only now obvious] bug.

My sense is that the correct thing to do, here, is change the logic so that the `reflexData` variable actually is a dictionary that is keyed to the reflexId.

I am going to try some experiments.

PR #379, opened 2020-11-22
make element.reflexController a dictionary

existentialmutt:

bug fix

## Description

Following up on #377, this PR makes element.reflexController a dictionary to prevent multiple reflexes on the same element from overwriting eachothers' data

Fixes # (issue)

## Why should this be added

prevents race conditions when running multiple reflexes on the same element

## Checklist

### DIFF
--- a/javascript/lifecycle.js
+++ b/javascript/lifecycle.js
@@ -14,7 +14,7 @@ import { camelize } from './utils'
 //
 const invokeLifecycleMethod = (stage, element, reflexId) => {
   if (!element || !element.reflexData[reflexId]) return
-  const controller = element.reflexController
+  const controller = element.reflexController[reflexId]
   const reflex = element.reflexData[reflexId].target
   const reflexMethodName = reflex.split('#')[1]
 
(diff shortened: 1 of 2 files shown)


leastbad:
Hey Rafe, another good catch.

My understanding of things is that an element can only have one instance of a controller on it. That means we could solve this "simply" by removing line 56 of lifecycle.js, right?

I'm trying to wrap my head around a scenario where there would be multiple different Reflex controllers on an element with different names. We might have to ask @hopsoft for clarity here.

existentialmutt:
Good call- I suspect removing `delete element.reflexController` would do the trick.  I'll test that next chance I get.

hopsoft:
I'll have to think on this a bit harder, but the reason we call `delete element.reflexController ` is to clean up references that might cause memory leaks, so we'd still want to remove any references to controllers embedded in the new dictionary datatype.

leastbad:
I usually shy away from shotgun solutions, but I have an idea:
```js
if (reflexes.length == 1) delete element.reflexController
```
The premise here is that if there's only one active Reflex in the dictionary, it's the current one, so you're safe to clear out the controller reference.

hopsoft:
Makes sense, but I expect that this should be sufficient. `delete element.reflexController[reflexId]`

hopsoft (approved):

PR #386, opened 2020-11-27
Allow `success` and `after` lifecycle methods on replaced elements

marcoroth:

Enhancement

## Description

This PR allows the `success`, `after` and `finalize` lifecycle methods to be called on elements which were replaced via a morph operation.

## Why should this be added

Consistency. When you call a reflex you also expect the `success` and `after` lifecycle methods to be fired even if they were replaced via a morph operation.

## Checklist


javascript/lifecycle.js
```diff
@@ -13,7 +14,7 @@ import { camelize } from './utils'
 // - element - the element that triggered the reflex (not necessarily the Stimulus controller's element)
 //
 const invokeLifecycleMethod = (stage, element, reflexId) => {
-  if (!element || !element.reflexData[reflexId]) return
+  if (!element || !element.reflexData || !element.reflexData[reflexId]) return
```
marcoroth, line 17: We should probably warn the user if the lifecycle method couldn't be executed. If we just return here the user is probably wondering why his lifecycle method wasn't executed - because there is no indication or error messages which would suggest that the execution failed.

javascript/lifecycle.js
```diff
@@ -125,17 +126,29 @@ document.addEventListener(
...
+  const { controller } = element.reflexController[reflexId] || {}
+
   element.dispatchEvent(
     new CustomEvent(`stimulus-reflex:${stage}`, {
       bubbles: true,
       cancelable: false,
       detail: {
         reflex: target,
-        controller: element.reflexController[reflexId],
+        controller,
```
leastbad, line 151: I already really like what you did here; it's what I should have done initially. Thanks for having my back.

javascript/lifecycle.js
leastbad: As I write this, I'm still in the grokking phase but if we go down this path, we just have to make sure that we're only emitting warnings if debugging is true. That's actually a bit tricky because the functions in `lifecycle.js` don't currently have access to the `debugging` flag. This is because `invokeLifecycleMethod` is never invoked directly by the SR codebase! It's *only* invoked by the event listeners in `lifecycle.js` that drive the library's behaviour.

That means we'd have to figure out a way to pass the `debugging` flag through. Our options for that are:
- global (`window`) variable
- attach `debugging` variable to `BODY`
- add `data-reflex-debugging` attribute to body
- add `META` tag to HEAD
- move contents of `lifecycle.js` back into `stimulus_reflex.js` so that `debugger` is in file scope

I like/hate some of those more than others. @hopsoft will need to render an opinion. 🔢

javascript/lifecycle.js
```diff
@@ -125,17 +126,29 @@ document.addEventListener(
 // - element - the element that triggered the reflex (not necessarily the Stimulus controller's element)
 //
 export const dispatchLifecycleEvent = (stage, element, reflexId) => {
+  const { reflexData, reflexController } = element
+
+  if (!document.body.contains(element)) {
+    const attrs = extractElementAttributes(element)
+    element = findElement(attrs, true)
```
leastbad, line 133: [I see you.](https://github.com/hopsoft/stimulus_reflex/pull/386/files#r531824146) 👁️

javascript/lifecycle.js
```diff
@@ -125,17 +126,29 @@ document.addEventListener(
 // - element - the element that triggered the reflex (not necessarily the Stimulus controller's element)
 //
 export const dispatchLifecycleEvent = (stage, element, reflexId) => {
+  const { reflexData, reflexController } = element
+
+  if (!document.body.contains(element)) {
```
leastbad, line 131: TIL `contains`! Never thought of running this at `document.body` scope. Slick man.

leastbad:
Alright, I really, really like this even if there are some questions about how to handle `debugger` - I don't love passing `true`.

For me, the bigger concern/issue/task is that this only covers half of the problem. That is, people who don't understand how the library *behaves* who would previously be confused that their design led to their controller disappearing if they replace or delete their controller with a morph... those people are still going to be exactly the same kind of confused if they delete their controller with a morph.

That said, bringing `findElement` into the picture does mean that they would get the benefit of a warning that they don't currently get. And that might actually be the best we're going to be able to do.

The more radical approach - which I'm not convinced is a good design decision - is to test to see if a controller is on an element that is due to be replaced (using your fancy `contains` method!) and if so, we **could** move the controller to the `BODY` element. Again... is this a bad design call? Yes, the post-Reflex callbacks could fire, but would those same people who are cargo culting and not confident in their understanding know to code their callbacks to expect that they are now being run from `BODY`? It's a genuinely interesting consideration.

javascript/lifecycle.js
marcoroth: Alright, I just implement a solution for this (even though it wasn't one on your list 🙈). 
Everywhere you want to access the current debugging value you can now `impor  Debugg from './debug.js'`. With this you can access the value of it in every file you want with `Debug.enabled` or just `Debug.value`, always returns a boolean.

I you think we need another solution let me know!

marcoroth:
I now implementent some warnings. I understand that this doesn't solve the bigger concern, but at least it informs the user that he did something which is not recommended. 

For the use case, that the same element is also available after the morph we now show this message and now also execute the `success` and `finalize` callbacks method which weren't triggered before:

Screenshot 2020-11-28 at 04 59 46

For the use case, where the element is not available anymore after the morph was executed we show a warning that the lifecycle methods couldn't be executed.

Screenshot 2020-11-28 at 05 00 02

I think this is a good intermediate solution for now without breaking the backward compatibility. What do you think?

leastbad:
What I think is that it's beautiful, and I love what you did with the memoized Debug. I'm going to be borrowing that pattern in the future. **Excellent work.**

Particularly impressed by how you show them the element that was deleted. Really drives the message home.

I tweaked the language slightly for tense, case and one minor typo. (I try to consistently uppercase Reflex.)

hopsoft (approved):
This looks really good. Nice work.

javascript/lifecycle.js
```diff
@@ -125,17 +126,50 @@ document.addEventListener(
...
+  if (!element) {
+    if (Debug.enabled)
+      console.warn(
+        `StimulusReflex was not able execute the "${stage}" lifecycle method on the element which triggered the Reflex. The following element is no longer present in the DOM: `,
+        oldElement
+      )
+    return
+  }
+
+  element.reflexData = reflexData
```
leastbad, line 160: I'm a little bit concerned about these...

I think they should be:
```js
element.reflexData[reflexId] = reflexData
element.reflexController[reflexId] = reflexController
```

leastbad:
FYI I'm doing a conflict resolution on my local machine right now.

javascript/lifecycle.js
marcoroth: I think this should be fine, because we are setting of these variables with:
```js
  const reflexData = element.reflexData || {}
  const reflexController = element.reflexController || {}
```

hopsoft (approved):

PR #387, opened 2020-11-28
dont exit in sanity checker on `stimulus_reflex:install`

marcoroth:
# Type of PR

Bug Fix

## Description

I just noticed, that when you create an new Rails app with the current master ([`45e12ed`](https://github.com/hopsoft/stimulus_reflex/commit/45e12ede8a75ccb90270210dcc4a606122efd874)), that the sanity checker kicks in and aborts the `rails stimulus_reflex:install` script because it can't find the `stimulus_reflex` npm package in the `package.json`. This package isn't included yet, because this installer is responsible for adding it in the first place.

/cc @RolandStuder 

## Why should this be added

New users shouldn't be stopped by the sanity check on their first hands-on with StimulusReflex.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/sanity_checker.rb
+++ b/lib/stimulus_reflex/sanity_checker.rb
@@ -6,6 +6,7 @@ class StimulusReflex::SanityChecker
   class << self
     def check!
       return if StimulusReflex.config.on_failed_sanity_checks == :ignore
+      return if called_by_installer?
       return if called_by_generate_config?
 
       instance = new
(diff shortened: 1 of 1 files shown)


leastbad:
Good catch!

I could have sworn that #262 covered this case but I realize that it covered generators not rake tasks.

My only comment, which isn't a blocker, is that I wonder if there's a way we could standardize the approach taken by `called_by_installer?` and `called_by_generate_config?`. The answer could very well be "nope!"

I don't build a lot of tooling like this, so I'm learning in real-time.

PR #392, opened 2020-11-30
don't warn folks twice

leastbad:

## Description

Only warn developers that their first life-cycle stage isn't going to happen

## Why should this be added

Repeated messages become invisible.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -304,7 +304,7 @@ const extendStimulusController = controller => {
 
 const registerReflex = data => {
   const { reflexId } = data
-  reflexes[reflexId] = { finalStage: 'finalize' }
+  reflexes[reflexId] = { finalStage: 'finalize', warned: false }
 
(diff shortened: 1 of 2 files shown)


javascript/stimulus_reflex.js
```diff
@@ -304,7 +304,7 @@ const extendStimulusController = controller => {
 
 const registerReflex = data => {
   const { reflexId } = data
-  reflexes[reflexId] = { finalStage: 'finalize' }
+  reflexes[reflexId] = { finalStage: 'finalize', warned: false }
```
hopsoft, line 307: Not sure we need to initialize `warned: false` but I really like what this PR accomplishes.

hopsoft (approved):

javascript/stimulus_reflex.js
leastbad: When I did it, I think I was worried about cross-tab stuff... but you are correct, sir.

PR #393, opened 2020-11-30
`invokeLifecycleMethod()`: handle undefined `element` parameter

marcoroth:
# Type of PR

Bug Fix

## Description

In some situations when you triggered a reflex this error would pop in the console:

Screenshot 2020-11-30 at 18 25 55

I wasn't able to find an explanation for why and when it exactly happened. But I removed [the line](https://github.com/hopsoft/stimulus_reflex/pull/386/files#diff-1025ac8883ebdd00e6755481d958067ac87d99c5e8e3d3be4e67541cc0ed615aL16) this PR adds in #386 🙈 

## Why should this be added

It seems that everything worked even without this fix, but one potential error less for the user to stumble upon is always better ;)

## Checklist

### DIFF
--- a/javascript/lifecycle.js
+++ b/javascript/lifecycle.js
@@ -15,6 +15,8 @@ import Debug from './debug'
 // - element - the element that triggered the reflex (not necessarily the Stimulus controller's element)
 //
 const invokeLifecycleMethod = (stage, element, reflexId) => {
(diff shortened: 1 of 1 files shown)


leastbad:
It certainly does seem impossible that `invokeLifecycleMethod` could succeed without a valid element.

PR #395, opened 2020-11-30
Trigger piggy back operations after SR operations

hopsoft:
# Fix

## Description

Call any piggy back operations after standard StimulusReflex behavior. This ensures that any problems or changes introduced by such operations don't interfere with StimulusReflex behavior.

Refactor proxy object to be explicit about what is delegated so things more predictable and easier to debug.

## Checklist


javascript/stimulus_reflex.js
```diff
@@ -90,8 +88,8 @@ const createSubscription = controller => {
                 )
               )
             )
-            if (urls.length !== 1 || urls[0] !== location.href) return
 
+            if (urls.length !== 1 || urls[0] !== location.href) return
```
leastbad, line 92: What changed on this line? I'm trying to understand why the diff shows this as different.

leastbad:
I love reading your PRs because I always learn a ton. This is no exception.

I've been thinking about how you/we are moving the CableReady call to 2nd priority. I think that you could have ended the discussion sooner if you'd pointed out that the developer can still make a broadcast that will execute before the SR morphs simply by calling `.broadcast` on the chain as they are encouraged to. So in this way, the developer can have it both ways.

I feel like I should document this.

javascript/stimulus_reflex.js
hopsoft: Oh, I think it's because I deleted the empty line between this one and the next. 🤷‍♂️

hopsoft:
One interesting side effect here is that my `push_state` code that modified the URL querystring prevented the reflex from running. This was true whether or not I broadcast before or piggy backed the SR payload. By moving the additional CR operations to execute at the end, I can `push_state` and modify the URL without issue.

leastbad:
I'm going to use your `push_state` concept in the documentation. Working on it now, actually.

PR #398, opened 2020-12-05
Support for Stimulus 2

marcoroth:
# Type of PR

Enhancement

## Description

Seems like this is the only thing we need to do in order to use Stimulus 2 but still support Stimulus 1.x. We are not using `data-target` inside the StimulusReflex project itself so it seems most of the changes need to be done in user-land.

## Why should this be added

Keep up to date with the latest Stimulus

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,6 @@
     "jsdom": "^16.0.1",
     "mocha": "^8.0.1",
     "prettier-standard": "^16.1.0",
-    "stimulus": "^1.1.1"
+    "stimulus": ">= 1.1"
   }
 }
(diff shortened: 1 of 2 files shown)


hopsoft:
Nice. Have you tested with Stimulus 2.0?

marcoroth:
Yes, I tried to run both Expo and Boxdrop. Everything seems to work as intended. The only thing I noticed is the `data-target` deprecation warnings, but other than that it seems good. But will keep it as draft as of now.

leastbad:
Thanks so much for doing this, @marcoroth! I'd had serious problems with one of the later builds and SR; glad the problem 👻ed.

What a relief!

hopsoft:
FYI - I'm running @marcoroth's branch for this PR in staging for a production app with Stimulus 2. So far, so good.

hopsoft (approved):

PR #399, opened 2020-12-06
Allow StimulusReflex to process Rack middlewares

### LINKED ISSUE #289: Rendering issue
👋 

Playing with stimulus_reflex on our production app, and there is an issue I've encountered.

Our app is classic SAAS app. We store different accounts in the same database. We don't scope per subdomain, but we scope with url, for example : `app.demo/{:account_slug}/`.

Account extraction happens inside the `AccountSlug` middleware which looks like this:

```ruby
module AccountSlug
  PATTERN = /(\d{1,})/
  FORMAT = '%07d'

  def self.decode(slug) slug.to_i end
  def self.encode(id) FORMAT % id end

  # We're using account id prefixes in the URL path. Rather than namespace
  # all our routes, we're "mounting" the Rails app at this URL prefix.
  #
  # The Extractor middleware yanks the prefix off PATH_INFO, moves it to
  # SCRIPT_NAME, and drops the account id in env['docsnote.account.id'].
  #
  # Rails routes on PATH_INFO and builds URLs that respect SCRIPT_NAME,
  # so the main app is none the wiser. We look up the current account using
  # env['docsnote.account.id'] instead of request.subdomains.first
  class Extractor
    PATH_INFO_MATCH = /\A(\/#{AccountSlug::PATTERN})/

    def initialize(app)
      @app = app
    end

    def call(env)
      request = ActionDispatch::Request.new(env)

      # $1, $2, $' == script_name, slug, path_info
      if request.path_info =~ PATH_INFO_MATCH
        request.script_name   = $1
        request.path_info     = $'.empty? ? '/' : $'

        # Stash the account's ID.
        env['docsnote.account.id'] = AccountSlug.decode($2)
      end

      @app.call env
    end
  end
end
```

```ruby
# frozen_string_literal: true

module SetAccount
  extend ActiveSupport::Concern
  included do
    before_action do
      Current.account = Account.find request.env["docsnote.account.id"] if request.env["docsnote.account.id"]
    end
  end
end
```

```ruby
# frozen_string_literal: true

class ApplicationController < ActionController::Base
  include SetAccount
  ...
end
```

---------------------------------------------------------------------------------

Inside the reflex everything works fine, except page morphing, when you try to re-render current page [this line](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/stimulus_reflex/reflex.rb#L83) throws an error
Screenshot 2020-08-12 at 16 43 06

Screenshot 2020-08-12 at 16 49 32

(Which is understandable since `localhost:3000/1/` doesn't exist it needs `AccountSlug` middleware to work properly )

marcoroth:
# Type of PR

Feature / Bug Fix

## Description

This PR allows StimulusReflex to process Rack middlewares. StimulusReflex has it's own middleware stack. You need to declare which middlewares you want StimulusReflex to process on every reflex. 

It works the same way as registering a rack middleware in Rails. All the methods are available, have a look at the [docs here](https://guides.rubyonrails.org/rails_on_rack.html#configuring-middleware-stack). 

In the example of #289 the initializer would look like this: 

```ruby
  config.middleware.use AccountSlug::Extractor
end
```

Fixes #289

## Why should this be added

Any Rack middleware used in a Rails app isn't currently processed by StimulusReflex when you trigger a reflex. This could break some existing advanced apps which for example do request rewriting in a Rack middleware.

/cc @dixpac, @excid3, @leastbad 

## Checklist

### DIFF
--- a/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
+++ b/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
@@ -17,4 +17,10 @@ StimulusReflex.configure do |config|
   # eg. if your connection is `identified_by :current_user` and your User model has an email attribute, you can access r.email (it will display `-` if the user isn't logged in)
 
   # config.logging = proc { "[#{session_id}] #{operation_counter.magenta} #{reflex_info.green} -> #{selector.cyan} via #{mode} Morph (#{operation.yellow})" }
+
+  # Set `process_middleware` to `false` if StimulusReflex shouldn't process
+  # the Rails middleware stack before executing the reflex. This is useful if
+  # you don't need rack middleware processing and want to gain some performance.
+
+  # config.process_middleware = true
 end
--- a/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
+++ b/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
@@ -18,9 +18,12 @@ StimulusReflex.configure do |config|
 
   # config.logging = proc { "[#{session_id}] #{operation_counter.magenta} #{reflex_info.green} -> #{selector.cyan} via #{mode} Morph (#{operation.yellow})" }
 
-  # Set `process_middleware` to `false` if StimulusReflex shouldn't process
-  # the Rails middleware stack before executing the reflex. This is useful if
-  # you don't need rack middleware processing and want to gain some performance.
+  # StimulusReflex uses it's own middleware stack. If you need a middleware to
+  # be executed on every reflex register them here. It works the same way as
+  # registering a rack middleware in Rails.
+  #
+  # Learn more about registering rack middleware in Rails here:
+  # https://guides.rubyonrails.org/rails_on_rack.html#configuring-middleware-stack
 
-  # config.process_middleware = true
+  # config.middleware.use YourRackMiddleware
 end
--- a/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
+++ b/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
@@ -15,15 +15,15 @@ StimulusReflex.configure do |config|
   # Available colors: red, green, yellow, blue, magenta, cyan, white
   # You can also use attributes from your ActionCable Connection's identifiers that resolve to valid ActiveRecord models
   # eg. if your connection is `identified_by :current_user` and your User model has an email attribute, you can access r.email (it will display `-` if the user isn't logged in)
+  # Learn more at: https://docs.stimulusreflex.com/troubleshooting#stimulusreflex-logging
 
   # config.logging = proc { "[#{session_id}] #{operation_counter.magenta} #{reflex_info.green} -> #{selector.cyan} via #{mode} Morph (#{operation.yellow})" }
 
-  # StimulusReflex uses it's own middleware stack. If you need a middleware to
-  # be executed on every reflex register them here. It works the same way as
-  # registering a rack middleware in Rails.
+  # Optimized for speed, StimulusReflex doesn't enable Rack middleware by default.
+  # If you are using Page Morphs and your app uses Rack middleware to rewrite part of the request path, you must enable those middleware modules in StimulusReflex.
   #
-  # Learn more about registering rack middleware in Rails here:
-  # https://guides.rubyonrails.org/rails_on_rack.html#configuring-middleware-stack
+  # Learn more about registering Rack middleware in Rails here: https://guides.rubyonrails.org/rails_on_rack.html#configuring-middleware-stack
 
-  # config.middleware.use YourRackMiddleware
+  # config.middleware.use FirstRackMiddleware
+  # config.middleware.use SecondRackMiddleware
 end
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -49,11 +49,33 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
       rescue => invoke_error
         message = exception_message_with_backtrace(invoke_error)
         body = "Reflex #{target} failed: #{message} [#{url}]"
+
         if reflex
           reflex.rescue_with_handler(invoke_error)
           reflex.broadcast_message subject: "error", body: body, data: data, error: invoke_error
         else
           puts "\e[31m#{body}\e[0m"
+
+          if body.to_s.include? "No route matches"
+            initializer_path = Rails.root.join("config", "initializers", "stimulus_reflex.rb")
+
+            puts <<~NOTE
+              \e[33mNOTE: It looks like we couldn't re-render the page because we couldn't find a matching route.
+              If you are using rack middleware to rewrite part of the request path make sure you also configured
+              the middleware accordingly in the StimulusReflex initializer.
+              The initializer should be located at #{initializer_path}.
+              If it doesn't exist you can generate it with:
+
+                $ bundle exec rails generate stimulus_reflex:config
+
+              Configure the middleware in your initializer:
+
+                StimulusReflex.configure do |config|
+                  config.middleware.use YourRackMiddleware
+                end\e[0m
+
+            NOTE
+          end
         end
         return
       end
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -60,18 +60,18 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
             initializer_path = Rails.root.join("config", "initializers", "stimulus_reflex.rb")
 
             puts <<~NOTE
-              \e[33mNOTE: It looks like we couldn't re-render the page because we couldn't find a matching route.
-              If you are using rack middleware to rewrite part of the request path make sure you also configured
-              the middleware accordingly in the StimulusReflex initializer.
-              The initializer should be located at #{initializer_path}.
-              If it doesn't exist you can generate it with:
+              \e[33mNOTE: StimulusReflex failed to locate a matching route and could not re-render the page.
+
+              If your app uses Rack middleware to rewrite part of the request path, you must enable those middleware modules in StimulusReflex.
+              The StimulusReflex initializer should be located at #{initializer_path}, or you can generate it with:
 
                 $ bundle exec rails generate stimulus_reflex:config
 
-              Configure the middleware in your initializer:
+              Configure any required middleware:
 
                 StimulusReflex.configure do |config|
-                  config.middleware.use YourRackMiddleware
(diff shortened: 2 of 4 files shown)


excid3:
That was fast!

leastbad:
Great work as usual, Marco!

@excid3 can you pull this commit and report on how/if it works in your use cases?

If so, could you give me the classes people might expect to use if they are using ActsAsTenant in the way that we've been discussing? I just want to spell it out for folks in the docs such that they have nothing to bug you with. 👼

marcoroth:
@leastbad I think [this repo](https://github.com/marcoroth/stimulus_reflex-issue-289) has all the classes and the setup you'd expect. 

The thing here is, that this really has nothing to do with ActsAsTenant per se. In this case the middleware @dixpac provided in #289 just rewrites the request information in [`account_slug.rb`](https://github.com/marcoroth/stimulus_reflex-issue-289/blob/master/app/middleware/account_slug.rb#L29). This allows you to set the tenant in [`set_account.rb`](https://github.com/marcoroth/stimulus_reflex-issue-289/blob/master/app/controllers/concerns/set_account.rb#L7) based on the information which was written in the middleware. 

So both ActsAsTenant and Rails don't actually know about this account ID in the URL and just see the request as if it would have come with the ID.

excid3:
Sorry, try and erase multitenancy from the conversation, it's not about that at all. 😅 That is just one example of a middleware.

excid3 (approved):
This fixes the issue and is exactly what I planned on implementing. Good job @marcoroth 👍

excid3:
A long term solution for this is to copy the Rails middleware stack and use it in Stimulus Reflex (stripping out unnecessary items).

dixpac:
Awesome work @marcoroth 👍🏼 . This fixes the issue.
Down the line I think it would be a good idea for SRE to copy Rails middleware stack.

marcoroth:
Thanks for reporting back @excid3 and @dixpac! 

I'm with you on this!

I tried to copy the Rails middleware stack in my first implementation, but it's not as easy as you probably guessed ;) 

The main problem is, that the middleware can take different (amount of) args which makes it harder to process them in a general way in StimulusReflex.  

My first implementation looked like this. I'll attach it here so we have it somewhere:
```ruby
Rails.application.middleware.middlewares.each do |middleware|
  instance = middleware.new(Rails.application.routes)
  instance.call(env)
end
```
This will probably work when we find a way to also get the right args which need to get passed to each individual middleware.

But until we find the right way to handle this in StimulusReflex this should do it for now (as long as we don't need to support args for middleware).

marcoroth:
Never mind, looks like it's even easier to just call the Rails middleware stack 😅 

No need for a config entry in the initializer. Could you guys give this another spin?

excid3:
This is really close. 

It seems like it's mutating as it works once and then the `@app` is nil the second time in the middleware.

`undefined method `call' for nil:NilClass`

```
# Crashes here in my middleware
@app.call(request.env)
```

marcoroth:
Interesting, it threw the error for me as well - but it worked as expected. 
But I think I found the cause for that. The build method just needs the routes.

lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
```diff
@@ -17,4 +17,10 @@
   # eg. if your connection is `identified_by :current_user` and your User model has an email attribute, you can access r.email (it will display `-` if the user isn't logged in)
 
   # config.logging = proc { "[#{session_id}] #{operation_counter.magenta} #{reflex_info.green} -> #{selector.cyan} via #{mode} Morph (#{operation.yellow})" }
+
+  # Set `enable_middleware_processing` to `true` if StimulusReflex should process
+  # the Rails middleware stack before executing the reflex. This could be useful
+  # if you do request rewriting in your rack middleware.
+
+  # config.enable_middleware_processing = false
```
dixpac, line 25: Maybe nit-picking but not sure about the name that has `enable` prefix, since it will looks strange on getters such as `if config.enable_middleware_processing`, for me `middleware_processing` or `middeware_procesing_enabled` reads better

dixpac (commented):
This is awesome 💯 


> The behavior is opt-in, so you need to enable the rack middleware processing in the StimulusReflex initializer:

Personally I would set it to `true` by default, since I don't believe it would introduce meaningful performance downgrade on any real world apps, I guess it is gonna be couple of milliseconds that are gonna be unnoticeable since majority of pie will be taken by DB and network.  That being said maybe we could mesure it :)

lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
excid3: What about `config.process_middleware = true/false`? That gives you `if config.processmiddleware` which reads nicely.

marcoroth: Thanks! I update the name to `middleware_processing` 🙌🏼

excid3:
Latest changes work for me @marcoroth 👍

leastbad:
> > The behavior is opt-in, so you need to enable the rack middleware processing in the StimulusReflex initializer:
> 
> Personally I would set it to `true` by default, since I don't believe it would introduce meaningful performance downgrade on any real world apps, I guess it is gonna be couple of milliseconds that are gonna be unnoticeable since majority of pie will be taken by DB and network. That being said maybe we could mesure it :)

Here's the thing @dixpac: we've had *very* few reports of issues with this. (That doesn't make it any less shitty for those it impacts!)

Meanwhile, Nothing morphs can run in as little as 2ms. Why blow up that number when the <1% can just turn on a setting in the initializer which will be clearly described and well documented?

excid3:
Afaik, nobody has measured the impact with benchmarks so the performance argument is entirely theoretical right now.

I'd much rather this mimic Rails by default. It's much more friendly to newcomers that way which I thought was the goal. 

Maybe performance is more important that user friendliness. Personally, I was surprised that Stimulus Reflex was not using the Rails request stack out of the box. I don't want to run into that every time and would rather enable performance optimizations.

leastbad:
That's not a kind interpretation of our priorities, Chris, given our track record as a library and the fact that we've been "tools down" to find a fix that would serve a vanishingly small group of our users since you brought it up.

v3.4 isn't going out the door until we have this locked down but **I'm not confident that opting everyone into this is worth a major library version bump**. The new isolation mode - which we figure >90% of developers will want - is technically disabled by default, but the installation for new projects inserts `isolate: true` into the client-side SR initialize options. For us, this is an example of the right balance and a commitment to SemVer.

On the subject of performance and benchmarks, a lot of effort in the v3.4 release has been expended to ensure that we offer a reasonable alternative to the sub-1ms times of Elixir-based solutions constantly shoved in our faces as "proof" that we are, relatively speaking, a toy alternative. I've personally refactored a huge amount of client code just to make sure that the logged Reflex durations are ms-accurate and honest (eg. they include the time spend running morphdom or other DOM mutations).

I threw up a quick middleware performance benchmark: https://github.com/leastbad/stimulus_reflex_harness/tree/middleware

Here are 19 Reflexes' worth of Nothing morphs executing with middleware turned off (left) and on (right):

Off | On
--- | --
9 | 16
9 | 11
9 | 12
10 | 13
8 | 13
9 | 24
9 | 10
9 | 7
7 | 8
13 | 9
8 | 9
9 | 10
8 | 14
10 | 13
7 | 15
11 | 12
7 | 12
9 | 13
7 | 20

Granted, we are talking about milliseconds - but hope that you can respect that **we've worked very hard to achieve those numbers**. It's not some after-thought. We are trying to signal, loud and clear, that there's no meaningful advantage to using an Elixir alternative but that gets a LOT harder when you start getting into 20ms territory.

Meanwhile, the initializer itself is getting a ton of attention in the v3.4 release. Indeed, it's one of the stars of the show. I have nothing but confidence that the people who need to turn on a boolean value will be able to find it.

marcoroth:
How do you feel about when we just enable the middleware processing on `page` morphs per default? 
So `selector` and `nothing` morphs aren't affected by this change and the regular `page` will work with the least surprise.

leastbad:
Is that actually a less-confusing situation? I honestly don't know.

At this point, I'm going to leave the question of whether this needs to be done (remember, it would require a major version bump) to @hopsoft.

marcoroth:
> Is that actually a less-confusing situation? I honestly don't know.

Page morphs are the only morphs which also execute the controller actions so it would make sense in that way. 
But I just noticed that we don't know the type of the reflex yet at this location. So It anyway wouldn't have worked.

hopsoft:
@excid3 and I had a good conversation about this PR and he brought up some concerns that we should work to address. Here are some notes from that convo.

1. We need good defaults to simplify onboarding new folks but also need to balance that goal with performance and the risk of breaking existing apps
1. A reflex controller/action invocation over the socket is similar to but NOT the same as a traditional HTTP request. We want to avoid recreating HTTP over sockets because it's a slippery slope and will likely lead to even more confusion for users
1. Performance considerations likely warrant having the middleware be opt-in... certainly for `selector` and `nothing` morphs
1. An explicit middleware list is probably preferred as reflexes don't need things like `ActiveRecord::Migration::CheckPending` etc...

marcoroth:
Okay, I rewrote the PR to take advantage of `ActionDispatch::MiddlewareStack`. The syntax for registering a middleware in StimulusReflex is therefore exactly the same as in Rails. So all methods and the optional arguments are supported. Have a look at the updated PR description and let me what you think.

hopsoft:
This looks great @marcoroth and I think it's the right direction for the 3.4 release. **Thank you** for your efforts on this.  _We'll let the community debate setting up different generator defaults and/or always processing certain middleware in the next release._ 

One thing that we need to be clear about in the docs is that a `Reflex` is not a Rack application and that the optional middleware processing is only for preparing the `env` that the default `Reflex` behavior (page morph) uses to help re-render the current page. /cc @leastbad

excid3:
What if we take the Rails middleware stack and split it at some common place and use after thing after that by default? 

For example, everything after `Rack::TempfileReaper` would be none, or very few middlewares, but should catch a lot of custom middlewares out of the box.

hopsoft:
I like the idea of having a convenient mechanic to opt-in to the most common middleware likely to cause problems, but while also retaining the flexibility to register individual middleware. Perhaps an option like:

```ruby
  # register all middleware that might be needed to properly setup 
  # the environment used by the default behavior of page morph
  config.register_common_middleware = true

  # register middleware not in the predefined set above
  config.middleware.use MySpecialMiddleware
end
```

I'm hesitant to make this an implicit default that auto runs for the 3.4 release, but I'd be ok adding something like that to the generator for 3.4.

_My naming scheme here isn't great. Would love to see something better proposed._

---

Of course we really need to emphasize to our users that the middleware is only used to setup the `env` used to create the `request` and `controller` instances available to the reflex... and that SR is **not a Rack app** itself.

hopsoft:
Planning on merging this later today "as is" unless there are objections from anyone.

hopsoft (approved):

PR #403, opened 2020-12-09
Check package version from yarn.lock if node_modules folder is not av…

### LINKED ISSUE #402: Sanity checker will fail if node_modules folder isn't present
# Bug Report

## Describe the bug

The sanity checker relies on node_modules to verify that the stimulus_reflex npm package is installed. The node_modules folder isn't guaranteed to exist in production, because a user could pre-compile their frontend assets to public/ and never push the node_modules folder to the server (Or run yarn install on the server at all).

## To Reproduce

Run `assets:precompile`, delete the node_modules folder, and then run rails in production mode.

## Expected behavior

I would expect the sanity checker to either check the package.json, package-lock.json, or yarn.lock to determine the stimulus_reflex version, rather than checking the node_modules folder.

The actual code in question is here: https://github.com/hopsoft/stimulus_reflex/blob/ede440ca931943914108f217c54b7eb300c4ab35/lib/stimulus_reflex/sanity_checker.rb#L97

### StimulusReflex

- Gem: 3.4.0.pre8
- Node package: 3.4.0-pre8

### External tools

- Ruby: 2.7.2
- Rails: 6.0.3.4
- Node: 15.0.1

RolandStuder:
Fixes #402

For production deploys the sanity check fails, because it wants to check against the node_modules folder, which may not be present if assets are precompiled. This patch falls back to use the yarn.lock file to detect the package version.

### DIFF
--- a/lib/stimulus_reflex/sanity_checker.rb
+++ b/lib/stimulus_reflex/sanity_checker.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class StimulusReflex::SanityChecker
+  NODE_VERSION_FORMAT = /(\d+\.\d+\.\d+.*):/
   JSON_VERSION_FORMAT = /(\d+\.\d+\.\d+.*)"/
 
   class << self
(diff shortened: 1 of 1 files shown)


joshleblanc:
This looks fine. 

Strictly speaking, though yarn is the rails default, users could well be using NPM, in which case we should be checking package-lock.json.  Probably just deal with that if it comes up though.

RolandStuder:
Yeah, kinda surprising nobody has complained about that yet. SR just assumes you use yarn, also uses yarn to install the package during the install rake task.

leastbad:
Rails uses yarn, which is the why if not the should.

hopsoft (approved):

PR #404, opened 2020-12-09
Don't run sanity checker in production

joshleblanc:
# Enhancement

## Description

This is an anxilliary PR to #403 regarding #402 

I don't think the sanity checker should be running in production at all, and that would be consistent with the sanity checks that Rails and Webpacker perform (migrations, npm dependency checks). I think it's safe to assume that if a user has progressed to production, their code should already be passing the sanity checks.

## Checklist

### DIFF
--- a/lib/stimulus_reflex.rb
+++ b/lib/stimulus_reflex.rb
@@ -24,7 +24,7 @@ require "stimulus_reflex/logger"
 module StimulusReflex
   class Engine < Rails::Engine
     initializer "stimulus_reflex.sanity_check" do
-      SanityChecker.check!
+      SanityChecker.check! unless Rails.env.production?
     end
   end
 end


marcoroth:
Maybe we shouldn't check for the different versions but the caching check could also make sense in production, see: https://discord.com/channels/629472241427415060/733725826411135107/781217090526773280

RolandStuder:
I think we should keep the checks even in production. If you don't want them, you can say so in an initializer. Why should the checks not run for production, where they actually might be most important? Though it's true that we started this checker as a tool for developer happiness and to reduce confusion. Not sure.

joshleblanc:
I'm just throwing this out there as a proposal. I don't have strong feelings, especially since you can skip them with the ignore setting

RolandStuder:
But interesting I was not aware of this: 
> that would be consistent  with the sanity checks that Rails and Webpacker perform (migrations, npm dependency checks)

hopsoft (approved):
I think aiming for consistent behavior with other Rails defaults is the correct path on this.

PR #407, opened 2020-12-14
Update authentication.md

leoplct:
In case you have multiple Devise User model (eg. Admin and User) you will need to specify :user otherwise is nil.
Reference: https://stackoverflow.com/questions/43258458/envwarden-not-working-with-rails-5

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -97,7 +97,7 @@ module ApplicationCable
     protected
 
     def find_verified_user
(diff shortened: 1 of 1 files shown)


leastbad:
Hi @leoplct, as noted in the PR guidelines, we cannot accept documentation patches. The correct way to update the docs is to go to the #docs channel on Discord and leave a message.

Thanks for the tip, regardless. I'll do my best to work it in.

PR #414, opened 2020-12-22
Fix issue where element may no longer contain reflexData

### LINKED ISSUE #413: Lifecycle events not being issued correctly
# Bug Report
This issue is half feature proposal, half bug report. 

## Describe the bug
Assume that we have the following code, and when clicking on the element we issue a full page morph. 

```
<td
  data-reflex="click->CalendarReflex#new_calendar_event"
  id="Day{{ date|date:'Ymd' }}"
  data-date='{{ date|date:'c' }}'>
</td>
```

When you click on that element the standard stimulus reflex controller will pass everything correctly to the backend. However none of the lifecycle methods will work. 

To get the `beforeNewCalendarEvent` working correctly you would need to add `data-controller="calendar" which points to the controller that contains the method. 

If have the data-controller included `afterNewCalendarEvent` will work the first time, but will stop working if you continue clicking. The reason for this is that `element.reflexData` for some reason no longer contains any data. 

The proposal is that you should never need to include `data-controller` for lifecycle methods to work. If that is a must and you for some reason forgot to include `data-controller` lifecycle methods will fail silently which is no bueno. 

## Expected / desired behavior
Given that we know the method we are looking for (in this case `afterNewCalendarEvent` as an example. We should be able to find that method among the registered controllers and then execute that.

Assuming that there are two methods with the same name we would issue a warning that the lifecycle method won't be executed because we couldn't identify it properly.  

(which may be problematic if you work with inheritance). 

## Additional info
This is the data that is being sent to front-end. 

```
{
    "identifier": "{\\"channel\\":\\"StimulusReflex::Channel\\"}",
    "type": "message",
    "cableReady": true,
    "operations": {
        "morph": [
            {
                "selector": "body",
                "html": "[..snip...]",
                "childrenOnly": true,
                "permanentAttributeName": "data-reflex-permanent",
                "stimulusReflex": {
                    "target": "CalendarReflex#new_calendar_event",
                    "args": [],
                    "url": "http://localhost:8000/calendar/",
                    "attrs": {
                        "class": "fri",
                        "data-reflex": "click->CalendarReflex#new_calendar_event",
                        // data-controller is stimulus-reflex if data-controller is not defined
                        "data-controller": "calendar",
                        "id": "Day20201204",
                        "data-add-class-id": "backdrop",
                        "data-add-class-classes": "modal-backdrop fade show",
                        "data-date": "2020-12-04",
                        "data-action": "click->calendar#__perform",
                        "checked": false,
                        "selected": false,
                        "tagName": "TD"
                    },
                    "dataset": {
                        "data-reflex": "click->CalendarReflex#new_calendar_event",
                        // data-controller is stimulus-reflex if data-controller is not defined
                        "data-controller": "calendar",
                        "data-add-class-id": "backdrop",
                        "data-add-class-classes": "modal-backdrop fade show",
                        "data-date": "2020-12-04",
                        "data-action": "click->calendar#__perform"
                    },
                    "selectors": [],
                    "reflexId": "286a45a6-1d84-43b2-8c35-80c85040e511",
                    "resolveLate": false,
                    "xpath": "//*[@id=\'Day20201204\']",
                    "cXpath": "//*[@id=\'Day20201204\']",
                    // reflexController is stimulus-reflex if data-controller is not defined
                    "reflexController": "calendar",
                    "permanentAttributeName": "data-reflex-permanent",
                    "formData": "",
                    "identifier": "{\\"channel\\":\\"StimulusReflex::Channel\\"}"
                }
            }
        ]
    }
}
```

### StimulusReflex
- Node package: 3.4.0

### External tools
Django

### Browser
Firefox / Chorme.

jonathan-s:
Bug fix
@marcoroth Since you touched these lines last I think it would be appropriate if you could review this. 

## Description
Fixes #413 

In the lifecycle method element.reflexData sometimes only contain an empty object. Since that data is already available elsewhere in `reflexes` we can use the data from there to circumvent the issue. 

This means that when a controller is defined in html as `data-controller="mycontroller"` lifecycle methods will work even for my edge case where it doesn't work. 

## Why should this be added
This seems like a more reliable way of getting the reflex data. 

## Checklist

### DIFF
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -102,6 +102,9 @@ export const extractDataAttributes = element => {
   return attrs
 }
 
+// A reference to the Stimulus application registered with: StimulusReflex.initialize
+export let stimulusApplication
+
 // Finds an element based on the passed represention of the DOM element's attributes.
 //
 // NOTE: This is the same set of attributes extrated via extractElementAttributes and forwarded to the server side reflex.
--- a/javascript/lifecycle.js
+++ b/javascript/lifecycle.js
@@ -1,5 +1,9 @@
 import { camelize } from './utils'
-import { findElement, extractElementAttributes } from './attributes'
+import {
+  findElement,
+  extractElementAttributes,
+  stimulusApplication
+} from './attributes'
 import Debug from './debug'
 
 // Invokes a lifecycle method on a StimulusReflex controller.
@@ -15,7 +19,18 @@ import Debug from './debug'
 // - element - the element that triggered the reflex (not necessarily the Stimulus controller's element)
 //
 const invokeLifecycleMethod = (stage, element, reflexId) => {
-  if (!element || !element.reflexData[reflexId]) return
+  if (!element || !reflexes[reflexId]) return
+
+  if (Object.entries(element.reflexData).length === 0) {
+    element.reflexData[reflexId] = reflexes[reflexId].promise.data
+    let controllerName = reflexes[reflexId].promise.data.reflexController
+    element.reflexController[
+      reflexId
+    ] = stimulusApplication.getControllerForElementAndIdentifier(
+      element,
+      controllerName
+    )
+  }
 
   const controller = element.reflexController[reflexId]
   const reflex = element.reflexData[reflexId].target
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -12,15 +12,14 @@ import {
   attributeValues,
   extractElementAttributes,
   extractElementDataset,
-  findElement
+  findElement,
+  // A reference to the Stimulus application registered with: StimulusReflex.initialize
+  stimulusApplication
 } from './attributes'
 import { extractReflexName, elementToXPath, xPathToElement } from './utils'
 
 // A lambda that does nothing. Very zen; we are made of stars
 const NOOP = () => {}
-
-// A reference to the Stimulus application registered with: StimulusReflex.initialize
-let stimulusApplication
 
 // A reference to the ActionCable consumer registered with: StimulusReflex.initialize or getConsumer
 let actionCableConsumer


marcoroth:
Hey @jonathan-s, thanks for your PR!

I've dug down a little bit to understand where the problem lies. While it seemed like a good idea to "restore" the `reflexData` and `reflexController` in your the `invokeLifecycleMethod` function I think the problem lies somewhere different.

Serious question though, does this PR as is fixes your issue? I would be highly confused if it does! 
I had some reference issues with the `stimulusApplication` you changed. After fixing that locally the reflexes started to work again, I don't know what you've done there 😅 

 To provocate/simulate your issue I did this:

```javascript
const invokeLifecycleMethod = (stage, element, reflexId) => {
  if (!element || !reflexes[reflexId]) return

  element.reflexData = {}; // <---
  ```
Otherwise it wouldn't enter the `Object.entries(element.reflexData).length === 0` condition. But then the `getControllerForElementAndIdentifier` couldn't find a controller which also leads to an undefined `reflexController`. So if this works on your end I think there is something _completely_ off.

But then I realized that the issue has to be somewhere else. If you log the stages in `invokeLifecycleMethod()` you notice that it just enters the `before` callback if you invoke a reflex:

```javascript
const invokeLifecycleMethod = (stage, element, reflexId) => {
  console.log(stage)
  if (!element || !reflexes[reflexId]) return
  // ...
  
// just prints 'before' and nothing else
```

So it doesn't even receive the other life-cycle events, which would explain why your life-cycle methods aren't executed.

I think your current fix wouldn't help in this case, because it doesn't even enter the `invokeLifecycleMethod` and with that your fix won't even be executed.

To make progress on this issue: You said you have a MVCE in Django for this? Could your share this MVCE with me so I can have a look on it and to see where something could go wrong. Thanks you!

marcoroth:
I just opened https://github.com/jonathan-s/stimulus_reflex/pull/2 so that we are on the same page.

jonathan-s:
@marcoroth Thank you for your patience and for having a look at this. I've had a second look at the MVCE and it looks like I can no longer replicate it! Which also means that I found the root cause for it. 

The root cause happened to be a Turbolink that also issued a get request. After fixing the issue with the above code I noticed this turbolink, since it was causing other issues as well and removed it. With that turbolink thing removed it is working as expected. 

So apologies for this. 

On a related note, I still think it would be nice to not need to mention data-controller. Another idea for this would be to have it working by naming convention, if you have a `CalendarReflex` it would try to find a `calendar` as a controller and execute the lifecycle methods there. 

So since this PR doesn't solve anything I'm going to close it.

PR #416, opened 2021-01-03
Bug fix: Updating changelog generator to push to master, not main

MikeRogers0:
## Description

The default branch pushed to in `ad-m/github-push-action` was changed to be `main` instead of `master` ( https://github.com/ad-m/github-push-action/pull/69 ), this means the changelog would be pushed to `main`. 

I updated the changelog generator to match what I used on my repo ( https://github.com/bt-rb/bridgetown-minify-html/blob/master/.github/workflows/changelog.yml ) which originally I copied from here.

The new version explicitly pushes to the branch the GitHub action is running on & adds a few "nice to haves" for anyone who wants to copy this Action.

## Why should this be added

While I totally think using `main` as the default branch is cool, I imagine you don't want to get caught out next time you merge a fix.

## Checklist

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,6 +1,9 @@
 name: Changelog
 
 on:
+  workflow_dispatch:
+  release:
+    types: [created]
   push:
     branches:
       - master
@@ -12,26 +15,27 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[nodoc]')"
     steps:
     - uses: actions/checkout@master
-    - name: Set up Ruby 2.6
-      uses: actions/setup-ruby@v1
+    - name: Set up Ruby 3.0
+      uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6.x
-    - uses: actions/cache@v1.1.2
+        ruby-version: 3.0
+    - uses: actions/cache@v2
       with:
         path: vendor/bundle
-        key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
+        key: ${{ runner.os }}-changelog-gem-${{ hashFiles('**/Gemfile.lock') }}
         restore-keys: |
-          ${{ runner.os }}-gem-
+          ${{ runner.os }}-changelog-gem-
     - name: Create local changes
       run: |
         gem install github_changelog_generator
-        github_changelog_generator -u hopsoft -p stimulus_reflex --token ${{ secrets.GITHUB_TOKEN }} --exclude-labels duplicate,question,invalid,wontfix,nodoc
+        github_changelog_generator -u ${{ github.repository_owner }} -p ${{ github.event.repository.name }} --token ${{ secrets.GITHUB_TOKEN }} --exclude-labels duplicate,question,invalid,wontfix,nodoc
     - name: Commit files
       run: |
         git config --local user.email "github-actions@example.com"
         git config --local user.name "GitHub Actions"
-        git commit -m "[nodoc] update changelog" -a
+        git commit -am "[nodoc] Update Changelog" || echo "No changes to commit"
     - name: Push changes
       uses: ad-m/github-push-action@master
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
+        branch: ${{ github.ref }}


.github/workflows/changelog.yml
```diff
@@ -1,6 +1,9 @@
 name: Changelog
 
 on:
+  workflow_dispatch:
```
MikeRogers0, line 4: This means you can manually trigger this workflow, which sometimes I want to be able to do.

.github/workflows/changelog.yml
```diff
@@ -1,6 +1,9 @@
...
+  release:
+    types: [created]
```
MikeRogers0, line 6: I normally create releases on my GH projects, so kind of like it when the changelog auto updates for me.

.github/workflows/changelog.yml
```diff
@@ -12,26 +15,27 @@ jobs:
...
     - uses: actions/checkout@master
-    - name: Set up Ruby 2.6
-      uses: actions/setup-ruby@v1
+    - name: Set up Ruby 3.0
+      uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6.x
-    - uses: actions/cache@v1.1.2
+        ruby-version: 3.0
+    - uses: actions/cache@v2
```
MikeRogers0, line 22: Using the latest version of Ruby & tracking only major versions where I can.

.github/workflows/changelog.yml
```diff
@@ -12,26 +15,27 @@ jobs:
...
-        key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
+        key: ${{ runner.os }}-changelog-gem-${{ hashFiles('**/Gemfile.lock') }}
         restore-keys: |
-          ${{ runner.os }}-gem-
+          ${{ runner.os }}-changelog-gem-
     - name: Create local changes
       run: |
         gem install github_changelog_generator
-        github_changelog_generator -u hopsoft -p stimulus_reflex --token ${{ secrets.GITHUB_TOKEN }} --exclude-labels duplicate,question,invalid,wontfix,nodoc
+        github_changelog_generator -u ${{ github.repository_owner }} -p ${{ github.event.repository.name }} --token ${{ secrets.GITHUB_TOKEN }} --exclude-labels duplicate,question,invalid,wontfix,nodoc
```
MikeRogers0, line 31: This is the coolest change! If someone copies this, it'll work without them needing to set the `-u`/`-p` arguments.

.github/workflows/changelog.yml
```diff
@@ -12,26 +15,27 @@ jobs:
...
     - name: Commit files
       run: |
         git config --local user.email "github-actions@example.com"
         git config --local user.name "GitHub Actions"
-        git commit -m "[nodoc] update changelog" -a
+        git commit -am "[nodoc] Update Changelog" || echo "No changes to commit"
```
MikeRogers0, line 36: I had a repo where I pushed to master directly, then I had a red `x` next to my build because the changelog failed. This means if there are no changes to commit, it won't fail the job.

.github/workflows/changelog.yml
```diff
@@ -12,26 +15,27 @@ jobs:
...
     - name: Push changes
       uses: ad-m/github-push-action@master
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
+        branch: ${{ github.ref }}
```
MikeRogers0, line 41: This is the main change :) It matches the current branch we're running on, so when run on `master` it'll push to master. However if we change to `main` it'll still work without any work required.

leastbad:
Thanks for doing this, Mike! It looks good to me but I'm hoping @andrewmcodes and/or @marcoroth can sign off because infrastructure is like legalese to me. I'm terrible!

andrewmcodes (approved):
These are correct changes. Great addition @MikeRogers0!

I know I originally added this but I've since found much better ways to handle this. Not sure it's worth the effort unless others think it is cc @leastbad

andrewmcodes:
Also just to serve as encouragement for future viewers of this, adding comments to your code like Mike does above is an awesome way to add context and discussion points to your change for reviewers! Love to see it ❤️

PR #417, opened 2021-01-04
controller element holds Reflex metadata

leastbad:

Enhancement

## Description

This PR proposes to shift the element which holds Reflex metadata from the element which initiated the Reflex to the element which holds the Reflex controller. While it's frequently true that the element which initiated the Reflex is also holding the controller, it's also quite common for people to pass an element to `stimulate` or put a Stimulus controller on an ancestor of the initiating element so that the developer can make use of SR method callbacks.

## Why should this be added

We frequently see people morphing initiating elements out of the DOM, but still seeing tons of scary warnings about missing elements and failed life-cycle stages. While we can't prevent someone from blowing away their whole DOM, this should fix 90% of the complaints people have about those warnings (and fix callbacks for those Reflexes).

## Checklist

### DIFF
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -41,7 +41,6 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
           client_attributes: {
             reflex_id: data["reflexId"],
             xpath: data["xpath"],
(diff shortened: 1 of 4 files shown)


marcoroth (approved):

leastbad:
Thanks @marcoroth for the code review. Thanks @memoht for the inspiration.

PR #421, opened 2021-01-08
Update to work with mutatable CR config

hopsoft:
# Enhancement

Ensure that StimulusReflex works with custom user-added CableReady operations.

Designed to work with: https://github.com/hopsoft/cable_ready/pull/96

## Checklist

### DIFF
--- a/lib/stimulus_reflex/cable_ready_channels.rb
+++ b/lib/stimulus_reflex/cable_ready_channels.rb
@@ -18,6 +18,7 @@ module StimulusReflex
 
     def method_missing(name, *args)
       return stimulus_reflex_channel.public_send(name, *args) if stimulus_reflex_channel.respond_to?(name)
+      super
     end
 
     def respond_to_missing?(name, include_all)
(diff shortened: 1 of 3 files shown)


leastbad:
This is ready to go.

leastbad:
1. What will the status of the `dom_id` method be? Can we go back to delegating this to `CableReady::Broadcaster` instead of redefining it in `reflex.rb`?
2. Is there a way that we can alert the developer that they have included `CableReady::Broadcaster` in their Reflex class? Even boot them out of their server process? It might seem heavy-handed but it is still generating a significant number of support inquiries.

hopsoft:
1. `CableReady::Broadcaster#dom_id` is defined on a module that is not used by StimulusReflex. The proper solution here will be to move the definition of that helper method somewhere else in the CR library so SR can more easily lean on that implementation.
2. I tried a few ways of tackling this, but settled on putting it in the reflex constructor as it guarantees we're checking the assembled runtime ancestor tree. It provides some nice guidance for anyone still pulling `CableReady::Broadcaster` into their reflexes.

Issue #425, opened 2021-01-10
Multiple text inputs with the same name lose value in Reflex


jonsgreen:
# Bug Report

This has been discussed, reported and confirmed in [this discord message thread](https://discord.com/channels/629472241427415060/683317077423292582/797329509564809216)

## Describe the bug

When triggering a reflex on a text input from a page that has multiple inputs with the same name attribute value the element's value is always reported as blank in the reflex action. 

It has been concluded that [these lines](https://github.com/hopsoft/stimulus_reflex/blob/2dcd244bd45084aaf47a4b5627f5b486591ea22f/javascript/attributes.js#L4-L7) are the source of the problem and should be limited to just inputs that are checkboxes or radio buttons. 

## To Reproduce

Using [this MVCE](https://github.com/jonsgreen/stimulus_reflex_input_bug) on the root page you can observe how changing the input text that has the same name clears the value of the message comment in MessageReflex#update. 

## Expected behavior

The element value in the reflex should be whatever the input value was when the reflex was triggered. In the MVCE the message comment value should get updated using the input with the same name as other inputs on the page. 

### StimulusReflex

- Gem: 3.4.0
- Node package: 3.4.0

### External tools

- Ruby: 2.7.2
- Rails: 6.1
- Node: [e.g. 12.4.0]

### Browser

- Browser: Chrome
- Version: 87.0.4280.141

PR #426, opened 2021-01-10
Fixes bug where multiple inputs with the same name lost element value

jonsgreen:
# Bugfix

## Description

- allows only checkbox and radio inputs to be multipleInstances
- adds failing test for multiple text inputs with same name attribute

Fixes https://github.com/hopsoft/stimulus_reflex/issues/425

## Why should this be added

To prevent confusion for future users of the gem particularly in cases where there are multiple nested forms using rails helpers that would generate the same name for inputs on the same page. 

## Checklist

### DIFF
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -65,11 +65,6 @@ export const extractElementAttributes = element => {
     attrs.value = collectedOptions.join(',')
   } else {
     attrs.value = element.value
-    if (element.tagName.match(/select/i)) {
-      if (element.selectedIndex > -1) {
-        attrs.value = element.options[element.selectedIndex].value
-      }
-    }
   }
   return attrs
 }
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -1,11 +1,17 @@
 import { defaultSchema } from './schema'
 import Debug from './debug'
 
-const multipleInstances = element =>
-  document.querySelectorAll(
-    `input[type="${element.type}"][name="${element.name}"]`
-  ).length > 1
-
+const multipleInstances = element => {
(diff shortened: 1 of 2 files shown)


javascript/attributes.js
```diff
@@ -1,11 +1,17 @@
...
-
+const multipleInstances = element => {
+  if (['checkbox', 'radio'].includes(element.type)) {
+    return (
+      document.querySelectorAll(
+        `input[type="${element.type}"][name="${element.name}"]`
+      ).length > 1
+    )
+  } else {
+    return false
```
leastbad, line 12: I suspect you can just get rid of the whole `else {}` and just return false here. False is a very reasonable default return for this function.

javascript/attributes.js
```diff
@@ -59,11 +65,6 @@ export const extractElementAttributes = element => {
     attrs.value = collectedOptions.join(',')
   } else {
     attrs.value = element.value
-    if (element.tagName.match(/select/i)) {
```
leastbad, line 62: @julianrubisch I'm >99% sure that this code was intended to be deleted when you added support for `values` but it forgot to die.

javascript/attributes.js
jonsgreen: Good suggestion, thanks.

jonsgreen:
@leastbad I am guessing this is known but there was and still is a failing test for me at `StimulusReflex::ReflexTest#test_dom_id [/Users/jonathangreenberg/dev/stimulus_reflex/test/reflex_test.rb:30]`

leastbad:
That's weird - are you able to use `dom_id` in one of your Reflex actions?

PR #427, opened 2021-01-10
Extract callbacks to module

julianrubisch:
# Refactoring

## Description

Very simplistic refactoring to extract `callbacks` to a module, giving the `Reflex` class a bit more space and create a dedicated space for callback behavior.

Please advise on style, I simply wrapped the singleton class definition in a `class_eval` block, let me know if there's a more idiomatic way!

## Checklist


lib/stimulus_reflex/callbacks.rb
```diff
@@ -0,0 +1,46 @@
+module StimulusReflex
+  module Callbacks
+    def self.included(klass)
```
hopsoft, line 3: Considering that we have Rails as a dependency, we may want to look at using `ActiveSupport::Concern` and benefit from the prettier semantics. Thoughts? https://api.rubyonrails.org/classes/ActiveSupport/Concern.html

julianrubisch: Will happily rewrite if this is the direction 👍🏻

julianrubisch: made the change 👍

andrewmcodes (approved):
Looks like a nice abstraction with the benefit of the AS enhancement. This code is more readable barring objections I say we get it in. 

However now that we have created this abstraction, would it be easier to add some test coverage? That would seem like a win to me. 

@hopsoft @leastbad

leastbad:
I agree that this would be an improvement. The only reason I haven't merged it is because I don't want to short-circuit any initiatives that might be underway to add some test coverage, as Andrew suggested.

PR #431, opened 2021-01-15
life-cycle callbacks return correct element reference

leastbad:

Bug fix

## Description

Changes to how SR Controller Element stores metadata in #417, while positive in intention and generally well-executed, did not cover an important use-case, which is that the `element` parameter in life-cycle callbacks needs to point to the Reflex element, not the SR Controller Element. (You can always obtain a reference to the Controller Element via `this.element`.)

Fixing this was unfortunately non-trivial, and required a major refactoring and testing effort, specifically because all life-cycle functions now require a reference to the Reflex Controller Element in addition to the previous arguments.

One major upside to this change is that the `dispatchLifecycleEvent` function has been greatly simplified while still keeping the critical functions. It can still tell you if the Controller Element has been removed from the DOM, and it will even warn you if the SR Controller has been replaced with a new instance; the Reflex can continue, but callbacks + events are lost. I think that this is pretty slick and highly forgiving.

It's also interesting/historical that the `findElements` function has been fully removed. There's simply no need for it, since we can use XPath queries to locate the Reflex Element and Controller Element.

**A heartfelt thank-you to @jonsgreen for catching this issue. We were on track to an unfortunate and likely embarrassing point release, and this is not the best time for that.**

Fixes #417

## Why should this be added

Without this fix, all currently deployed apps which rely on life-cycle callbacks will break if they attempt to operate on the `element` parameter.

## Checklist


javascript/lifecycle.js
```diff
@@ -125,56 +186,51 @@ document.addEventListener(
...
+    reflexId,
+    element: reflexElement
+  }
 
-  element.dispatchEvent(
+  controllerElement.dispatchEvent(
     new CustomEvent(event, { bubbles: true, cancelable: false, detail })
   )
-  if (window.jQuery) window.jQuery(element).trigger(event, detail)
+  if (window.jQuery) window.jQuery(controllerElement).trigger(event, detail)
```
hopsoft, line 235: Shouldn't we dispatch lifecycle events on the `reflexElement` given that the event will bubble up to the controller if someone wants to handle it there?

leastbad: Yes, that's an exceptionally smart call.

PR #432, opened 2021-01-15
more explicit quick_start doc increment example

adipasquale:
Hi and thank you for a super nice approach. 

I'm just learning discovering Stimulus Reflex, and while going through the quick start example, I found it slightly complicated to understand. I had a hard time understanding how instance variables set inside Reflexes would become available in the view, and how it interacted with controller set instance variables.

I think part of this blur came from the fact that your example currently relies on the fact that `nil.to_i == 0` which is not super obvious or expected. I'm suggesting a slightly more explicit code that explicitly tests for the presence of the instance variable in the view and then does something. I hope it may help future readers to understand that the variable will only be set when coming from a reflex, and not from a regular page load.  

I had to watch the nice and short screencast to actually understand this. Maybe it'd be nice to actually a small piece of doc about this interaction with instance variable set in controllers, but I'll leave that up to you!

## Description

Fixes # (issue)

## Why should this be added

Explain value.

## Checklist

- [ ] This is not a documentation update

### DIFF
--- a/docs/hello-world/quickstart.md
+++ b/docs/hello-world/quickstart.md
@@ -42,7 +42,7 @@ This example updates the page with the latest count when the link is clicked:
   data-reflex="click->Counter#increment"
   data-step="1" 
   data-count="<%= @count.to_i %>"
->Increment <%= @count.to_i %></a>
+>Increment <%= @count.present? ? @count.to_i : "from 0" %></a>
 ```
 {% endcode %}
 


adipasquale:
haha okay sorry I just saw the `this is not a doc update` check after opening the PR, I'll close it right now!

leastbad:
No worries @adipasquale! The concern is that our documentation system is just broken enough that accepting PRs makes any WIP documentation efforts break in an unfortunate way.

All input regarding the way the docs is structured is extremely appreciated - these words are quite literally for you. There's no question that there's room for improvement.

One resource you might find really helpful is [this chart](https://app.lucidchart.com/documents/view/e83d2cac-d2b1-4a05-8a2f-d55ea5e40bc9/0_0) showing the code paths of the different kinds of Reflexes. Page Morphs, which are the default and re-run the controller, are complex until they suddenly are not! I believe that your effort will soon be rewarded.

The other thing I cannot recommend enough is that you join and ask lots of questions on the Discord server. There's folks there 24/7 who will help get you over the hump.

adipasquale:
thank you for such a helpful answer @leastbad 🙇

PR #434, opened 2021-01-30
version check

### LINKED ISSUE #433: SR+CR version checks on startup in development
Desktop software gets version updates. Hardware gets firmware updates. Web apps just need to be refreshed.

We get a ton of support requests for people who don't check to see if their version is out of date.

What if StimulusReflex and CableReady each checked their version against a file hosted on GH when the app starts up, as an extension of the sanity check process?

Obviously lots of implementation specifics to chatter about, but I'm posting this to take everyone's temperature on the idea.

leastbad:

Enhancement

## Description

During Rails startup, phone home to Github and compare local version against `LATEST`.

Adds new `config.on_new_version_available` flag to SR initializer which can be :exit, :warn or :ignore.

Defaults to :ignore (no connection attempted). Test only runs in development mode.

Network call is set to timeout 1s connect, 1s read (2s total).

Test will emit a message about skipping checks if running a pre-release build.

Test will not run if sanity checks are set to :ignore, regardless of `on_new_version_available` setting.

Updating `LATEST` will have to become part of the version deployment process.

Fixes #433

## Why should this be added

This should be recommended to users struggling with keeping versions up to date. Goal is to reduce support version.

## Checklist


marcoroth:
Awesome, I love this addition!

I also like that the check is disabled by default, so we are only requesting stuff in the background if the user "permits" and actually wants it.

The only thing I would reconsider is the use of `LATEST`. It's another file we have to maintain if we release a new version. Maybe it would make more sense to request the latest version over the GitHub API, because the information is already there.

Either via the git tag:
```bash
curl https://api.github.com/repos/hopsoft/stimulus_reflex/tags
```

or via the GitHub releases:
```bash
curl https://api.github.com/repos/hopsoft/stimulus_reflex/releases/latest
```

lib/stimulus_reflex/sanity_checker.rb
```diff
@@ -62,6 +64,31 @@ def check_javascript_package_version
...
+            Run `bundle install && yarn install` to complete the upgrade.
+
+          WARN
+          exit if StimulusReflex.config.on_new_version_available == :exit
+        end
+      rescue
+        puts "StimulusReflex #{StimulusReflex::VERSION} update check skipped: connection timeout"
+      end
+    else
+      puts "StimulusReflex #{StimulusReflex::VERSION} update check skipped: pre-release build"
```
RolandStuder, line 88: Found the branching here a bit hard to read, maybe do line #88 as a guard clause instead.

RolandStuder: Ah, no, I realize with the puts it a bit hard to do a guard clause.

PR #436, opened 2021-01-31
implicit dom_id for morph selectors

leastbad:

Enhancement

## Description

This proposes to allow `morph` to accept a resource in lieu of a CSS selector. Any resource will be converted to a string with `dom_id`.

Before:
```rb
morph dom_id(post), "I am a teapot"
```
After:
```rb
morph post, "I am a teapot"
```
Before:
```rb
morph {dom_id(post1) => render(post1), dom_id(post2) => render(post2)}
```
After:
```rb
morph {post1 => render(post1), post2 => render(post2)}
```

I've also made the `Reflex#dom_id` helper more powerful than its Rails counterpart, so that it now accepts ActiveRecord relations and arbitrary objects (that get converted to strings):
```rb
dom_id(post1) # "#post_1" (the current behaviour)
dom_id(Post.all) # "#posts"
dom_id("teapots", "search") # "#teapots_search"
```

That's all pretty cool, but with the above in place, we can dial it up.

```rb
morph User.first
```
This is equivalent to:
```rb
morph dom_id(User.first), render(User.first)
```

And, thanks to [a quirk in the way that morphdom works](https://docs.stimulusreflex.com/appendices/troubleshooting#different-element-type-altogether-who-cares-so-long-as-the-css-selector-matches), this is now possible:

```rb
morph Post.all
```

This is equivalent to:
```rb
morph dom_id(Post.all), "<div id=\"#{dom_id("posts")}\">#{ApplicationController.render(Post.all)}</div>"
```

morphdom DGAF so long as the outermost container element has the same `id` as the target. If you happen to be rendering a collection with a partial that lines up with render defaults, we can now infer everything but the collection itself.

Finally, I added a `wrap` method to `Reflex` that developers can use for more complex morphs:
```rb
wrap(render(posts, partial: "home/posts", locals: {pagy: pagy}), "#my_content_div")
```
```html
<div id="my_content_div">post1post2</div>
```

Thanks to @RolandStuder for the initial inspiration.

## Why should this be added

Developer happiness and the principle of least surprise.

## Checklist


lib/stimulus_reflex/reflex.rb
```diff
@@ -129,7 +129,13 @@ def params
     @_params ||= ActionController::Parameters.new(request.parameters)
   end
 
-  def dom_id(record_or_class, prefix = nil)
-    "#" + ActionView::RecordIdentifier.dom_id(record_or_class, prefix).to_s
+  def dom_id(record, prefix = nil)
+    return "#" + record.class.to_s.split("::")[0].pluralize.underscore if record.class < ActiveRecord::Relation
```
marcoroth, line 133: Maybe you could simplify this to `record.table_name`

leastbad: I was actually wondering about that! The reason I didn't was because I wondered if there could be issues if the model class specified a non-standard table name. Is that a thing?

I spent a long time looking at `record.methods - Object.methods` trying to spot something that would return exactly what I needed, and saw `table_name` but was concerned:

```rb
class Task < ActiveRecord::Base
  self.table_name = 'projects'
end	end
```

marcoroth: This is a thing, but `table_name` should reflect that if you overwrite the table name in the model.

leastbad: You're saying that even if I change my User model's underlying table to be backed by "people", that calling User.first.table_name will still return "user"?

Because what we need here is undoubtedly "user".

marcoroth: Ah right, I understood it the other way around.

Okay in that case, this could work:
```ruby
record.model.model_name.collection
# or
record.model.model_name.plural
```

leastbad: I don't know what I'd do without you, @marcoroth!

lib/stimulus_reflex/reflex.rb
```diff
@@ -129,7 +129,13 @@ def params
...
-  def dom_id(record_or_class, prefix = nil)
-    "#" + ActionView::RecordIdentifier.dom_id(record_or_class, prefix).to_s
+  def dom_id(record, prefix = nil)
+    return "#" + record.model_name.plural if record.is_a?(ActiveRecord::Relation)
+    return "#" + ActionView::RecordIdentifier.dom_id(record, prefix).to_s if record.is_a?(ActiveRecord::Base)
+    [record.to_s, prefix].compact.join("_")
+  end
+
+  def wrap(content, resource)
+    "<div id=\"#{dom_id(resource)}\">#{content}</div>"
```
marcoroth, line 139: Small change, but imo it somehow looks cleaner than interpolating variables into the string. 
```suggestion
    tag.div content, id: dom_id(resource)
```

(maybe needs an include for `ActionView::Helpers::TagHelper`)

leastbad: This was a great suggestion. Thanks!

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
```diff
@@ -6,9 +6,14 @@ def broadcast(_, data = {})
       morphs.each do |morph|
         selectors, html = morph
         updates = selectors.is_a?(Hash) ? selectors : Hash[selectors, html]
-        updates.each do |selector, html|
+        updates.each do |key, value|
+          updates[key] = reflex.render(key) if key.is_a?(ActiveRecord::Base) && value.nil?
+          updates[key] = reflex.wrap(reflex.render(key), key) if key.is_a?(ActiveRecord::Relation) && value.nil?
+        end.transform_keys { |key|
+          key.is_a?(ActiveRecord::Base) || key.is_a?(ActiveRecord::Relation) ? reflex.dom_id(key) : key
+        }.each do |selector, html|
```
hopsoft, line 14: We could eliminate a loop here by moving the `transform_keys` logic into the `each`.

```ruby
.each do |key, html|
  selector = key.is_a?(ActiveRecord::Base) || key.is_a?(ActiveRecord::Relation) ? reflex.dom_id(key) : key
  
  ...
```

lib/stimulus_reflex/reflex.rb
```diff
@@ -129,7 +130,13 @@ def params
     @_params ||= ActionController::Parameters.new(request.parameters)
   end
 
-  def dom_id(record_or_class, prefix = nil)
-    "#" + ActionView::RecordIdentifier.dom_id(record_or_class, prefix).to_s
+  def dom_id(record, prefix = nil)
+    return "#" + [record.model_name.plural, prefix].compact.join("_") if record.is_a?(ActiveRecord::Relation)
+    return "#" + ActionView::RecordIdentifier.dom_id(record, prefix).to_s if record.is_a?(ActiveRecord::Base)
+    "#" + [record.to_s, prefix].compact.join("_")
```
hopsoft, line 136: Shouldn't `prefix` be the first item in the array?

lib/stimulus_reflex/reflex.rb
```diff
@@ -129,7 +130,13 @@ def params
...
+  end
+
+  def wrap(content, resource)
```
hopsoft, line 139: Would love a code comment as to why wrap exists. i.e. to satisfy morphdom when working with lists

hopsoft (changes requested):
I wonder if there is any additional syntactic sugar we could include by leveraging [to_partial_path](https://api.rubyonrails.org/classes/ActiveModel/Conversion.html#method-i-to_partial_path). Too magical?

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
leastbad: Actually, I'm not sure that you can!

The whole reason I did this is that Ruby complains about adding keys during enumeration.

If you do it with `each`, you're essentially creating a new key and deleting the old one. `transform_keys` actually does what it suggests.

If you know of a way around this, I only learned all of this two days ago. I'm confident you could have a better answer. I just wanted you to know that this was an intentional thing.

leastbad:
I think that we're implicitly leveraging `to_partial_path` when we support the `morph @posts` syntax, even though it's not something we have to do directly.

What do you have in mind? Tell me like I'm not in your head! 🤓

lib/stimulus_reflex/reflex.rb
```diff
@@ -133,9 +133,13 @@ def params
   def dom_id(record, prefix = nil)
     return "#" + [record.model_name.plural, prefix].compact.join("_") if record.is_a?(ActiveRecord::Relation)
```
marcoroth, line 134: @leastbad you probably want to reverse the order also on L134

leastbad: 🤦

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
hopsoft: I haven't tested this yet, but I suspect that we can refactor broadcast into this.

```ruby
def broadcast(_, data = {})
  morphs.each do |morph|
    selectors, html = morph
    updates = selectors.is_a?(Hash) ? selectors : Hash[selectors, html]

    updates.each do |key, value|
      html = reflex.render(key) if key.is_a?(ActiveRecord::Base) && value.nil?
      html = reflex.wrap(reflex.render(key), key) if key.is_a?(ActiveRecord::Relation) && value.nil?
      fragment = Nokogiri::HTML.fragment(html.to_s)

      selector = key.is_a?(ActiveRecord::Base) || key.is_a?(ActiveRecord::Relation) ? reflex.dom_id(key) : key
      match = fragment.at_css(selector)

      if match.present?
        operations << [selector, :morph]
        cable_ready.morph(
          selector: selector,
          html: match.inner_html,
          children_only: true,
          permanent_attribute_name: permanent_attribute_name,
          stimulus_reflex: data.merge({
            morph: to_sym
          })
        )
      else
        operations << [selector, :inner_html]
        cable_ready.inner_html(
          selector: selector,
          html: fragment.to_html,
          stimulus_reflex: data.merge({
            morph: to_sym
          })
        )
      end
    end
  end

  cable_ready.broadcast
  morphs.clear
end
```

leastbad: There were a few small tweaks (eg what gets passed to the fragment has to fall back to an empty string, now) but this worked. I'm always thrilled when you see these optimizations.

leastbad:
I reworked the new `dom_id` method so that it was a case structure, but standardrb kicked my ass and turned it into an if/elsif and I'm no longer sure it's still cute.

Either way, I had to allow the `wrap` method to squash the `#` being prepended to prevent a dreaded double-hash. I also added a `sub` to the fall-through `dom_id` case to prevent people from accidentally sending an extra `#` character into the method.

All of this is open to iteration. 💯

lib/stimulus_reflex/reflex.rb
```diff
@@ -129,7 +130,21 @@ def params
...
 
-  def dom_id(record_or_class, prefix = nil)
-    "#" + ActionView::RecordIdentifier.dom_id(record_or_class, prefix).to_s
+  def dom_id(record, prefix = nil, hash: "#")
+    hash + if record.is_a?(ActiveRecord::Relation)
+      [prefix, record.model_name.plural].compact.join("_")
+    elsif record.is_a?(ActiveRecord::Base)
+      ActionView::RecordIdentifier.dom_id(record, prefix).to_s
+    else
+      [prefix, record.to_s.sub(/^#/, "")].compact.join("_")
```
hopsoft, line 139: I gotten into the habit of using `/\A/` vs `/^/` for matching the beginning of a string. 

- `^` Matches beginning of line
- `\A` Matches beginning of string

It probably doesn't matter here, but I almost never use `^` for this scenario.

Of course we could also use `String#squeeze`

```ruby
record.to_s.squeeze("#")
```

hopsoft (approved):

lib/stimulus_reflex/reflex.rb
leastbad: 🤦 would you believe I spent time looking at the String rdoc because I was *sure* there was a method that would do this, but I didn't connect the dots.

This is obviously superior.

PR #438, opened 2021-02-02
deprecation warnings (consumer, form serialization, isolation mode)

leastbad:

Enhancement

## Description

- gave `createSubscription` the ability to pull ActionCable consumer from the Stimulus application
- set the `initialize()` function to issue a deprecation warning if `consumer` was passed
- modified the install rake task to add `application.consumer = consumer` to index.js
- introduced the `data-reflex-serialize-form` attribute
- added code to support declaring a form should be serialized
- set the `stimulate()` method to issue a deprecation warning if a form is detected but `serializeForm` hasn't been set
- set the `initialize()` function to issue a deprecation warning if isolation mode is not enabled

## Why should this be added

This is the first polite step towards a full transition to `application.consumer` in SR v4. This will allow us to drop `consumer.js` from the project.

## Checklist


javascript/stimulus_reflex.js
```diff
@@ -480,6 +481,12 @@ const getReflexRoots = element => {
 const initialize = (application, initializeOptions = {}) => {
   const { controller, consumer, debug, params, isolate } = initializeOptions
   actionCableConsumer = consumer
+  setTimeout(() => {
+    if (Debug.enabled && consumer)
```
marcoroth, line 485: I'm wondering if we should introduce another option to silence deprecation warnings? I have the feeling that deprecation warnings are not just "debug logs". What do you think?

leastbad: How would we transmit this to the client? Or do you mean a client option?

Given that you need to have debug on, don't we want them to see these?

marcoroth: Yep a client option. `debug` is disabled by default so if you don't change anything you won't see the deprecation warnings.

I think an option like `deprecationWarnings` which is enabled by default so we can make sure the user really sees them. If he wants to ignore the warnings he can set `deprecationWarnings` to false. 

So it's not deepened on the `debug` option if the users sees the warnings or not.

leastbad: Would you go so far as to essentially duplicate the Debug mechanism? Or do you have something less intense in mind?

It seems like there are shortcuts we'd just regret if we push it down the line...

marcoroth: Either to duplicate the mechanism we use in Debug or that we refactor the Debug stuff into a more general "config" construct so we can use it for multiple options. I don't have something less intense is mind.

But maybe it goes too far idk.

leastbad: I think I'm going to just duplicate the Debug mechanism you built. It works amazingly well, and we can leave it in the library for future deprecations.

leastbad: Ready for review!

javascript/deprecate.js
```diff
@@ -0,0 +1,19 @@
+let deprecationWarnings = false
```
marcoroth, line 1: I would just set this to `true` by default so the users can actually see them

leastbad: Oh, right. That will mean that I should reconsider how line 507 works, too.

jonathan-s:
As `transports/action_cable` was started in #444, perhaps it would be a good idea to start thinking about the naming here. Ie instead of `consumer` we might name this `transport`. The naming would force us to consider a broader scope that is about to come.

leastbad:
I appreciate where you're coming from with this, but the `application.consumer` thing is both specific to ActionCable and its consumer.js and it's a pattern I've been pushing HARD for Stimulus + AC libraries to adopt in general for almost two years. It's the only way Stimulus packages don't have to individually deal with the fact that consumer.js might be in a different relative location on each host app.

Long/short: when we are working with AC, application.consumer will be the source of the AC consumer. It's much bigger than StimulusReflex.

javascript/deprecate.js
```diff
@@ -0,0 +1,19 @@
...
+  get disabled () {
+    return !deprecationWarnings
+  },
+  get value () {
+    return deprecationWarnings
+  },
+  set (value) {
+    deprecationWarnings = !!value
+  },
+  set debug (value) {
```
marcoroth, line 16: Small detail
```suggestion
  set deprecate (value) {
```

PR #439, opened 2021-02-03
remove unnecessary gate

leastbad:

"Enhancement"

## Description

Back in #45 (aka callbacks) we introduced a gate around the bottom-half of the main `stimulus_reflex.js` to make sure it was only executed once. The thing is, I cannot for the life of me see what purpose this serves. Presumably, with or without Turbolinks, this code is only going to be executed once per page - or else the `const initialize` immediately above it would already be complaining loudly, right?

This PR just removes the gate. There could be a very good reason to keep it! I just don't know what we were guarding against.

## Why should this be added

Code clarity

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -491,117 +491,108 @@ const initialize = (application, initializeOptions = {}) => {
   Debug.set(!!debug)
 }
 
-if (!document.stimulusReflexInitialized) {
-  document.stimulusReflexInitialized = true
-
-  window.addEventListener('load', () => {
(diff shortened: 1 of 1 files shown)


leastbad:
Note that the file view on this PR looks intimidating, but it's literally just removing lines **494-496**, **604**.

The changeset is the result of the indentation formatting moving everything two spaces to the left.

hopsoft (approved):
This should be safe considering that SR is typically initialized in `app/javascript/controllers/index.js` or somewhere similar.

PR #440, opened 2021-02-04
refactor declarative reflex observer

leastbad:

Bug fix

## Description

SR uses a MutationObserver to watch the DOM for any changes. This is unnecessary because we're just looking for updates to the `data-reflex` and `data-controller` attributes. Observing all attributes uses more resources and has un-cool side-effects for libraries that rapidly change attributes such as GSAP animation.

By moving the observer setup to the SR initialize (which only gets run once) we gain access to the Stimulus application schema, which can be used to provide an array of attributes to monitor.

**Question**: do we need to monitor `data-action`? For that matter, do we need to monitor `data-controller`? I think there's a good chance we *only* care about `data-reflex`.

**Question**: did you choose to use the `window` `load` event over `DOMContentLoaded` for a specific reason? I don't know if it's right or wrong, just curious for insight.

Fixes leastbad/radiolabel#3

## Why should this be added

radiolabel users are seeing their DOM scanned over 200 times every time there's a Reflex operation. Even `setupDeclarativeReflexes` will cause `setupDeclarativeReflexes` to be fired. This is over-zealous.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -489,20 +489,20 @@ const initialize = (application, initializeOptions = {}) => {
     controller || StimulusReflexController
   )
   Debug.set(!!debug)
+  const observer = new MutationObserver(setupDeclarativeReflexes)
+  observer.observe(document.documentElement, {
+    attributeFilter: [
+      stimulusApplication.schema.reflexAttribute,
+      stimulusApplication.schema.controllerAttribute,
+      stimulusApplication.schema.actionAttribute
+    ],
+    childList: true,
+    subtree: true
+  })
 }
 
 if (!document.stimulusReflexInitialized) {
   document.stimulusReflexInitialized = true
-
-  window.addEventListener('load', () => {
-    setupDeclarativeReflexes()
-    const observer = new MutationObserver(setupDeclarativeReflexes)
-    observer.observe(document.documentElement, {
-      attributes: true,
-      childList: true,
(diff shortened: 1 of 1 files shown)


hopsoft:
My understanding is that `window` `load` waits for all resources to load and process. I considered it the safer option just in case any scripts modified the DOM in a way that SR might care about... but that expectation/assumption could be incorrect or overzealous.

javascript/stimulus_reflex.js
```diff
@@ -489,21 +489,21 @@ const initialize = (application, initializeOptions = {}) => {
     controller || StimulusReflexController
   )
   Debug.set(!!debug)
+  const observer = new MutationObserver(setupDeclarativeReflexes)
+  observer.observe(document.documentElement, {
+    attributeFilter: [
+      stimulusApplication.schema.reflexAttribute,
+      stimulusApplication.schema.controllerAttribute,
+      stimulusApplication.schema.actionAttribute
+    ],
```
hopsoft, line 498: Monitoring `stimulusApplication.schema.reflexAttribute` should be sufficient.

hopsoft (changes requested):
This is a great optimization.

leastbad:
I'm sure that `load` continues to be just fine. Probably good to give everything a few ms to warm up.

Shall we squash and merge, then?

hopsoft (approved):

PR #444, opened 2021-02-05
refactor stimulus_reflex.js

leastbad:
While this PR will seem radical and a little intimidating at first glance, **no features have been added**. The PR file view is likely only useful to see how I split the project up; you really need to pull it down and test it out.

I was very intentional about making commits at meaningful junctures so I could "show my work" as the teachers said. I had my comprehensive CableReady test harness open and running during the process, always checking each step that everything seems to be working.

Isolation Mode is working fine. You'll see that it has been pulled into a file which can be conveniently removed for v4.

There are no more local variables. What used to be `window.reflexes` is now `reflexes` (eg. same API, different storage).

I started the process of treating ActionCable like an adapter. There is a new `javascript/transports` folder. Since I was committed to not changing functionality, there is still AC-specific code in the `register` function, but it is now self-evident where the offending code is and extracting it will not be difficult. I *did* remove the old `consumer.js`; it is almost certainly vestigal at this point; there's never been any documentation encouraging people to attach their consumer to `window`, and all tasks/generators pass `consumer` into the options of `initialize`. Most significantly, I've not seen or heard of anyone doing it in 18 months of daily support. Considering it was generating deprecation errors in Chrome (since it prods every member of `window`) and we have `application.consumer` support coming in, I figured that it was time to sunset that particular history.

I also took the opportunity to merge what used to be `extendStimulusController` into `register`. Whatever reason there used to be to keep them separate is no longer in play.

In conclusion, while the SR client is still involved and there's lots to be done to refactor `stimulate` into more manageable, testable chunks, I believe that this PR makes the codebase substantially easier to reason about. Not coincidentally, it makes it feel much more like a Stimulus controller instead of a jumble of functions that only two people can follow without frustration.

## Why should this be added

Code clarity and maintainability. Preparation for pluggable transport mechanisms.

## Checklist


javascript/stimulus_reflex.js
```diff
@@ -1,166 +1,75 @@
...
-// - stimulate
-// - __perform
+// controller - the Stimulus controller
+// options - [optional] configuration
 //
-const extendStimulusController = controller => {
+const register = (controller, options = {}) => {
+  const channel = 'StimulusReflex::Channel'
+  controller.StimulusReflex = { ...options, channel }
+  actionCable.createSubscription(controller)
```
jonathan-s, line 72: May I suggest that you modify register to this.
```
const register = (controller, options = {}, consumer = actionCable) => 
```
then you'd have 
```
consumer.createSubscription(controller)
```

javascript/reflexes.js
```diff
@@ -0,0 +1,207 @@
...
+import { XPathToElement, debounce, emitEvent } from './utils'
+import { allReflexControllers, findControllerByReflexName } from './controllers'
+import { attributeValue, attributeValues } from './attributes'
+import isolationMode from './isolation_mode'
+
+const reflexes = {}
+
+export default reflexes
+
+export const performOperations = data => {
```
jonathan-s, line 13: It would also be very nice if this could be exported from `stimulus_reflex.js`.

javascript/stimulus_reflex.js
julianrubisch: Yep, my gut feeling tells me that even `createSubscription` is something very websocket-specific.

my take would be to make `transports/action_cable` an actual adapter `class`, like in @jonathan-s 's original implementation, so that we can later swap it out for a different (AJAX) adapter more easily

hopsoft:
This looks great on cursory review. I definitely want to test some things locally. Is your comprehensive CableReady test harness publicly available? I'd love to start using it.

javascript/reflexes.js
leastbad: That's a good idea, although as someone just learning this stuff... wouldn't it make more sense to just import them from `reflexes.js`?

BTW - are you talking about `reflexes`, `performOperations` or both?

javascript/stimulus_reflex.js
leastbad: There will be no `consumer` option in v4; see #438. Consumer will come from the Stimulus application.

As I mentioned in the PR description, I exercised significant restraint in not changing the API. Even there, I did remove a vestigal and antiquated technique we had in place as a fallback for finding consumers because it was itself raising deprecation errors in Chrome. So, save your good ideas for changes until after v3.4.2 comes out!

leastbad:
> This looks great on cursory review. I definitely want to test some things locally. Is your comprehensive CableReady test harness publicly available? I'd love to start using it.

I would be thrilled for others to make use of my test harness for CR.

Note that there's lots of stuff commented out for PRs that haven't necessarily been approved, yet.

Just switch to the `cr_tests` branch of `stimulus_reflex_harness`:

https://github.com/leastbad/stimulus_reflex_harness/tree/cr_tests

javascript/reflexes.js
jonathan-s: Ah, yes. Exporting the entire module `reflexes.js` from stimulus_reflex.js might make more sense. From my point of view someone just learning this stuff shouldn't really need to learn about this since this is still pretty much internals of stimulus_reflex. Exposing these would help in being able to build on top of stimulus_reflex in a better way.

leastbad: Exporting `reflexes` actually makes good sense because it is the new location for what used to be `window.reflexes` - it's where all of the magic happens.

Exporting all of the functions in reflex.js through the main file to save someone implementing plugins to SR from having to import both is not a strategy that makes sense to me. If you need the functions from a file in a library, you import it. This is not an unreasonable expectation of someone building a transport plugin. You need to be at least comfortable with the internals, that's not something we can hide away.

hopsoft (approved):

javascript/stimulus_reflex.js
```diff
@@ -1,166 +1,75 @@
...
   Object.assign(controller, {
```
jonathan-s, line 73: There's possibly one more improvement that could be made here. You could move the object that you assign here into a named object, similar to what has been done elsewhere in the project.

leastbad: You'll have to forgive me, it's been over a month. Could you give me a reference to an example of the style/implementation that you're referring to so I can get my brain back in this timeline?

leastbad: To anyone watching: Jonathan and I back-channeled and we're on the same page, but what he's suggesting will come in later PRs which are more specifically/themeatically about removing AC-specific code from the main event loop.

PR #445, opened 2021-02-06
delegate broadcast to cable_ready_channels

leastbad:

Bug fix

## Description

@josefrichter discovered a bug in the CableReadyChannels proxy:

```rb
cable_ready["timeline"].inner_html(...)
cable_ready.broadcast # nothing happens
```

I realized that we need to delegate `broadcast` along with `[]`. @hopsoft will this have unintended side-effects?

## Why should this be added

Devs should be able to call `cable_ready.broadcast` in a Reflex and expect it to work consistently.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/cable_ready_channels.rb
+++ b/lib/stimulus_reflex/cable_ready_channels.rb
@@ -2,7 +2,7 @@
 
 module StimulusReflex
   class CableReadyChannels
-    delegate :[], to: "cable_ready_channels"
+    delegate :[], :broadcast, to: "cable_ready_channels"
 
     def initialize(stream_name)
       @stream_name = stream_name


julianrubisch:
I'd like to take the opportunity of this PR to point out a general smell:

We're making liberal use of `delegate` in our codebase because it's convenient, but tends to be brittle. 

Consider adding more methods to `CableReady::Channels`, which is likely to occur, we'll have to add it here, too. Classic shotgun surgery.

Using SimpleDelegator or Forwardable could have prevented this.

hopsoft:
`StimulusReflex::CableReadyChannels` exposes both of the following methods.

1. `#cable_ready_channels` - a thread local singleton that holds a set of operations for N number of ActionCable identifiers
 
    ```ruby
    def cable_ready_channels
      CableReady::Channels.instance
    end
    ```

2. `#stimulus_reflex_channel` - a representation of a single ActionCable identifier and its enqueued operations *(the default StimulusReflex channel)*

    ```ruby
    def stimulus_reflex_channel
      CableReady::Channels.instance[@stream_name]
    end
    ```

Both of these objects expose `broadcast` and `broadcast_to`. Consider the following.

```ruby
# returns an instance of a CableReady::Channel identified by "foo"
CableReady::Channels.instance["foo"] 

# example of using the thread local singleton
CableReady::Channels.instance["foo"].inner_html(...) 
CableReady::Channels.instance["bar"].inner_html(...)
CableReady::Channels.instance["baz"].inner_html(...)

# calling broadcast on a CableReady::Channel instance will only emit the broadcast for that specific identifier
# this line will only broadcast operations enqueued for the "foo" identifier
CableReady::Channels.instance["foo"].broadcast

# calling broadcast on the thread local singleton iterates all encapsulated
# channel instances and calls broadcast on each of them
# this line will broadcast all enqueued operations for the "foo", "bar", and "baz" identifiers  
CableReady::Channels.instance.broadcast 
```

`StimulusReflex::CableReadyChannels` currently delegates any unimplemented method calls to the default StimulusReflex channel `stimulus_reflex_channel`. That means that when we call `cable_ready.broadcast` in a reflex, it's broadcasting to the single default StimulusReflex channel instance.

I'm still undecided on what makes the most intuitive sense for developers regarding the reported problem. For example, the following examples would all work for the scenario outlined above.

```ruby
cable_ready["timeline"].inner_html(...).broadcast
```

```ruby
cable_ready["timeline"].inner_html(...)
cable_ready["timeline"].broadcast
```

```ruby
# this is basically what's currently proposed in the PR
cable_ready["timeline"].inner_html(...)
cable_ready_channels.broadcast 
# this would have the side effect of broadcasting all 
# operations enqueued for all identifiers held in the thread local singleton 
```

I'm pretty certain we don't want the behavior described in that last example, but I'm curious to hear what others think.

marcoroth:
I agree that we don't want the behavior described in the last example. 

Spontaneous thought: What about a `broadcast_all` or similar method?

julianrubisch:
It seems like we’re hitting some limitations of delegation here. 

FWIW I think it’s just confusing that the usual CableReady methodology (that’s mentioned in the CR 101) _doesn‘t work_.

I think this really calls for `SimpleDelegator` where we could put some additional logic into the `broadcast` method.

leastbad:
**Thank you** for spelling this out in plain language, Nate.

My $0.02 is that given the above, I should close this PR and explain in the documentation that when using the special "Reflex version" of the `cable_ready` method, `broadcast` has to be at the end of a chain.

Given that there's nothing stopping them from using `cable_ready[stream_name]` if it suits them, I have no problem at all with how things currently work.

Keep in mind that they can still chain addition operations on separate lines. It's just that one of those will have to [eventually] call `broadcast`.

julianrubisch:
I agree with @leastbad. I think it's better to defer this in favor of a later, more exhaustive delegation refactor. We should keep an eye on it, though.

PR #449, opened 2021-02-08
permessage deflate

leastbad:

Enhancement

## Description

My [currently failed] attempt to enable permessage_deflate as a monkey patch in SR.

Before:
chrome_Dx9qik8Txm
After:
chrome_DsMCPolonR

Despite promising initial results, it's shitting the bed with segfaults and I don't really know what seems to be the problem. Seeing "Invalid frame header" responses. What's weird is that some messages seem to manage to get through.

I've tried all of the possible values of `:max_window_bits` from 8 to 15.

My current solution is stolen verbatim from https://sapandiwakar.in/compressing-websocket-actioncable-frames-on-rails-with-permessage-deflate/ and was helped/influenced by https://github.com/faye/permessage-deflate-ruby

Anyhow, just sharing here in case someone has some detective juice.

## Why should this be added

Trying to enable brotli compression for WS traffic.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,6 +4,7 @@ PATH
     stimulus_reflex (3.4.1)
       cable_ready (>= 4.5)
       nokogiri
+      permessage_deflate
       rack
       rails (>= 5.2)
       redis
@@ -81,7 +82,7 @@ GEM
     erubi (1.10.0)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
-    i18n (1.8.7)
+    i18n (1.8.8)
       concurrent-ruby (~> 1.0)
     loofah (2.9.0)
       crass (~> 1.0.2)
@@ -95,13 +96,14 @@ GEM
     mini_mime (1.0.2)
     mini_portile2 (2.5.0)
     minitest (5.14.3)
-    nio4r (2.5.4)
+    nio4r (2.5.5)
     nokogiri (1.11.1)
       mini_portile2 (~> 2.5.0)
       racc (~> 1.4)
     parallel (1.20.1)
     parser (3.0.0.0)
       ast (~> 2.4.1)
+    permessage_deflate (0.1.4)
     pry (0.12.2)
       coderay (~> 1.1.0)
       method_source (~> 0.9.0)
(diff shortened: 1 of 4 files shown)


leastbad:
So, I really appreciate you opening this topic of conversation today. It's gotten me thinking about a lot of issues, even if your problem is still outstanding. For example, I'm going to attempt to add an optional `permessage-deflate` setup step to the upcoming SR installer. Exciting!

Anyhow, in the abstract, if someone lands on this page wondering how to set up gzip for WS traffic, it comes down to two steps. You need to add the `permessage_deflate` gem to your project, and you need this in an initializer:
```ruby
module ActionCable
  module Connection
    class ClientSocket
      alias old_initialize initialize
      def initialize(env, event_target, event_loop, protocols)
        old_initialize(env, event_target, event_loop, protocols)
        @driver.add_extension(PermessageDeflate.configure(level: Zlib::BEST_COMPRESSION))
      end
    end
  end
end
```
I decided to omit the modification to the window size until such time as @jcoglan or someone else can explain what it actually does in mortal/layman terms. Try to restrict our cargo culting here!

jcoglan:
@leastbad So I don't understand the internals of zlib in depth; if you read the source of https://github.com/faye/permessage-deflate-ruby you'll see it's a pretty thin wrapper around zlib, plus the parameter negotiation rules that the WebSocket standard requires. But I'll give my rough understanding of what this parameter does.

In order to compress a stream of bytes, zlib maintains a data structure that buffers the previous N bytes of the stream and uses it to compress the next bytes. This buffer is the "sliding window". The bigger this buffer is, the more chance there is that the next bytes will match something in it, and get compressed. So making the buffer bigger improves the compression ratio, but uses more memory. Per the [docs](https://www.zlib.net/manual.html):

> The `windowBits` parameter is the base two logarithm of the window size (the size of the history buffer). It should be in the range 8..15 for this version of the library. Larger values of this parameter result in better compression at the expense of memory usage. The default value is 15 if `deflateInit` is used instead.

So for example a `windowBits` of 8 means zlib will buffer at most the last 256 bytes of the stream at any point in time.

Whatever window size the compressor uses, the decompressor must use a window of _at least_ that size in order to parse the compressed data and reconstruct the original data. The compressed data encodes pointers into previous portions of the stream, so the decompressor needs to remember enough of the stream to make sense of these pointers.

You don't want the other peer to use a window that uses more memory than you can handle, and that's why `request_max_window_bits` exists -- to tell the other peer not to use a window that's bigger than you have memory for. The `max_window_bits` option just sets your own window size, and the protocol negotiates a window size that both peers can work with.

Note that since zlib v1.2.9, setting `windowBits` to 8 is not allowed. We handle this in `permessage_deflate` by setting this value to 9 if the caller requests 8.

leastbad:
Thank you for this, @jcoglan - that's exactly what I needed.

PR #451, opened 2021-02-09
Extract RequestParameters class

julianrubisch:
# Refactoring

## Description

I extracted a `StimulusReflex::RequestParameters` class that should wrap the handling of path params etc. nicely, and in the end also allow for some testing to be performed around it.

## Why should this be added

Housekeeping.

## Checklist


lib/stimulus_reflex/reflex.rb
```diff
@@ -71,13 +71,9 @@ def request
...
-      path_params = Rails.application.routes.recognize_path_with_request(req, url, req.env[:extras] || {})
-      path_params[:controller] = path_params[:controller].force_encoding("UTF-8")
-      path_params[:action] = path_params[:action].force_encoding("UTF-8")
-
-      req.env.merge(ActionDispatch::Http::Parameters::PARAMETERS_KEY => path_params)
-      req.env["action_dispatch.request.parameters"] = req.parameters.merge(@params)
-      req.tap { |r| r.session.send :load! }
+      # fetch path params (controller, action, ...) and apply them
+      request_params = StimulusReflex::RequestParameters.new(params: @params, req: req, url: url)
+      request_params.apply!
```
marcoroth, line 76: Tbh I think it's a bit confusing that `request_params.apply!` returns the update `req` object which then gets returned as the result of the `#request` method.

julianrubisch: what do you suggest?

marcoroth: Maybe it would be enough to just put the `req` at the end of the method so you know what you are returing. 

If I read that code without knowing what `RequestParameters#apply!` actually does I would assume that it returns an instance of `StimulusReflex::RequestParameters`

```suggestion
      request_params.apply!

      req
```

julianrubisch: Well IIRC the `req` isn't mutated in place, so we'd have to re-assign it first - but you're correct about the intent

hopsoft (approved):

PR #464, opened 2021-03-04
Stimulus reflex  concern

julianrubisch:
# Enhancement

## Description

This PR adds a `ConcernEnhancer` module that can be included alongside an `ActiveSupport::Concern` (on which's existence it depends), producing a concern that can be used by both reflex and controller (and models, for that matter) without having to guard against undefined methods (like `helper_method`, `before_action` etc.).

It basically adds a `method_missing` class method to the includer, which will stub class methods for the counterpart after switching on the ancestor (whether it's a `StimulusReflex::Reflex`, `ActionController::Base` etc.)

**NOTE for docs:** this module has to be `include`d as opposed to `ActiveSupport::Concern`, which has to be `extend`ed:

```rb
module TestConcern
  extend ActiveSupport::Concern
  include StimulusReflex::ConcernEnhancer

  # .. more methods
end
```

## Why should this be added

To more resiliently share code between controllers/models and reflexes

## Checklist


hopsoft:
Can this be updated so usage looks more like this?

```ruby
module TestConcern
  extend StimulusReflex::SharableConcern
end
```

hopsoft:
We also need to think about where we recommend that such concerns would exist in the project structure so it can be documented.

```
app/controllers/concerns
app/reflexes/concerns
app/???
```

hopsoft:
My preference is `app/controllers/concerns`

PR #475, opened 2021-03-20
add `useReflex` js function to "support" the composable pattern

marcoroth:
# Type of PR 

"Feature"

## Description

This PR adds the`useReflex()` js function to "support" the composable pattern.

## Why should this be added

This can be useful if you include a handful of mixins in your `connect()` function (e.g. some mixins from `stimulus-use`) and want them to all look similar. `StimulusReflex.register(this)` would stick out in that matter.

### Example

```js
import { useReflex } from 'stimulus_reflex'
import { useResize, useIdle } from 'stimulus-use'

  connect() {
    useReflex(this)
    useResize(this)
    useIdle(this)
    // ...
  }

  // ...
}
```

Reference:
* https://github.com/stimulus-use/stimulus-use
* https://www.betterstimulus.com/architecture/mixins.html

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -224,6 +224,10 @@ const register = (controller, options = {}) => {
   })
 }
 
+const useReflex = (controller, options = {}) => {
+  register(controller, options)
+}
+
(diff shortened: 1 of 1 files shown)


leastbad:
Noice!

PR #476, opened 2021-03-22
Log reflex payloads client-side

julianrubisch:
# Enhancement

## Description

Added the newly added `payload` to the client side logger

## Why should this be added

DX? 🤷‍♂️

## Checklist

### DIFF
--- a/javascript/log.js
+++ b/javascript/log.js
@@ -20,7 +20,7 @@ function request (
 
 function success (event) {
   const { detail } = event || {}
-  const { selector } = detail || {}
+  const { selector, payload } = detail || {}
   const { reflexId, target, morph, serverMessage } = detail.stimulusReflex || {}
   const reflex = reflexes[reflexId]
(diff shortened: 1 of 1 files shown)


leastbad:
On face value, this is amazing as delivered. What if someone uses the payload to deliver a large string, however? I'd recommend converting the `payload` to a String and constraining the `payload` to something really modest like 6-8 characters, and then appending a "..." if it exceeds that.

julianrubisch:
Hmm not sure this would be as helpful... if I'm returning a JavaScript object, I want to inspect that, and not a string

What am I missing? What's the harm with rendering a large string in the debug output?

leastbad:
Actually, given that this will be inside of the object following the log message, my concern is abated.

My fear was that the payload would become part of the main debug message, but that's not what's being proposed here. I have no problem with the value being available for inspection inside of an object you have to click to expand, anyhow.

PR #477, opened 2021-03-24
Reflex return payloads for events and callbacks

leastbad:

Enhancement

## Description

Adds Reflex return payloads to life-cycle events and callback methods.

## Why should this be added

Consistent API.

## Checklist


javascript/callbacks.js
```diff
@@ -58,7 +61,7 @@ export const afterDOMUpdate = event => {
         element: reflexElement,
         event,
         data: promise.data,
-        payload: event.detail.payload
+        payload
```
hopsoft, line 64: Would it make sense to use a named argument for payload similar to what we do with data?

```js
data: promise.data,
payload: payload
```

hopsoft (approved):

PR #483, opened 2021-04-05
Fix Changelog Action

marcoroth:
# Type of PR

Bug Fix

## Description

As noted in the [README of `ad-m/github-push-action`](https://github.com/ad-m/github-push-action#example-workflow-file) it seems that the `actions/checkout` action needs the following params:

```yml
    - uses: actions/checkout@v2
      with:
        persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
        fetch-depth: 0 # otherwise, it will fail to push refs to dest repo
```

(It now also uses the built-in bundler cache provided by the `ruby/setup-ruby` action)

## Why should this be added

The Changelog Actions currently doesn't work

## Checklist

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -15,28 +15,30 @@ jobs:
     timeout-minutes: 4
     if: "!contains(github.event.head_commit.message, 'Update Changelog')"
     steps:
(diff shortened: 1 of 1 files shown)


leastbad:
Let's give it a shot!

PR #485, opened 2021-04-06
delegate dom_id to CableReady::Identifiable

leastbad:

Enhancement

**Hold until the next version of CableReady is released.**

Tests will not pass until SR points to `CR 5.0.0-pre0`.

## Description

1. Reflexes will use `CableReady::Identifiable.dom_id`
2. `render_collection` method no longer passes an empty `hash` parameter to `dom_id`

## Why should this be added

No need to maintain two separate implementations of the server-side `dom_id` method.

## Checklist


lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
```diff
@@ -11,7 +13,7 @@ def broadcast(_, data = {})
           html = reflex.wrap(reflex.render(key), key) if key.is_a?(ActiveRecord::Relation) && value.nil?
           fragment = Nokogiri::HTML.fragment(html&.to_s || "")
 
-          selector = key.is_a?(ActiveRecord::Base) || key.is_a?(ActiveRecord::Relation) ? reflex.dom_id(key) : key
+          selector = key.is_a?(ActiveRecord::Base) || key.is_a?(ActiveRecord::Relation) ? dom_id(key) : key
```
marcoroth, line 16:
```suggestion
          selector = dom_id(key)
```

Can't we just anytime pass this to `dom_id`? 

`CableReady:: Identifiable` should handle both the `ActiveRecord::Base` and `ActiveRecord::Relation` case:

https://github.com/hopsoft/cable_ready/blob/d3663cc98c992633c96724c2232a44ca02830bc8/lib/cable_ready/identifiable.rb#L8-L11

leastbad: Maybe. I'm sleepy so I should be careful, but I think that in this scenario, CSS class selectors and other stuff like `[attribute]` are valid.

marcoroth: That's right. Forget my comment 😅

hopsoft (approved):
Beautiful.

PR #486, opened 2021-04-08
Factor reflex creation and reflex data management out of channel

julianrubisch:
# Refactor

## Description

Basically most of the attributes derived from the `data` hash passed to `receive` have been factored out to an own `ReflexData` class. Furthermore the logic for actually creating the reflex has been moved to a `ReflexFactory`.

## Why should this be added

This not only makes the class more readable but will also facilitate discussions about what part of it should be public API in an upcoming plugin system

## Checklist


app/channels/stimulus_reflex/channel.rb
```diff
@@ -144,4 +120,8 @@ def report_failed_basic_auth(reflex)
   def exception_message_with_backtrace(exception)
     "#{exception}\n#{exception.backtrace.first}"
   end
+
+  def arguments
```
julianrubisch, line 124: Oops, forgot that reference...

PR #478, opened 2021-03-25
data-reflex-dataset and data-reflex-dataset-all

leastbad:
# Type of PR (feature, enhancement, bug fix, etc.)

Enhancement

## Description

@joshleblanc gave me the idea to expand what could be passed to a `data-reflex-dataset` attribute.

What I've done in this PR is re-imagined the `extractElementDataset` function as a power tool for grabbing attributes from relative elements.

`combined` has been deprecated in favour of `ancestors`. ~~The deprecation itself is commented until #438 is merged.~~

`data-reflex-dataset` and `data-reflex-dataset-all` support one or multiple space-separated tokens. eg

```
data-reflex-dataset="siblings div.success #sleeper parent"
```

Tokens can be keywords or CSS queries. They are evaluated in the order that they are included.

Currently proposed and supported keywords are:
- combined (will emit a deprecation warning)
- ancestors
- parent
- siblings
- children
- descendants

## data-reflex-dataset

Current (3.4.1) behaviour of grabbing and holding the first unique value for a given key is maintained:

```html
<div data-reflex-dataset=".post">
  <div class="post" data-post-id="1"></div>
  <div class="post" data-post-id="2"></div>
  <div class="post" data-post-id="3"></div>
  <div class="post" data-post-id="4"></div>
</div>
```

The server will receive:

```
{
 'data-reflex-dataset': '.post',
 'data-post-id': '1'
}
```

## data-reflex-dataset-all

This introduces a 2nd form which enables the scooping of a group of matching elements:

```html
<div data-reflex-dataset-all=".post">
  <div class="post" data-post-id="1"></div>
  <div class="post" data-post-id="2"></div>
  <div class="post" data-post-id="3"></div>
  <div class="post" data-post-id="4"></div>
</div>
```

The server will receive:

```
{
 'data-reflex-dataset-alls': ['.post'],
 'data-post-ids': [
   '1',
   '2',
   '3',
   '4'
 ]
}
```

Since the data-post-id always contains the first occurring value (like it was before) it's fully backwards compatible, because just the data-post-ids is added additionally.

You can of course also mix-and-match the two attributes.

## Let's try both!

```html
<div data-reflex-dataset=".post" data-reflex-dataset-all=".post">
  <div class="post" data-post-id="1"></div>
  <div class="post" data-post-id="2"></div>
  <div class="post" data-post-id="3"></div>
  <div class="post" data-post-id="4"></div>
</div>
```

The server will receive:

```
{
 'data-reflex-dataset-all': '.post',
[...]

PR #489, opened 2021-04-14
Allow to execute CableReady operations on `StimulusReflex::Element`

marcoroth:
# Type of PR

Feature

## Description

As of today we just have one `StimulusReflex::Element` instance which is available as `element` in your reflex.

This PR enables to call  CableReady operations on any instance of the said `StimulusReflex::Element` class. Like this we enable a very DOM-like and natural API in a reflex action to operate on those elements.

Additionally it adds another element accessor to the reflex class named `controller_element`. This is the element in the DOM which holds the Stimulus Controller you are operating on.

This enables scenarios like:

```ruby
class CounterReflex < StimulusReflex::Reflex
  def increment
    @count = element.dataset.count.to_i + element.dataset.step.to_i

    element.inner_html(html: @count)
    element.dispatch_event(name: 'increment', detail: { count: @count })
    element.set_focus

    controller_element.set_dataset_property(name: "counterValue", value: @count)

    morph :nothing
  end
end
```

This PR is part of a bigger story. The PR #XYZ proposes the introduction of StimulusReflex targets. With that you can declare any DOM node as a StimulusReflex target (like you do in your Stimulus controller). This enables you to then access and operate on the targets in your Reflexes the same way you can with `element` and `controller_element`.

This work is based on #478 and only meant to be merged after #478 is merged.

## Why should this be added

Makes the API even more awesome and straight-forward, it feels like you are operating on an actual DOM Node via the actual DOM API. It also makes the use of CableReady more transparent.

## Checklist

### DIFF
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -23,7 +23,6 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
     reflex_name = reflex_name.camelize
     reflex_name = reflex_name.end_with?("Reflex") ? reflex_name : "#{reflex_name}Reflex"
     arguments = (data["args"] || []).map { |arg| object_with_indifferent_access arg }
-    element = StimulusReflex::Element.new(data)
     permanent_attribute_name = data["permanentAttributeName"]
     form_data = Rack::Utils.parse_nested_query(data["formData"])
     params = form_data.deep_merge(data["params"] || {})
@@ -31,9 +30,10 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
     begin
       begin
         reflex_class = reflex_name.constantize.tap { |klass| raise ArgumentError.new("#{reflex_name} is not a StimulusReflex::Reflex") unless is_reflex?(klass) }
-        reflex = reflex_class.new(self,
+        reflex = reflex_class.new(
+          self,
           url: url,
-          element: element,
+          data: data,
           selectors: selectors,
           method_name: method_name,
           params: params,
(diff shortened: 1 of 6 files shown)


Matt-Yorkley:
Wooooah this is wild :heart: 

Is `element.broadcast` getting called implicitly somewhere when the reflex is finished? It's not called explicitly in the example...
```ruby
element.inner_html(html: @count)
element.dispatch_event(name: 'increment', detail: { count: @count })
element.set_focus
```

marcoroth:
@Matt-Yorkley Yeah, StimulusReflex calls `cable_ready.broadcast` at the end of every reflex execution. This is also already happening today if you use the `cable_ready` inside a reflex action.

But since calling `element.inner_html(...)` just enqueues a regular operation using `cable_ready` under the hood it's also getting flushed at the same time.

Matt-Yorkley:
I just gave this a little live-test on top of `v3.5.rc2` and it almost works perfectly, except the `selector` attribute on the `element` is in xpath format and the cable ready operations default to `xpath: false`.

```
StimulusReflex::Channel transmitting {"cableReady"=>true, "operations"=>[{"selector"=>"/html/body/section[1]/div[1]/div[2]/button[1]", "xpath"=>false, "text"=>"TEST", "reflexId"=>"b913191f-bf74-49b0-8e07-c15ecae87964", "operation"=>"textContent"}
```

Looks like it just needs a single slash instead of a double in `StimulusReflex::Element#method_missing`:
```diff
-- xpath = selector ? selector.starts_with?("//") : false
++ xpath = selector ? selector.starts_with?("/") : false
```
It works!

marcoroth:
@Matt-Yorkley this is great news, thanks for giving this a shot!

PR #490, opened 2021-04-14
Introduce StimulusReflex Targets

marcoroth:
# Type of PR

Feature

## Description

This PR is part of a bigger story. This work is based on #478  and #489.

With this PR we allow to declare DOM nodes as StimulusReflex targets so you are able to access and operate on them with CableRady operations in your reflexes. 

Here is an example:

```html
<div data-controller="folder" data-folder-id="42">
  <input type="text" data-reflex-target="folder.input">
  <button data-reflex="click->Folder#search" data-reflex-dataset="combined">Search</button>

  <span id="count" data-reflex-target="folder.count"></span>

  <div id="results" data-reflex-target="folder.results"></div>
</div>
```

```ruby
class FolderReflex < StimulusReflex::Reflex
  def search
    folder = Folder.find_by(id: element.dataset.folder_id)

    controller_element.append(html: '<div id="spinner"></div>')
    element.set_attribute(name: 'value', value: '').set_focus.update

    results = folder.search(input_target.value)

    count_target.text_content(text: results.count)
    results_target.inner_html(html: render(partial: "folders/results", locals: { results: results }))
 
    cable_ready.remove(selector: '#spinner')

    morph :nothing
  end
end
```

## Why should this be added

Targets are well known to Stimulus developers. This PR applies the same target-concept to reflexes. This, in combination with #489, enables a whole new world on how you write your CableReady operations while writing clean, super understandable and easy-to-look-at code.

## Checklist

### DIFF
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -19,6 +19,7 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
     selectors = (data["selectors"] || []).select(&:present?)
     selectors = data["selectors"] = ["body"] if selectors.blank?
     target = data["target"].to_s
+    targets = data["targets"] || {}
     reflex_name, method_name = target.split("#")
     reflex_name = reflex_name.camelize
     reflex_name = reflex_name.end_with?("Reflex") ? reflex_name : "#{reflex_name}Reflex"
@@ -36,6 +37,7 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
           data: data,
           selectors: selectors,
           method_name: method_name,
+          targets: targets,
           params: params,
           client_attributes: {
             reflex_id: data["reflexId"],
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -23,7 +23,6 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
     reflex_name = reflex_name.camelize
     reflex_name = reflex_name.end_with?("Reflex") ? reflex_name : "#{reflex_name}Reflex"
     arguments = (data["args"] || []).map { |arg| object_with_indifferent_access arg }
-    element = StimulusReflex::Element.new(data)
     permanent_attribute_name = data["permanentAttributeName"]
     form_data = Rack::Utils.parse_nested_query(data["formData"])
     params = form_data.deep_merge(data["params"] || {})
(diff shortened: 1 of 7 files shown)


Matt-Yorkley:
After playing with a working version of this and taking it for a bit of a test drive, I had a few notes and bits of feedback regarding the **open questions** in the description.

#### Restricting named targets to the controller element

In the case where there's a custom Stimulus controller on a whole section of the page (maybe a form for example) and it has multiple child elements and calls `this.stimulate()`, the idea of potentially only including targets that are children of `controller_element` makes a lot of sense and might be preferable in some cases.

But in the simpler case of a stand-alone element (like a button with `data-reflex="click->Example#action"`) then _the element is the controller_ and the concept of "only include targets that are child elements of the controller" doesn't make sense. In that case it wouldn't be possible to interact with any potential targets, because they would necessarily be outside of `controller_element`. Know what I mean?

It really feels like it should be at least _possible_ to use named targets that are anywhere on the current page.

#### Named targets anywhere on the page

If named targets can be anywhere on the page and aren't restricted to being children of `controller_element` then that creates some other potential issues, like: if the additional data about all named targets on the page is added into every client->server interaction by default would it would probably end up needlessly increasing the wire size in lots of situations where the targets might not even be used/required in a way that isn't preferable.

My thoughts around these questions are coalescing towards the idea that the ability to control whether or not this additional data gets added and what the scope of it should be really wants to be represented by an optional parameter of some kind. Roughly what I'm leaning towards is: by default the additional data about targets does not get included. Optionally; data for only the named targets that are children of `controller_element` can be included by declaring something like `include_targets: "controller"`, or; data about all named targets on the current page can be included with something like `include_targets: "page"`. Either way it feels like it wants to be _declared_. Maybe on the element? Like...

```html
<span data-reflex-target="whatever"> Target Me! </span>

<button data-reflex="click->Example#action_one"> Not using targets </button>

<button data-reflex="click->Example#action_two" data-include-targets="page"> Targetting! </button>

<form data-controller="stimulating" data-include-targets="controller" data-action="submit->stimulating#submit">
  <input data-reflex-target="launchcodes" type="text" />
  <button data-reflex-target="bigredbutton" class="big red"> Boom! </button>
</from>
```

#### Support for multiple targets with the same name

This feels like it would be amazing, but it'll need some changes to how the targets are put together and what the interface for accessing them looks like. Also... there are some interesting questions about calling chainable operations directly on a _collection_ of targets.

I had a quick look and I'm pretty sure that all operations that could feasibly be called on an _element_ have the optional `select_all` parameter. So... in the case of a collection, if there's an accessor method exposed for something like `post_targets` which represents a _collection_ of elements with the same target name... then maybe calling `post_targets.add_css_class(name: "active")` could just automatically set `select_all: true` when building that operation...? Hmmm... it feels like there's a bit of finesse needed there with regards to the underlying mechanics of it, but I think it can work. It would need the `selector` argument to be changed to a `selector: "[data-reflex-target='<target_name>']"` format in that case, but currently it's using xpath when extracting the target data on the client side.

#### A side note on simplicity

An interesting part of this change is that by removing the need to specify the `selector` argument explicitly (because it gets inserted automatically) it means that certain operations don't require any arguments at all... for example, you could write: 

```ruby
button_target.remove
```

And that translates to a valid `cable_ready` operation which does exactly what you'd expect... :eyes:

PR #491, opened 2021-04-21
Log the full exception trace

rmckayfleming:

Enhancement

## Description

Render the entire stack trace instead of just the top level error.

## Why should this be added

Simply put, this makes debugging a lot easier especially for exceptions caused by ViewComponents. I've often run into cases where the exception comes deep from within Rails and without the full trace, it's hard to know what part of your code tripped it.

If it'd be better to gate this behaviour change behind an option or something, I'd be happy to make the change (I'd just need to know how best to handle it).

## Checklist

### DIFF
--- a/lib/stimulus_reflex/broadcasters/broadcaster.rb
+++ b/lib/stimulus_reflex/broadcasters/broadcaster.rb
@@ -23,8 +23,7 @@ module StimulusReflex
       false
     end
 
-    def broadcast_message(subject:, body: nil, data: {}, error: nil)
-      logger.error "\e[31m#{body}\e[0m" if subject == "error"
+    def broadcast_message(subject:, data: {}, error: nil)
       operations << ["document", :dispatch_event]
       cable_ready.dispatch_event(
         name: "stimulus-reflex:server-message",
--- a/app/channels/stimulus_reflex/channel.rb
+++ b/app/channels/stimulus_reflex/channel.rb
@@ -24,14 +24,14 @@ class StimulusReflex::Channel < StimulusReflex.configuration.parent_channel.cons
         delegate_call_to_reflex reflex
       rescue => exception
         error = exception_with_backtrace(exception)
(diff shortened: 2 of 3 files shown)


leastbad:
Hey Ryan!

I'm curious: would you be happiest if full stack traces occured:
1. only on the server
2. only on the client console
3. both

I'm currently thinking that what makes the most sense is to `puts` the full trace and keep what the client shows short. What do you think?

leastbad:
Hey again! 😀 

Apologies for moving into your PR like I own the place. Straight talk: you had the right idea, but the implementation needed was closer to a refactoring of the module. I want you to retain credit and be recognized as a contributor to StimulusReflex.

Thanks for kicking things off!

lib/stimulus_reflex/broadcasters/broadcaster.rb
```diff
@@ -23,8 +23,7 @@ def selector?
       false
     end
 
-    def broadcast_message(subject:, body: nil, data: {}, error: nil)
-      logger.error "\e[31m#{body}\e[0m" if subject == "error"
+    def broadcast_message(subject:, data: {}, error: nil)
```
hopsoft, line 26: I think changing this method signature will be a breaking change. We don't know if or how people might be using this in the wild.

hopsoft (changes requested):
This looks good. I had one minor concern.

lib/stimulus_reflex/broadcasters/broadcaster.rb
leastbad: Uhh! Since when is `StimulusReflex::Broadcasters` a public API?

I can't think of a single legitimate reason for someone to call it unless they are us. What am I missing?

julianrubisch: I'm with @leastbad here

hopsoft (approved):

PR #492, opened 2021-04-23
Stop adding whitespace when parsing our html

### LINKED ISSUE #487: Nokogiri usage to process HTML adds whitespace nodes, causing unneeded morphdom updates
Stimulus Reflex uses `Nokogiri::HTML` to parse the rendered output of the Rails controller and selects a fragment of the HTML using a CSS selector:

https://github.com/hopsoft/stimulus_reflex/blob/ae318edc1e96d531daf420db6e715d76bf24e90c/lib/stimulus_reflex/broadcasters/page_broadcaster.rb#L11-L15

These fragments are then sent across a WebSocket to the browser where they are handed to morphdom and merged into the browser's DOM. Morphdom has [some issues](https://github.com/patrick-steele-idem/morphdom/issues/51) where changes in text nodes, even if they are just whitespace, will cause unneeded element removal and addition. 

This manifests in our app as the autocomplete on a form field closing unexpectedly. You can see the same behavior in the [Stimulus Reflex demo validation application](http://expo.stimulusreflex.com/demos/validation):

Example of morphdom needlessly changing the input element

Here we:

1. Load the page
2. Click on the field (seeing the autocomplete)
3. Type a letter (triggering a Reflex action)
4. **Morphdom deletes and recreates the `input` tag** (closing the autocomplete)
5. Any future Reflex actions will have the same kinds of extra whitespace, so the autocomplete field will stay open.

In our application, the HTML has very few newlines as Slim strips almost all of them out. This means that something in the Stimulus Reflex pipeline is *adding* newlines. Through debugging, this is caused by the usage of Nokogiri:

```ruby
input = "<!DOCTYPE html><html><head><title>Example</title></head></html>"
nokogiri = Nokogiri::HTML(input)

# This now has newlines inserted!
nokogiri.to_html
# => "<!DOCTYPE html>\n<html><head>\n<meta http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\">\n<title>Example</title>\n</head></html>\n"
```

These differences in whitespace cause the morphdom issue.

## Expected behavior

The HTML sent by Stimulus Reflex should have the same amount of whitespace as what Rails created.

### StimulusReflex

- Gem: 3.4.1
- Node package: 3.4.1

### External tools

- Ruby: 2.7.2p137
- Rails: 6.1.3.1
- Node: 14.16.0

### Browser

- Chrome Canary 91.0.4471.0
- Safari 14.0.3

lmatiolis:

Bug fix.

## Description

As shown by @shepmaster on #487, we were adding extra whitespace
to the DOM when parsing. That was causing morphdom to do several
replaces on the first page load since it was comparing a page without
the whitespace and the "new" page with all the whitespaces added.

The solution was just saying for Nokogiri to use the DEFAULT_HTML
but without doing the FORMAT part.

Fixes #487

## Why should this be added

With the extra whitespace being added, we were re-rendering several parts of the DOM without it really being different. It was unnecessary re-renders and, for a more specific issue, it was causing the browser to close its autocomplete on a field cause it was replacing the field with a new one even though they were still the exact same DOM (but with extra whitespace).

## Checklist

### DIFF
--- a/lib/stimulus_reflex/broadcasters/broadcaster.rb
+++ b/lib/stimulus_reflex/broadcasters/broadcaster.rb
@@ -5,6 +5,9 @@ module StimulusReflex
     attr_reader :reflex, :logger, :operations
     delegate :cable_ready, :permanent_attribute_name, :payload, to: :reflex
 
+    DEFAULT_HTML_WITHOUT_FORMAT = Nokogiri::XML::Node::SaveOptions::DEFAULT_HTML &
+      ~Nokogiri::XML::Node::SaveOptions::FORMAT
+
     def initialize(reflex)
       @reflex = reflex
       @logger = Rails.logger if defined?(Rails.logger)
--- a/lib/stimulus_reflex/broadcasters/page_broadcaster.rb
+++ b/lib/stimulus_reflex/broadcasters/page_broadcaster.rb
@@ -12,7 +12,7 @@ module StimulusReflex
       selectors = selectors.select { |s| document.css(s).present? }
       selectors.each do |selector|
         operations << [selector, :morph]
-        html = document.css(selector).inner_html
+        html = document.css(selector).inner_html(save_with: Broadcaster::DEFAULT_HTML_WITHOUT_FORMAT)
         cable_ready.morph(
           selector: selector,
           html: html,
--- a/lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
+++ b/lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
@@ -17,7 +17,7 @@ module StimulusReflex
             operations << [selector, :morph]
             cable_ready.morph(
               selector: selector,
-              html: match.inner_html,
+              html: match.inner_html(save_with: Broadcaster::DEFAULT_HTML_WITHOUT_FORMAT),
               payload: payload,
               children_only: true,
               permanent_attribute_name: permanent_attribute_name,
--- a/test/broadcasters/page_broadcaster_test.rb
+++ b/test/broadcasters/page_broadcaster_test.rb
@@ -12,7 +12,7 @@ class StimulusReflex::PageBroadcasterTest < StimulusReflex::BroadcasterTestCase
   test "performs a page morph on body" do
     class << @reflex.controller.response
       def body
-        "<html><head></head><body>New Content</body></html>"
+        "<html><head></head><body><div>New Content</div><div>Another Content</div></body></html>"
       end
     end
 
@@ -24,7 +24,7 @@ class StimulusReflex::PageBroadcasterTest < StimulusReflex::BroadcasterTestCase
         "morph" => [
           {
             "selector" => "body",
-            "html" => "New Content",
+            "html" => "<div>New Content</div><div>Another Content</div>",
             "payload" => {},
             "childrenOnly" => true,
             "permanentAttributeName" => nil,
--- a/test/broadcasters/selector_broadcaster_test.rb
+++ b/test/broadcasters/selector_broadcaster_test.rb
@@ -1,59 +1,61 @@
 # frozen_string_literal: true
 
-# require_relative "broadcaster_test_case"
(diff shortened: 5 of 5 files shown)


lmatiolis:
Please let me know if I should rebase the commits after a review. Thanks! o/

leastbad:
This could be a silly question/approach, but...

```
[1] pry(#<StimulusReflex::PageBroadcaster>)> Nokogiri::XML::Node::SaveOptions::DEFAULT_HTML
=> 71
[2] pry(#<StimulusReflex::PageBroadcaster>)> Nokogiri::XML::Node::SaveOptions::FORMAT
=> 1
[3] pry(#<StimulusReflex::PageBroadcaster>)> (Nokogiri::XML::Node::SaveOptions::DEFAULT_HTML & ~Nokogiri::XML::Node::SaveOptions::FORMAT)
=> 70
[4] pry(#<StimulusReflex::PageBroadcaster>)> (Nokogiri::XML::Node::SaveOptions::DEFAULT_HTML & ~Nokogiri::XML::Node::SaveOptions::FORMAT).class
=> Integer
```

Doesn't this mean that we could just call:

```rb
html = document.css(selector).inner_html(save_with: 70)
```

It just seems performative if we know that the return value of that method is **always** going to be `70`... doesn't it?

leastbad:
Hey, sorry for pushing ahead without waiting for reply. We're working really hard to get a release ready, and I wanted to make sure that this was in v3.5.

Could you give this a shot and verify that it solves your issue, please?

lmatiolis:
> Hey, sorry for pushing ahead without waiting for reply. 

No worries at all! I'm glad we're moving this forward, no need to apologize o/

> Doesn't this mean that we could just call:
> html = document.css(selector).inner_html(save_with: 70)
> It just seems performative if we know that the return value of that method is always going to be 70... doesn't it?

Yes, the end result will still be the same, but I think there's a risk involved that we're relying on a constant that's not us that are defining? Nokogiri/libxml2 could change that value (probably not likely, but it's a possibility) and everything would break for us. Using Nokogiri constant, we guarantee that if that's changed, they will update their constant to keep using the proper value.

I think that using just a "random" number like that would at least need a comment of some sort, cause we're losing readability on what that is and where it's coming from.

You mentioned performance, but I think that overhead is minimal on this case, no? I suppose we could instead do a constant ourselves instead of a function call too:

`DEFAULT_HTML_WITHOUT_FORMAT =  Nokogiri::XML::Node::SaveOptions::DEFAULT_HTML & ~Nokogiri::XML::Node::SaveOptions::FORMAT`

Let me know your thoughts on this, I can test it once we decide the best approach.

Thanks!

marcoroth:
I'm with @lmatiolis on this one. I think we should keep the `DEFAULT_HTML` and `FORMAT` constants Nokogiri provides. 
With that we don't have a random number floating around which someone might ask: "why is it 70 and where is it coming from?". Plus it would still work if Nokogiri changes anything in the future.

leastbad:
You two are both right. Sometimes I let myself forget that today, only me and god understand what 70 is for. In six months, it'll just be god.

@lmatiolis do you want to fix it up? I don't want to step on your toes any further 😀

lmatiolis:
Thanks @marcoroth and @leastbad !

Yeah, no worries! I'll fix it up here.

lmatiolis:
Hi @marcoroth and @leastbad , I've rebased on master and I've included one commit to do the test changes you've added, @leastbad . I also did the constant instead of the method that I was doing before cause I think that's better.

Let me know if everything is good! I did run `bundle exec rake test_ruby` and `standardrb --fix`, so hopefully it's everything as expected.

Thanks!

lmatiolis:
Thanks @leastbad and @marcoroth , we will keep working with stimulus reflex here and hope to contribute more in the future to this awesome project! o/

PS: I'm getting the praise here, but honestly, this was all @shepmaster 's work that I just drove to the finish line! <3 

Thanks!

PR #493, opened 2021-04-29
Always pick up instance variables when accessing controller

### LINKED ISSUE #471: Lost Instance Variables
## Describe the bug

Calling controller in a before_reflex keeps instance variables set in the reflex from being accessible in the view. Discord conversation [here](https://discord.com/channels/629472241427415060/733725826411135107/821352741020368907).

## To Reproduce

Here is an MVCE to demonstrate the behavior: https://github.com/jonsgreen/stimulus_reflex_input_bug/tree/lost-instance-variables-bug. 

If you update the message comment you will see that the message is updated but the `@comment` set in the does not transfer to the view.  If you comment out: https://github.com/jonsgreen/stimulus_reflex_input_bug/blob/lost-instance-variables-bug/app/reflexes/message_reflex.rb#L4 then it does. 

## Expected behavior

Expect all instance variables set in a reflex to be available when rendering. 

### StimulusReflex

- Gem: [e.g.  3.4.1]
- Node package: [e.g. 3.4.1]

### External tools

- Ruby: [e.g. 2.6.5]
- Rails: [e.g. 6.0.3]

leastbad:

Bug fix

## Description

Changes `Reflex#controller` so that it adds any ivars to the controller instance, even if it's been previously memoized.

Fixes #471

## Why should this be added

Since `controller` is memoized, if it is called in a `before_reflex` callback, any instance variables set in the main Reflex method will not be present in the controller when the page is rendered.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -98,11 +98,12 @@ class StimulusReflex::Reflex
     @controller ||= begin
       controller_class.new.tap do |c|
         c.instance_variable_set :"@stimulus_reflex", true
-        instance_variables.each { |name| c.instance_variable_set name, instance_variable_get(name) }
         c.set_request! request
         c.set_response! controller_class.make_response!(request)
       end
     end
+    instance_variables.each { |name| @controller.instance_variable_set name, instance_variable_get(name) }
+    @controller
   end
 
   def controller?


julianrubisch:
LGTM, would be cool if @jonsgreen could cross check

jonsgreen:
@leastbad @julianrubisch I have also now confirmed that this is fixed on my MVCE. Thanks you for attending to these bugs on top of all the great feature work, documentation and support communication!

PR #494, opened 2021-04-29
Prevent install script from choking on magic comments via hard-coded line numbers

### LINKED ISSUE #429: Installer can insert session_store config in wrong place
# Bug Report

## Describe the bug

The install rake task currently makes a [hard coded assumption](https://github.com/hopsoft/stimulus_reflex/blob/master/lib/tasks/stimulus_reflex/install.rake#L54) about where to add a line in the development.rb config. If however someone has altered that file (e.g. using `Style/FrozenStringLiteralComment` with rubocop which adds magic statement `# frozen_string_literal: true` to config files) then the line ends up outside the `Rails.application.configure` block, installation does not fully complete and the app will no longer start due to this error: 

```sh
undefined local variable or method `config' for main:Object
```

This is a pretty minor issue but could cause confusion and a bad first experience with the gem. 

## To Reproduce

Add `# frozen_string_literal: true` to top of `config/environments/development.rb` and then run `rails stimulus_reflex:install` on a new rails app. 

## Expected behavior

The `config.session_store :cache_store` line should get added within the configure block. 

### StimulusReflex

- Gem: 3.4.0

### External tools

- Ruby: [e.g. 2.7.2]
- Rails: [e.g. 6.1.0]

leastbad:

Bug fix

## Description

Modify install script to add line on the first line after the `Rails.application.configure` block begins.

Fixes #429

## Why should this be added

Hard-coded line numbers could break if a magic comment is added to the file.

## Checklist

### DIFF
--- a/lib/tasks/stimulus_reflex/install.rake
+++ b/lib/tasks/stimulus_reflex/install.rake
@@ -52,7 +52,8 @@ namespace :stimulus_reflex do
     filepath = Rails.root.join("config/environments/development.rb")
     lines = File.open(filepath, "r") { |f| f.readlines }
     unless lines.find { |line| line.include?("config.session_store") }
-      lines.insert 3, "  config.session_store :cache_store\n\n"
+      matches = lines.select { |line| line =~ /\A(Rails.application.configure do)/ }
+      lines.insert lines.index(matches.last).to_i + 1, "  config.session_store :cache_store\n\n"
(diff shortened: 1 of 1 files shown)


julianrubisch:
LGTM

leastbad:
> LGTM

Let's go to Mars!

PR #498, opened 2021-05-03
render_collection

leastbad:

Enhancement

## Description

Renames the `Reflex#wrap` method to `render_collection`.

Render collection method signature now includes a single resource passed, so:

`wrap(render(@posts), @posts)` now becomes `render_collection(@posts)`

I created a temporary `wrap` method that issues a stern deprecation warning, even though technically `wrap` has never been part of a shipped release. It issues a warning that is hard to ignore and then passes to `render_collection`. We'll remove this before shipping the final v3.5.

## Why should this be added

I strongly suspect that 90% of `wrap` calls will actually receive the same parameter essentially twice. You can still call it the old way, but if you cut through to general intent, this is clearly a superior name and mechanism.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -158,4 +158,18 @@ class StimulusReflex::Reflex
     content = render(resource) unless content
     tag.div(content.html_safe, id: dom_id(resource, hash: ""))
   end
+
+  def wrap(content, resource)
+    puts "DEPRECIATION WARNING! This method has been renamed to render_collection, and the method signature is reversed:"
+    puts
+    puts "render_collection(resource, content)"
+    puts
+    puts "Or, if you were calling wrap(render(@posts), @posts), you can now just call:"
+    puts
+    puts "render_collection(@posts)"
+    puts
+    puts "We're making this message as long and obnoxious as possible because soon, we're going to exit if you're still calling wrap."
+    render_collection(resource, content)
+  end
+
 end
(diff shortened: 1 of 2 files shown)


lib/stimulus_reflex/reflex.rb
```diff
@@ -153,8 +153,9 @@ def dom_id(record, prefix = nil, hash: "#")
   # morphdom needs content to be wrapped in an element with the same id when children_only: true
   # Oddly, it doesn't matter if the target element is a div! See: https://docs.stimulusreflex.com/appendices/troubleshooting#different-element-type-altogether-who-cares-so-long-as-the-css-selector-matches
   # Used internally to allow automatic partial collection rendering, but also useful to library users
-  # eg. `morph dom_id(@posts), wrap(render(@posts), @posts)`
-  def wrap(content, resource)
+  # eg. `morph dom_id(@posts), render_collection(@posts)`
+  def render_collection(resource, content = nil)
+    content = render(resource) unless content
```
marcoroth, line 158: Shouldn't we keep `wrap` around for backwards compatibility reasons?

```suggestion
    content = render(resource) unless content
    wrap(content, resource)
  end

  def wrap(content, resource)
```

marcoroth: Oh, I just realized `wrap` hasn't even been shipped yet.

marcoroth (approved):

lib/stimulus_reflex/reflex.rb
leastbad: 😁 https://github.com/hopsoft/stimulus_reflex/pull/498/commits/d60a705d55e8af8632e9427f689dfccf8b503407

hopsoft (approved):
Looks great. I think we should remove `wrap` and the deprecation warning as its never been released.

PR #500, opened 2021-05-07
Move object_with_indifferent_access to ReflexData

erlingur:
Bug fix

## Description

When code was refactored into new class called ReflexData in 680efc1, the `object_with_indifferent_access` method was forgotten. It subsequently fails when mapping over `data["args"]` with the error:

`Reflex Users#tag failed: undefined method 'object_with_indifferent_access' for #<StimulusReflex::ReflexData:0x00007fa737e58c98>`

## Why should this be added

It should work when mapping over arguments.


leastbad:
Great catch!

PR #506, opened 2021-05-17
tab_id accessor

leastbad:

Enhancement

## Description

Reflex class gains a `tab_id` accessor to compliment `reflex_id`. It is a UUIDv4 that is generated every time SR is initialized, eg once per tab. Refreshes and different/duplicated tabs will have different values.

Part of the ClientAttributes struct.

Updated the generator template to reflect this new accessor.

## Why should this be added

Sometimes, I really wish that I had a way of uniquely identifying the browser tab which made a request without having to pack it into a data attribute. It's especially significant for an upcoming gem I'm working on.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -155,6 +155,7 @@ const register = (controller, options = {}) => {
         target,
         args,
         url,
+        tabId,
         attrs,
         dataset,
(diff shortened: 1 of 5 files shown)


marcoroth (approved):
Nice! This is handy and looks straightforward!

PR #509, opened 2021-05-23
handle operations that have string-defined selector argument

leastbad:

Bug fix

## Description

Currently, if you try to do this in a Reflex method:
```rb
cable_ready.append("#stonknotes", html: html)
```
... you get a `no implicit conversion of Symbol into Integer` exception because CableReadyChannels is hard-coded to assume that the first argument is a Hash. This is not guaranteed to be the case since we merged https://github.com/stimulusreflex/cable_ready/pull/107 (see item 3 for context).

## Why should this be added

Failures are bad.

## Checklist


Gemfile.lock
```diff
@@ -176,7 +176,7 @@ GEM
     zeitwerk (2.4.2)
 
 PLATFORMS
-  ruby
+  x86_64-linux
```
marcoroth, line 179: Do you have an idea what impact this change might have?

marcoroth (commented):
Awesome, nice catch! 

I can confirm that it's broken in `3.5.0.pre0` and that this PR fixes that issue.

Gemfile.lock
leastbad: Not a clue, unfortunately. @hopsoft?

hopsoft (approved):

PR #510, opened 2021-05-23
Improve install experience

### LINKED ISSUE #508: Sanity checker should not abort stimulus_reflex install task
# Bug Report

## Describe the bug

Upon running `rake stimulus_reflex:install`, the sanity checker module aborts with the message about: "StimulusReflex strongly suggests that you set default_url_options in your environment files."

## Expected behavior

This shouldn't happen.

### StimulusReflex

- Gem: 3.5.0.pre0
- Node package: 3.5.0-pre0

leastbad:

Bug fix

## Description

It doesn't seem like the sanity check modules receive anything in the `ARGV`, and since it's basically impossible to obtain the CLI command used to run the rake task, I suspect that the `called_by_generate_config?` method isn't doing anything.

I came up with an admittedly hacky but seemingly effective way to determine if the sanity check is being run inside of rake:

```rb
    def called_by_rake?
      caller.find { |c| c.include?("/gems/rake-") }
    end
```
I have been able to verify that installation works great on a fresh Rails 6.1 project.

I also took the opportunity to freshen up the copy offered by the warning messages. I found it a bit hard to read.

Fixes #508

Install task needs to be rock solid.

## Checklist


lib/tasks/stimulus_reflex/install.rake
```diff
@@ -57,6 +57,20 @@ namespace :stimulus_reflex do
       File.open(filepath, "w") { |f| f.write lines.join }
     end
 
+    lines = File.open(filepath, "r") { |f| f.readlines }
```
marcoroth, line 60: Is there a specific reason we call `File.open(filepath, "r") { |f| f.readlines }` all over the install script? 
It seems that `File.readlines(filepath)` is doing the same

```suggestion
    lines = File.readlines(filepath)
```

lib/tasks/stimulus_reflex/install.rake
```diff
@@ -57,6 +57,20 @@ namespace :stimulus_reflex do
...
+    unless lines.find { |line| line.include?("config.action_mailer.default_url_options") }
+      matches = lines.select { |line| line =~ /\A(Rails.application.configure do)/ }
+      lines.insert lines.index(matches.last).to_i + 1, "  config.action_mailer.default_url_options = {host: \"localhost\", port: 3000}\n\n"
+      File.open(filepath, "w") { |f| f.write lines.join }
```
marcoroth, line 64: Similarly here: 

```suggestion
      File.write(filepath, lines.join)
```

marcoroth (commented):
Nice! I really like the overall improvements done here 👍🏼

lib/stimulus_reflex/utils/sanity_checker.rb
```diff
@@ -29,6 +30,10 @@ def called_by_installer?
     def called_by_generate_config?
       ARGV.include? "stimulus_reflex:initializer"
     end
+
+    def called_by_rake?
+      caller.find { |c| c.include?("/gems/rake-") }
+    end
```
marcoroth, line 36: Hmmm. After checking a third time I'm not sure if this catches all the possibilities.

If you upgrade an app from 3.4 to 3.5.0.pre0 and try to run it, it obviously fails with the expected error message:
```
StimulusReflex strongly suggests that you set default_url_options in your environment files. [...]
```

It also suggested that you can generate an initializer to change the config.

If you then try to run the initializer generator it still fails:
```bash
> rails generate stimulus_reflex:initializer
Heads up! 🔥

StimulusReflex strongly suggests that you set default_url_options in your environment files [...]
```

lib/tasks/stimulus_reflex/install.rake
leastbad: cargo-cult-plane_497x375

lib/stimulus_reflex/utils/sanity_checker.rb
leastbad: Is there something more dramatic that we can do, like the rake equivalent of a mutex? I'm even open to temp files if it gets the job done.

Either way, I actively need help getting this over the finish line - especially because what I'm "supposed" to be doing is writing documentation. (How's that thesis coming along? 😉 )

lib/stimulus_reflex/utils/sanity_checker.rb
```diff
@@ -29,18 +30,22 @@ def called_by_installer?
     def called_by_generate_config?
       ARGV.include? "stimulus_reflex:initializer"
     end
+
+    def called_by_rake?
+      caller.find { |c| c.include?("/gems/rake-") }
```
ParamagicDev, line 35: I _think_ this will fail for windows. It appears a more bullet-proof solution is to do this:

```suggestion
      File.basename($PROGRAM_NAME) == 'rake'
```

https://stackoverflow.com/questions/2467208/how-can-i-tell-if-rails-code-is-being-run-via-rake-or-script-generate

lib/stimulus_reflex/utils/sanity_checker.rb
ParamagicDev: Based on a brief perusal of how to deal with mutexes, the best answer for long running Rake tasks is to generate a `lockfile` and then check if that lockfile exists, if not, then create a lockfile and do your business.

https://stackoverflow.com/questions/7321083/rake-check-if-already-running/7851642

lib/stimulus_reflex/utils/sanity_checker.rb
```diff
@@ -101,39 +109,41 @@ def check_new_version_available
...
+    Rails.application.config.cache_store == :null_store
   end
 
   def default_url_config_set?
-    Rails.application.config.action_controller.default_url_options
+    if defined?(ActionMailer)
+      Rails.application.config.action_controller.default_url_options && Rails.application.config.action_mailer.default_url_options
+    else
+      Rails.application.config.action_controller.default_url_options
+    end
```
ParamagicDev, line 125: This will never fire.

The reason is that `default_url_options` returns `nil` but above you check:

`default_url_config_set? == false`

These should probably be blank checks to catch empty hashes and nil. but still return a boolean.

```suggestion
    if defined?(ActionMailer)
      Rails.application.config.action_controller.default_url_options.blank? && Rails.application.config.action_mailer.default_url_options.blank?
    else
      Rails.application.config.action_controller.default_url_options.blank?
    end
```

lib/stimulus_reflex/utils/sanity_checker.rb
```diff
@@ -49,39 +54,42 @@ def check_caching_enabled
...
 
   def check_default_url_config
-    unless default_url_config_set?
-      warn_and_exit <<~WARN
-        StimulusReflex strongly suggests that you set default_url_options in your environment files.
-        Otherwise, ActionController and ActionMailer will default to example.com when rendering route helpers.
+    return if StimulusReflex.config.on_missing_default_urls == :ignore
+    if default_url_config_set? == false
+      puts <<~WARN
+        StimulusReflex strongly suggests that you set default_url_options in your environment files. Otherwise, ActionController #{ "and ActionMailer " if defined?(ActionMailer) }will default to example.com when rendering route helpers.
```
ParamagicDev, line 60: Is this `#{` supposed to be here?

ParamagicDev (commented):
In my brief attempts to break this, it held up pretty well. Default urls not being set isnt currently firing. I suggested a fix. Other than that, this feels pretty solid.

lib/stimulus_reflex/utils/sanity_checker.rb
leastbad: Yeah, I think so! It's probably a little weird looking, but the logic is that we only want to show "and ActionMailer " if ActionMailer is in the project.

Aside from being odd looking, do you see a bigger concern?

lib/stimulus_reflex/utils/sanity_checker.rb
leastbad: Is the implication that there's a difference between calling `rake db:migrate` and `rails db:migrate`?

lib/stimulus_reflex/utils/sanity_checker.rb
ParamagicDev: Yea thats what I figured, just wanted to double check just because it caught my eye.

PR #516, opened 2021-06-01
export uuidv4 function from main module

leastbad:

Enhancement *proposal*

## Description

Export the `uuidv4` function from the main module.

## Why should this be added

I'm not 100% sure this *should* be added! But my logic was that if we expose this function as a convenience feature, it might save someone who needs to generate a uuid for any reason on the client from writing their own or importing a dependency.

It could be an over-reach. What do y'all think?

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -283,6 +283,7 @@ export default {
   initialize,
   register,
   useReflex,
+  uuidv4,
   get debug () {
     return Debug.value
   },


leastbad:
@marcoroth is correct: this is off-topic. I can write it up as a tip in the docs, maybe.

marcoroth:
Yeah.. You are probably better off if you add `uuid` to your `package.json`.

```js
import { v4 as uuid } from 'uuid';

const myUuid = uuid()
```

PR #517, opened 2021-06-03
Optionally provide innerHtml and textContent on StimulusReflex::Element

julianrubisch:
# Enhancement

## Description

This PR allows to optionally transmit the `reflexElement`'s `innerHTML` and/or `textContent` when calling a reflex.

This can be done:

1. by passing `includeInnerHTML: true` or `includeTextContent: true` as an option to `this.stimulate`
OR
2. by declaring `data-reflex-include-html` or `data-reflex-include-text` on the reflex HTML element

This PR also includes a refactoring of client-side reflexData handling into its own class, analog to the Ruby side, along with tests.

## Why should this be added

This PR is inspired by working with [contenteditable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#attr-contenteditable) divs (i.e., a "Notion"-like interface).

Prior to this, you would have to put the innerHTML into a data attribute manually like this:

```js
    currentTarget.dataset.content = currentTarget.innerHTML;

    this.stimulate("ContentElementsReflex#save", currentTarget);
```

and I figured this should be taken care of at the framework level.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -141,20 +141,17 @@ const register = (controller, options = {}) => {
       const reflexData = new ReflexData(
         options,
         reflexElement,
-        controllerElement
+        controllerElement,
+        this.identifier,
+        reflexes.app.schema.reflexPermanentAttribute,
+        target,
+        args,
+        url,
+        tabId
       )
 
       const reflexId = reflexData.reflexId
 
-      const data = {
-        target,
-        args,
-        url,
-        tabId,
-        ...reflexData.valueOf(),
-        reflexController: this.identifier,
-        permanentAttributeName: reflexes.app.schema.reflexPermanentAttribute
-      }
       const { subscription } = this.StimulusReflex
 
       if (!this.isActionCableConnectionOpen())
@@ -170,7 +167,7 @@ const register = (controller, options = {}) => {
       controllerElement.reflexError = controllerElement.reflexError || {}
 
       controllerElement.reflexController[reflexId] = this
-      controllerElement.reflexData[reflexId] = data
+      controllerElement.reflexData[reflexId] = reflexData.valueOf()
 
       dispatchLifecycleEvent(
         'before',
@@ -206,7 +203,7 @@ const register = (controller, options = {}) => {
               })
 
         controllerElement.reflexData[reflexId] = {
-          ...data,
+          ...reflexData.valueOf(),
           params,
           formData
         }
@@ -214,7 +211,7 @@ const register = (controller, options = {}) => {
         subscription.send(controllerElement.reflexData[reflexId])
       })
 
-      const promise = registerReflex(data)
+      const promise = registerReflex(reflexData.valueOf())
 
       if (Debug.enabled) {
         Log.request(
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -142,6 +142,8 @@ const register = (controller, options = {}) => {
         controllerElement
       )
 
+      const reflexId = reflexData.reflexId
+
       const data = {
         target,
         args,
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -2,21 +2,13 @@ import { Controller } from 'stimulus'
 import { defaultSchema } from './schema'
 import { dispatchLifecycleEvent } from './lifecycle'
 import { uuidv4, serializeForm } from './utils'
-import { elementToXPath } from './utils'
 import { beforeDOMUpdate, afterDOMUpdate, serverMessage } from './callbacks'
-import {
-  registerReflex,
-  getReflexRoots,
-  setupDeclarativeReflexes
-} from './reflexes'
-import {
-  attributeValues,
-  extractElementAttributes,
-  extractElementDataset
-} from './attributes'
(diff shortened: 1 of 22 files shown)


leastbad:
This is really exciting! 🦖

assuntaw:
This is so useful @julianrubisch ! 

The note-taking feature I'm working on has `contenteditable` divs all over the place, and there's a lot going on in the corresponding Stimulus controllers. Not having to assign the `innerHTML` to the dataset before saving would be super helpful. Thanks for taking this on 🎉

javascript/stimulus_reflex.js
```diff
@@ -136,33 +128,30 @@ const register = (controller, options = {}) => {
...
       }
-      const attrs = options['attrs'] || extractElementAttributes(reflexElement)
-      const reflexId = options['reflexId'] || uuidv4()
-      let selectors = options['selectors'] || getReflexRoots(reflexElement)
-      if (typeof selectors === 'string') selectors = [selectors]
-      const resolveLate = options['resolveLate'] || false
-      const dataset = extractElementDataset(reflexElement)
-      const xpathController = elementToXPath(controllerElement)
-      const xpathElement = elementToXPath(reflexElement)
+
```
leastbad, line 140: This is gonna blow up #505 but I'll live. ;)

Okay, I am highly in favour of this refactor but I would ask you to "go all of the way". In #505 I added two lines after 153 that declared `const`s for `reflexController` and `permanentAttributeName` so that everything in the `data` object could just be a dual key/value thing.

I would like to expand the method signature of `ReflexData#initialize` so that it accepts `target, args, url, tabId` as well as `reflexController, permanentAttributeName`. That way, there is no longer a need to create a `data` object. We can just call getters on the new `reflexData` object you returned, in the same way that you already access `reflexId` on line 147.

The current `data` object is only referenced on lines 173, 209 and 217.

lib/stimulus_reflex/element.rb
```diff
@@ -1,15 +1,19 @@
...
   def initialize(data = {})
     @attrs = HashWithIndifferentAccess.new(data["attrs"] || {})
+    @inner_html = data["inner_html"]
+    @text_content = data["text_content"]
+
     datasets = data["dataset"] || {}
     regular_dataset = datasets["dataset"] || {}
     @data_attrs = build_data_attrs(regular_dataset, datasets["datasetAll"] || {})
-    all_attributes = @attrs.merge(@data_attrs)
+
```
leastbad, line 14: Isn't `all_attributes` referenced on line 15?

lib/stimulus_reflex/element.rb
```diff
@@ -33,6 +37,10 @@ def dataset
 
   private
 
+  def all_attributes
```
leastbad, line 40: Oh, never mind. 🤦

leastbad (changes requested):

javascript/stimulus_reflex.js
julianrubisch: Right on! Back then I didn't wanna do too much but 👍🏻

julianrubisch: Okay, done! It does feel a bit of a data clump now to me, but it's been all contracted in a single location 👍

That said, immediately `dispatchLifecycleEvent` jumped at me, which could now potentially be passed an instance of `ReflexData`. I'll take a look at that next.

leastbad: This looks really good to me. Nicely done, and thanks for the iteration.

wrt `dispatchLifecycleEvent`: it's not going to be quite so simple, as it gets called in different contexts... also, ReflexData does not include `stage` or `payload`. So I'd say that you can table this, for now at least.

leastbad (approved):

javascript/stimulus_reflex.js
julianrubisch: Yeah, I noticed that later!

julianrubisch:
We'll have to hold this back a bit, I might have found a bug around checkbox list values not being pulled correctly from attrs

julianrubisch:
Reproduced here: https://github.com/stimulusreflex/stimulus_reflex/pull/517/commits/c595fe5c4101df58342c0a61c363397c96ddb3dc

leastbad (commented):
What if you call a succession of selector/nothing Reflexes, using the same instance of an SR controller, and the dataset/selectors change between invocations?

julianrubisch:
hmmm so all I can say is this mimicks behavior as it was before. I was seeing errors because I was pulling attrs of `reflexElement` which changed its `checked` status midway (in the `setTimeout` in `stimulate`), and suddenly the wrong set of `values` was sent back to the server

In other words: the state of `ReflexData` was mutable, whereas prior to that PR it was just stored as a simple object (`const data = { .... }`) that did not mutate.

Now we're back to "as before"

leastbad:
That's a great answer, thank-you.

With that insight in mind, I amend my review to LGTM.

Each `stimulate` invocation calls `new ReflexData` so there is no risk of stale values.

julianrubisch:
Exactly!

hopsoft:
This looks great. I propose that we make the data attributes more explicit to match the stimulate args. Even though it's annoyingly verbose.

- `data-reflex-include-inner-html`
- `data-reflex-include-text-content`

PR #521, opened 2021-06-19
Add localization example to ApplicationReflex template

erlingur:

Enhancement

## Description

Needed to add full localization to all reflexes at the same time. This is an easier way than the current docs suggest, which is to create a helper method `with_locale` and wrap it around any code that needs it.

## Why should this be added

This would be helpful for anyone that needs to maintain an app that has multiple languages.

## Other

Feel free to suggest a change to the wording, English is my second language. I also went with the `before_reflex :name, def name` code style instead of `before_reflex do CODE end` style since I prefer that for my own code but you can change it if you wish.

### DIFF
--- a/lib/generators/stimulus_reflex/templates/app/reflexes/application_reflex.rb.tt
+++ b/lib/generators/stimulus_reflex/templates/app/reflexes/application_reflex.rb.tt
@@ -1,24 +1,19 @@
 # frozen_string_literal: true
 
 class ApplicationReflex < StimulusReflex::Reflex
-  # Put application-wide Reflex behavior and callbacks in this file. 
-  #
-  # Example:
-  #
-  #   # If your ActionCable connection is: `identified_by :current_user`
-  #   delegate :current_user, to: :connection
-  #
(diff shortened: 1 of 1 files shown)


leastbad:
This might be a stupid question (and I am being being lazy by not just testing it) but what happens if someone uncomments the `before_reflex` in ApplicationReflex and also declares a `before_reflex` in their Reflex class?

erlingur:
They would have two `before_reflex` callbacks. I do that in my apps. You can have multiple `before_reflex`s at the same time. I'm assuming `before_reflex` works the same as `before_action`.

leastbad:
I trust you! I just would have expected that a developer would have to call `super` to have them both fire.

erlingur:
Oh no worries. I'm not in front of my computer so I can't test it but I'm basing this on how `before_action` works. Now I want to test it to make sure :)

leastbad:
You are 100% correct. You can have multiple `before_reflex` calls. I'm not sure how I didn't consciously realize that until just now.

erlingur:
Good to know, thanks! :)

PR #522, opened 2021-06-24
Morph stimulus reflex element AKA "single element page morph"

julianrubisch:
# Feature

## Description

This PR allows to morph over the `StimulusReflex::Element` that initiated the reflex, like so:

`morph element, render(MyComponent.new(...))`

This is especially handy when
1. replacing a single partial or ViewComponent, without having to look up the ID
2. when calling `this.stimulate` with a dedicated element as a target

## Why should this be added

Often, when performing a selector morph, one is altering the state of a single piece of the UI and patching it. This PR allows to do so without explicitly stating a selector or AR model/relation, 

## ❗ Attention 
depends on https://github.com/stimulusreflex/cable_ready/pull/135

## Checklist


julianrubisch:
Failing tests are also because of the dependency on cableready/master

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
```diff
@@ -2,23 +2,17 @@
 
 module StimulusReflex
   class SelectorBroadcaster < Broadcaster
-    include CableReady::Identifiable
```
leastbad, line 5: Do we really not need this?

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
```diff
@@ -64,5 +58,26 @@ def selector?
...
+    def create_update_collection(selectors, html)
+      updates = selectors.is_a?(Hash) ? selectors : {selectors => html}
+      updates.map do |key, value|
+        Update.new(key, value, reflex)
+      end
+    end
+
+    StimulusReflex::SelectorBroadcaster::Update = Struct.new(:key, :value, :reflex) do
+      def selector
+        identifiable?(key) ? dom_id(key) : key.to_s
```
leastbad, line 73: Perhaps we could memoize this?

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
```diff
@@ -64,5 +58,26 @@ def selector?
...
+      end
+
+      def html
+        html = reflex.render(key) if key.is_a?(ActiveRecord::Base) && value.nil?
+        html = reflex.render_collection(key) if key.is_a?(ActiveRecord::Relation) && value.nil?
+        html || value
+      end
+    end.include(CableReady::Identifiable)
```
leastbad, line 81: Wow, am I ever not familiar with this syntax. Does this just include it in each instance of the Struct?

Can you not just use `include` inside the block as you would with a class?

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
```diff
@@ -64,5 +58,26 @@ def selector?
...
+    private
+
+    def create_update_collection(selectors, html)
+      updates = selectors.is_a?(Hash) ? selectors : {selectors => html}
+      updates.map do |key, value|
+        Update.new(key, value, reflex)
+      end
+    end
+
+    StimulusReflex::SelectorBroadcaster::Update = Struct.new(:key, :value, :reflex) do
```
leastbad, line 71: Why not just inline the `Struct.new` on line 67?

Not saying you should, just learning from your approach.

lib/stimulus_reflex/element.rb
```diff
@@ -29,6 +29,10 @@ def dataset
     @dataset ||= OpenStruct.new(build_underscored(data_attrs))
   end
 
+  def to_dom_id
+    "##{id}"
```
leastbad, line 33: So where does `id` come from?

And what happens if `id` isn't present? Do all ViewComponents have an `id` attribute?

leastbad (commented):
LGTM with a suggestion to memoize `update.selector` if reasonable.

I did my best to trace the logic of your refactoring; it looks good but to offer you advice I frequently don't take myself: it's easier to click Approve if you keep refactoring and feature implementation in separate PRs. 😉

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
julianrubisch: it's only needed in `Update` :-)

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
julianrubisch: yep, you can include modules at runtime

and no, can't do this with a `Struct`, and I was reluctant to make it a full class. However, if desired, I can rewrite this

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
julianrubisch: could probably do so, yes... I thought it'd better have a name attached to it?

lib/stimulus_reflex/element.rb
julianrubisch: Good point, I forgot a `raise` here. also, you're right, it's an accessor on `element`, I'm just wondering where it comes from myself 🤔

julianrubisch: actually I think it's `super build_undescored` in `StimulusReflex::Element#initialize`

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
leastbad: I find the syntax really interesting. It reminds me of function passing in JavaScript.

I suppose the main reason to do this is code readability and ease of future changes. I'm just glad this could be done within the same file. I find it generally much harder to follow when things get split across multiple files.

lib/stimulus_reflex/element.rb
leastbad: I am not aware of any reason `element` would have an `id`, given that it's not declared in this class.

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
leastbad: I don't think that's necessary?

lib/stimulus_reflex/element.rb
julianrubisch: here: https://github.com/stimulusreflex/stimulus_reflex/blob/master/lib/stimulus_reflex/reflex_data.rb#L37

and here: https://github.com/stimulusreflex/stimulus_reflex/blob/master/lib/stimulus_reflex/element.rb#L12

It works because it's an OpenStruct (I think)

leastbad: You are right!

So what are you thinking? Custom exception class like `NoIDException` or the like?

julianrubisch: Yep, I will add that!

julianrubisch:
Re splitting: yes, but at the cost at producing merge conflicts 😜

julianrubisch:
All done!

leastbad:
LGTM!

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
hopsoft, line 71: We could omit the fully qualified namespace on the `Update` struct, but I suppose it helps grokkability. 

Also, considering that we've added some methods/behavior to `Update`, I propose that it deserves a separate file and actual class definition.

hopsoft (changes requested):

lib/stimulus_reflex/broadcasters/selector_broadcaster.rb
julianrubisch: Updated, but we still need CR 5.0.0.pre2 for the tests to pass 👍

lib/stimulus_reflex/broadcasters/selector_broadcaster/update.rb
```diff
@@ -0,0 +1,19 @@
...
+  def selector
+    @selector ||= identifiable?(@key) ? dom_id(@key) : @key.to_s
+  end
+
+  def html
+    html = @reflex.render(@key) if @key.is_a?(ActiveRecord::Base) && @value.nil?
+    html = @reflex.render_collection(@key) if @key.is_a?(ActiveRecord::Relation) && @value.nil?
+    html || @value
+  end
+end
```
hopsoft, line 19: When I noted that we could drop the namespace, it was in the context of defining a struct inside the proper scope. Now that this is an actual class definition we should properly namespace it as `StimulusReflex::SelectorBroadcaster::Update`. Sorry for the confusion.

hopsoft (changes requested):

lib/stimulus_reflex/broadcasters/selector_broadcaster/update.rb
julianrubisch: of course...

PR #527, opened 2021-06-29
Specify form selector

julianrubisch:
# Enhancement

## Description

This PR adds the possibility to declare a custom form selector to be used when scooping up form parameters:

```html
<div data-reflex="..." data-reflex-form-selector="#custom_form">....</div>
```

```js
this.stimulate("MyReflex#action", this.element, { formSelector: "#custom_form" })
```

**Note:** This is based on my work on #517, the actual diff is much smaller: https://github.com/stimulusreflex/stimulus_reflex/commit/4c39aff63f7b902a684e9becbc4672595cff32e2

## Why should this be added

At the moment I'm building a `WizardReflex` where the actual form is separated from the navigation buttons/tabs, thus form params wouldn't be submitted.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -179,6 +179,7 @@ const register = (controller, options = {}) => {
         }
 
         const form =
+          reflexElement.closest(reflexData.formSelector) ||
           document.querySelector(reflexData.formSelector) ||
           reflexElement.closest('form')
 
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -2,13 +2,13 @@ import { Controller } from 'stimulus'
 import { dispatchLifecycleEvent } from './lifecycle'
 import { uuidv4, serializeForm } from './utils'
 import { beforeDOMUpdate, afterDOMUpdate, serverMessage } from './callbacks'
-import { registerReflex, setupDeclarativeReflexes } from './reflexes'
+import reflexes, { registerReflex, setupDeclarativeReflexes } from './reflexes'
 import { attributeValues } from './attributes'
+import Schema from './schema'
 import Log from './log'
 import Debug from './debug'
 import Deprecate from './deprecate'
 import ReflexData from './reflex_data'
-import reflexes from './reflexes'
 import isolationMode from './isolation_mode'
 import actionCable from './transports/action_cable'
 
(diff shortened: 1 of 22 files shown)


leastbad:
Out of the gate: there are documentation changes in here. Please, please take them out.

julianrubisch:
easy, this is going to rebased once #517 lands

leastbad:
I can't stress enough how poorly GitBook handles non-clean merges. I get super jumpy when I see things creep in...

javascript/stimulus_reflex.js
```diff
@@ -203,7 +189,9 @@ const register = (controller, options = {}) => {
             options['serializeForm'] = true
         }
 
-        const form = reflexElement.closest('form')
+        const form =
+          document.querySelector(reflexData.formSelector) ||
+          reflexElement.closest('form')
```
marcoroth, line 194:
```suggestion
        const form = reflexElement.closest(reflexData.formSelector || 'form')
```

I think we should still use `Element.closest()` to determine the form since there could be more than one form. Also since it's not mandatory to pass in an ID.

marcoroth (commented):
I like the general direction of this PR 👍🏼

julianrubisch:
> I can't stress enough how poorly GitBook handles non-clean merges. I get super jumpy when I see things creep in...

Admittedly, it's hard to reason about this way. Still I wanted to put it out there. We'll have to wait for @hopsoft to review #517

leastbad:
Essentially, every paragraph in the docs that has at least one changed character gets duplicated 2-3 times, showing a before/after... but it's all just text. You have to scan every page looking for dupes, and fussing to make sure the right version persists.

It takes hours.

javascript/stimulus_reflex.js
julianrubisch: hmm in my case this didn't work because it's not a child, but a sibling... reverting...

hopsoft (approved):
Nice work. Let's wait for #517 to land then get this merged.

javascript/stimulus_reflex.js
marcoroth: But then I would recommend to do this instead:

```suggestion
        const form =
          reflexElement.closest(reflexData.formSelector) ||
          document.querySelector(reflexData.formSelector) ||
          reflexElement.closest('form')
```

PR #528, opened 2021-06-30
Provide proxy methods truthy? and numeric on Element

julianrubisch:
# Feature

## Description

This PR introduces two proxy methods:

- `truthy?`,
- `numeric`

on `StimulusReflex` that work the same way as `signed/unsigned`: They try to coerce a dataset attribute into boolean or numeric given to it in hash notation:

```ruby
element.truthy?[:prevent_save]
element.numeric[:price]
```

## Why should this be added

Easier access to parsed/cast boolean and numeric values in the element's dataset

## Checklist


leastbad (commented):
I'm generally positive about these additions.

The only question I have is about the practical and meaningful difference between `1` and `1.0`. Since Ruby does a pretty good job of coercing floats into integers, perhaps there's nothing more to say and float is a reasonable default.

After all, if someone truly needs an integer, they certainly have the power to make that happen.

marcoroth (commented):
I think it would also make sense to provide a `#falsy?` accessor, or does something speak against that?

I'm also not too sure about the general `#numeric` accessor. Wouldn't it make more sense to provide one for integer values and one for float values?

julianrubisch:
Personally I think falling back to float is just ok.

About falsey, I don't know... I mean there's nothing wrong here but is there additional value?

hopsoft:
I like this but recommend a few changes, some minor renaming and I'd like to have the abstraction available to `dataset` also for consistent API coverage.

I'm thinking of an API more like this.

```ruby
element.boolean[:prevent_save]
element.numeric[:price]
element.dataset.boolean[:prevent_save]
element.dataset.numeric[:price]
``` 

Thoughts?

julianrubisch:
Okay with me. I remember we went back and forth regarding "truthy" and settled with it for some reason 🤔

julianrubisch:
I think the rationale was

1. naming collisions (`boolean`)
2. blank data attributes that don't explicitly have a value of `"true"`, like `checked`

And as for consistency I'm unsure ... do we actually have methods for `signed` etc on the dataset?

leastbad:
> And as for consistency I'm unsure ... do we actually have methods for `signed` etc on the dataset?

https://github.com/stimulusreflex/stimulus_reflex/blob/db9dc1df62a1ded9a37287f1f963583d247e4dcf/lib/stimulus_reflex/element.rb#L20

julianrubisch:
Sorry, on `element.dataset`, as @hopsoft suggested above

hopsoft:
I propose creating `class Dataset` as a first class citizen, then moving the `signed` and the typed value accessor implementations into `Dataset`, then have `Element` delegate to the `Dataset` methods.

PR #529, opened 2021-07-12
Piggyback Turbo(links) progress bar if it is present

julianrubisch:
# Enhancement

## Description

Uses the Turbo(links) progress bar to display activity while morphing. This should improve the perceived UX in case of slow morphs

## Open Questions

1. not sure if `Turbolinks.adapter` is the correct extension point, since I don't have an app running Turbolinks lying around. Maybe @ParamagicDev could weigh in on this?
2. how should we allow configuring it? once per app, like `isolate` ? as an option when registering a controller? per reflex/element?

## Checklist


leastbad:
Wouldn't this be perfectly suited to a snippet on the [Useful Patterns](https://docs.stimulusreflex.com/rtfm/patterns#client-side) page of the SR docs? This seems like a clever idea that doesn't need to be in the core library to be effective.

julianrubisch:
so what precisely are you worried about?

bundle size? scope?

I'm striving to increase the perceived UX for everyone at once (on an opt-in basis, of course)

leastbad:
Scope. This just feels like a "hey, here's a thing you could do if it makes sense for your app".

I appreciate that you keep saying that it's opt-in, but considering that the best location for these sorts of global event listeners is your application.js, it seems weird that you'd handle this interaction as a library feature.

It seems like this would be really good for page morphs... but selector and nothing morphs, perhaps less so? For example, if there's a flurry of morphs, or multiple morphs resolving at the same time, a single progress bar might jump around in an unfortunate way.

Anyhow, I don't want to arm wrestle on this. If you folks are into it, I won't complain.

DamnedScholar:
As someone who encountered the TL progress bar from the outside first, using sites that use Rails without knowing anything about it, I associate it with page loads. It would feel weird to tie it to an individual component updating, and most apps with long component updates (such as comment sections) have progress placeholders where the component itself is in the layout.

Makes sense for page morphs, potentially confusing/redundant if it appeared any time you run a morph that isn't instant.

julianrubisch:
Understood!

I think a `data-reflex-display-progress` attribute could do it...

julianrubisch:
Closed in favor of https://github.com/stimulusreflex/stimulus_reflex_activity

PR #530, opened 2021-07-15
Move the StimulusReflex methods that are assigned to controller

jonathan-s:
enhancement

## Description
By moving the StimulusReflex methods assigned to the controller it
decouples the way a StimulusReflex controller is registered to the
methods.

## Why should this be added
This lets a developer outside of the Rails ecosystem re-use those
methods and make their own "register" method without needing to
install actioncable.

## Checklist


leastbad:
Hey Jonathan, we have discussed #333 several times, which is very much an ongoing concern. I'm a bit distressed by this PR given that we've fully explained our roadmap and it's clear that you've put real work into this. There's currently a Twitter thread playing out which addresses this scenario: https://twitter.com/snipeyhead/status/1414976362247426051

Although I want to make it clear that I'm feeling zero of the rage, I'm feeling a lot of the rest because I know full well that if you'd checked-in before assuming we'd just forgotten to do this, I would have set you straight and presumably wouldn't be feeling like I have to be the bad guy, now. And that's awful, because you and Sockpuppet are actually super duper important to us.

Folks need to talk about big, opinionated PRs to make sure that they are wanted and appropriate.

Here's the scoop: v3.5 wasn't ever supposed to exist. There's a lot that's been put on hold due to IRL issues that aren't appropriate to go into, here. v4 is going to be a significant refactor and in some aspects, rewrite of the client and it will fully address #333. In other words, there's no chance that this PR can be accepted. If you had asked first, I would have happily clarified.

I do apologize that this process of decoupling ActionCable has taken longer to land than we would have hoped. Like everything else I'm alluding to, it's simply beyond our control.

Let's keep the conversation moving forward on Discord.

jonathan-s:
I'm sorry that I'm causing some distress to you, and there is no need to feel like a bad guy here. I'm aware that StimulusReflex moves on its own schedule and there isn't much I can do about that. 

However, the good news is that me making this PR started to get some other thoughts flowing. I think that I could unblock myself in this regard by using a forked version for the time being, and it should be relatively easy to keep that fork up to date with upstream.

leastbad:
Phew! Please let me know if/how I can be of assistance.

Have you been following the conversation about porting CR to Python on Discord?

PR #532, opened 2021-07-21
add first line of stacktrace to console.log error in dev environment

RolandStuder:
# enhancement a bit more error to debug

## Description

I usually see errors on reflexes during development in the console.log first, so it would be nice to have a hint there,
where the error is coming from.

Therefore I added the first line of the stacktrace to the error info in the console.log for convenience.

## Why should this be added

Pure convenience, only done in dev environment

## Example output

image

## Checklist


leastbad:
Sure, why not! 😉

PR #534, opened 2021-07-27
reflex render layout defaults to false

leastbad:

Enhancement

## Description

Reflex `render` method now defaults to `layout: false`.

## Why should this be added

Discord user @jsantos was having issues with AuthLogic when rendering a template using a Nothing Morph. It was determined that `render` calls initiated inside of Reflex classes should default to `layout: false`.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -111,6 +111,8 @@ class StimulusReflex::Reflex
   end
 
   def render(*args)
+    options = args.extract_options!
+    args << options.reverse_merge!(layout: false)
     controller_class.renderer.new(connection.env.merge("SCRIPT_NAME" => "")).render(*args)
   end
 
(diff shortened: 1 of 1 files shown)


lib/stimulus_reflex/reflex.rb
```diff
@@ -111,6 +111,8 @@ def controller?
   end
 
   def render(*args)
+    options = args.extract_options!
+    args << options.reverse_merge!(layout: false)
```
julianrubisch, line 115: Isn’t the bang unnecessary here?

leastbad: It's possible that something was weird in my dev env, but my memory (it's been a few weeks) is that there was no "non-bang" version of this method available.

I'd be happy to be wrong...

julianrubisch: Haha, entirely possible 🤦‍♂️

leastbad: As a smart betting man would expect, you were right!

Great catch. 👍

PR #538, opened 2021-08-15
specify env headers for page morph controller

leastbad:

Enhancement

## Description

In a Page Morph Reflex action, you can now do this:

```rb
self.headers = {"HTTP_AUTHORIZATION" => "Bearer: foo"}
```

Any headers will be merged into the `request.headers` hash.

## Why should this be added

@drnic requested the ability to pass headers into the controller so that he could integrate with JWT middleware, and this seems very useful in general.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -8,7 +8,7 @@ class StimulusReflex::Reflex
   include ActionView::Helpers::TagHelper
   include CableReady::Identifiable
 
-  attr_accessor :payload
+  attr_accessor :payload, :headers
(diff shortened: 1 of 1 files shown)


julianrubisch:
This looks ace to me!

leastbad:
The only potential gotcha is if someone does something with the `controller` accessor in a `before_reflex`, before `self.headers` is assigned.

julianrubisch (approved):

hopsoft:
Would it make sense to merge headers in the `request` method vs the `controller` method or is it invoked too early to support the use case?

leastbad:
@hopsoft started there, but it happens too early!

PR #542, opened 2021-08-30
reverse merge params into locals

leastbad:

Enhancement

## Description

ActionDispatch automatically passes the `params` accessor into views as a local, but render inside of a Reflex does not.

This is a one-liner that adds params to the locals collection, but only if there's no previous value set for `params`.

## Why should this be added

Maintain consistency and principle of least surprise.

## Checklist


Laykou:
Great! I like it

PR #558, opened 2021-11-03
When searching a controller based off a reflex name, ignore hyphens

g-gagnon:
Bug fix

## Description
Stimulus controller identifiers are using kebab-case, while Reflex class name are CamelCase, which means if the controller and reflex combo are 2 or more words the controller lookup for client callbacks will fail. When it fails, it usually defaults to the first registered controller (https://github.com/stimulusreflex/stimulus_reflex/blob/master/javascript/controllers.js#L44).
It will then fail to call the callbacks if the first controller registered is a different one.

## Why should this be added

It is not farfetched to have more than one controller in the page, and also to have a controller name being more than 1 word. This combo lead to unexpected behaviour like calling the wrong callbacks.

## Checklist

### DIFF
--- a/javascript/controllers.js
+++ b/javascript/controllers.js
@@ -37,7 +37,7 @@ const findControllerByReflexName = (reflexName, controllers) => {
 
     return (
       extractReflexName(reflexName).toLowerCase() ===
-      controller.identifier.toLowerCase()
+      controller.identifier.replace('-', '').toLowerCase()
     )
   })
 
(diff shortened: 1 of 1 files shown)


g-gagnon:
Note that this code would still most likely fail if the controller/reflex is namespaced (eg.: `user--settings` controller, and `User::SettingReflex` reflex).
To support those, I assume the `extractReflexName` method would need to be updated, or the comparison to `replace('--', '::`) potentially.

leastbad:
Great catch on this, and sorry for the delay.

I had to tweak your fix slightly to support namespaced Stimulus controllers.

PR #568, opened 2022-01-04
Updated gemspec to specific Rails gems

paul:

Enhancement

## Description

Limit gem dependencies to only the Rails gems that we directly depend upon.

## Why should this be added

Most apps don't require the full suite of Rails gems, and can avoid installing them to save CI and deploy time. This gem, however, had a dependency on the Rails meta-gem which ended up installing them all anyways.

I'm not sure how to really test this, since the dummy app uses all of Rails. I grep'd through ./lib to see what parts of Rails were being used, but that might not be comprehensive.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,6 +4,7 @@ PATH
     stimulus_reflex (3.5.0.pre8)
       actioncable (>= 5.2)
       actionpack (>= 5.2)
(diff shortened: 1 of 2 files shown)


leastbad:
Same deal: can you please pull it, run bundle and push the changes? With my apologies and graditude.

PR #575, opened 2022-03-15
Bundle for Production, Dev-Builds and prebuilt JavaScript with the gem

marcoroth:
# Type of PR

Enhancement

## Description

This PR 
- bundles the JavaScript packages analog to https://github.com/stimulusreflex/cable_ready/pull/165
- introduces dev-builds analog to https://github.com/stimulusreflex/cable_ready/pull/177
- ships the pre-built StimulusReflex JavaScript client with the Rubygem analog to https://github.com/stimulusreflex/cable_ready/pull/191.

## Why should this be added

- This brings the bundling in-line with CableReady
- Eases the support for us to support all the different JavaScript/assets pipelines that come with Rails 6/7
- Bundles the JavaScript packages upfront so it's more compatible for setups where no build-pipeline is used
- Prepares a future where the `stimulus_reflex` gem also ships a pre-built JavaScript dist file
- Switches the internal dependency from `stimulus` to `@hotwired/stimulus`
- Also introduces dev-builds for StimulusReflex via https://github.com/stimulusreflex/dev-builds
- Rewrites the JavaScript test suite to `@web/test-runner`
- Bundles the pre-built JavaScript client with the Rubygem

## Checklist


leastbad:
I suppose that reality of the longer than anticipated release cycle is that it's harder and harder to justify waiting until a future v4 to switch to `@hotwired/stimulus`.

leastbad:
If it's still possible to consider releasing an officially blessed *alternative* package that imports from `stimulus` - something like `stimulus_reflex-compat` - I think that we should at least try to do so.

marcoroth:
@leastbad I don't think there is a need for it. It's just the internal dependency which is switched over. If you are still using `stimulus` in your app you don't even notice anything 🙂

PR #590, opened 2022-06-23
refactor AC transport module to support dropped connections

leastbad:

Enhancement

## Description

Moved all ActionCable-specific logic from `stimulus_reflex.js` to the ActionCableTransport module, simplifying the logic and reducing the complexity of the API.

`stimulate` method no longer calls `ActionCable#perform`. This responsibility is now owned by the transport adapter, which exports a `enqueueReflex` method.

When `enqueueReflex` is called, it either sends the Reflex to the server, or adds it to a queue.

When the connection is accepted, any enqueued Reflexes are delivered to the server.

Warnings about disconnection are no longer needed.

## Why should this be added

StimulusReflex is more compelling if Reflexes can be sent before the AC connection is established, or while the machine is offline.

I was originally holding back this functionality for v4, but I realized that there was no good reason to do so. I've been increasingly feeling like our reliance on a valid connection was a weakness, given that standard HTTP requests do not have this constraint.

This PR essentially represents a chunk of work that would have needed to happen in v4 to support multiple transports. Specifically, there's no AC-specific stuff in the main class file any longer and the separation of responsibility is much clearer. `stimulate` is now simply the method which takes a bunch of inputs and prepares the data structure required to launch a Reflex, as well as logging and returning a Promise.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -77,15 +77,8 @@ const initialize = (
 const register = (controller, options = {}) => {
   const channel = 'StimulusReflex::Channel'
   controller.StimulusReflex = { ...options, channel }
-  ActionCableTransport.createSubscription(controller)
+  ActionCableTransport.subscribe(controller)
   Object.assign(controller, {
-    // Indicates if the ActionCable web socket connection is open.
-    // The connection must be open before calling stimulate.
-    //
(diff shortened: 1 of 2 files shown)


hopsoft (approved):
Nice improvement. 👏🏻

PR #595, opened 2022-07-16
Extract CableReadyable

julianrubisch:
# Refactoring

## Description

Extracts a `CableReadyable` module to be prepended in the `Reflex` class (I'm open to better naming options 😅 )

## Why should this be added

This PR emanated from a pairing session with @hopsoft and the there perceived need to clean up the `Reflex` class a bit. This also serves further modularization. Observe that I've also given `Broadcaster` its own `StimulusReflex::CableReadyChannels` instance, so the `prepend StimulusReflex::CableReadyable` could become an optional add-on in v4.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/cable_readyable.rb
+++ b/lib/stimulus_reflex/cable_readyable.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+require "active_support/concern"
+
+module StimulusReflex
+  module CableReadyable
+    extend ActiveSupport::Concern
+
+    prepended do
+      attr_reader :cable_ready
+    end
+
+    def initialize(*args, **kwargs)
+      super(*args, **kwargs)
+
+      if is_a? CableReady::Broadcaster
+        message = <<~MSG
+
+          #{self.class.name} includes CableReady::Broadcaster, and you need to remove it.
+          Reflexes have their own CableReady interface. You can just assume that it's present.
+          See https://docs.stimulusreflex.com/rtfm/cableready#using-cableready-inside-a-reflex-action for more details.
+
+        MSG
+        raise TypeError.new(message.strip)
+      end
+
+      @cable_ready = StimulusReflex::CableReadyChannels.new(stream_name, reflex_id)
+    end
+  end
+end
--- a/lib/stimulus_reflex/cable_readyable.rb
+++ b/lib/stimulus_reflex/cable_readyable.rb
@@ -24,7 +24,7 @@ module StimulusReflex
         raise TypeError.new(message.strip)
       end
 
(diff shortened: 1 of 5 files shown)


lib/stimulus_reflex/cable_readyable.rb
```diff
@@ -0,0 +1,30 @@
...
+
+          #{self.class.name} includes CableReady::Broadcaster, and you need to remove it.
+          Reflexes have their own CableReady interface. You can just assume that it's present.
+          See https://docs.stimulusreflex.com/rtfm/cableready#using-cableready-inside-a-reflex-action for more details.
+
+        MSG
+        raise TypeError.new(message.strip)
+      end
+
+      @cable_ready = StimulusReflex::CableReadyChannels.new(stream_name, reflex_id)
```
hopsoft, line 27: We should probably add methods for `stream_name` and `reflex_id` to indicate that anything prepending this module should implement these methods?

```ruby
def stream_name
  super # prepending modules are required to implement this
end

def reflex_id
  super # prepending modules are required to implement this
end
```

hopsoft (changes requested):

lib/stimulus_reflex/cable_readyable.rb
julianrubisch: I concur, it just seems to me that this subset of attributes is pretty arbitrary. Maybe we should just pass the reflex instance to CableReadyChannels and refactor there?

hopsoft: Agree. That's a much better idea.

julianrubisch: Done!

julianrubisch: In the future, maybe pattern matching could be of help here. I think it will really pay off wrt Morph Types

julianrubisch:
I was thinking about naming it `CableReadyPiggyBackable`, but that's a bit long ;-)

leastbad:
Clever! LGTM 👍

marcoroth:
Just throwing it out there, how about `CableRideable `?

leastbad:
English is fussy. I'm trying to think of a word that is more ready than ready, and it's like trying to achieve a blacker black.

Ironically, Readyable actually manages to convey that it's *less-ready*. But then, I also don't know if this really matters.

`CableReadyAndAble` might be the height of wit on this topic. :trollface:

julianrubisch:
`prepend CableReadiness` ?

hopsoft:
I like `CableReadiness`. Here are a few other ideas.

- CableReadyAvailable
- CableReadyOperable
- CableReadyViable

hopsoft (approved):

PR #599, opened 2022-07-29
configure operations for morph and replace

leastbad:

Feature

## Description

Allow overriding the default CableReady operations used by page and selector morphs via the initializer. By default, these values are:
```ruby
config.morph_operation = :morph
config.replace_operation = :inner_html
```
The actual options passed to these operations is unchanged; it's up to the developer to implement a [custom CableReady operation](https://cableready.stimulusreflex.com/customization#custom-operations) that handles the data coming from the server.

If you are using this feature, be sure to be aware of the options being sent to the client by the [page](https://github.com/stimulusreflex/stimulus_reflex/blob/8d7f20f29d6fe4d9b92296d50bfb696622c5b0a5/lib/stimulus_reflex/broadcasters/page_broadcaster.rb#L15) broadcaster, as well as the selector broadcaster [morph](https://github.com/stimulusreflex/stimulus_reflex/blob/8d7f20f29d6fe4d9b92296d50bfb696622c5b0a5/lib/stimulus_reflex/broadcasters/selector_broadcaster.rb#L14) and [replace](https://github.com/stimulusreflex/stimulus_reflex/blob/8d7f20f29d6fe4d9b92296d50bfb696622c5b0a5/lib/stimulus_reflex/broadcasters/selector_broadcaster.rb#L24) mechanisms.

For reference, here is the existing CableReady [morph](https://github.com/stimulusreflex/cable_ready/blob/3f016909de88639c70d5ba6efcacec6767147aa4/javascript/operations.js#L83) and [inner_html](https://github.com/stimulusreflex/cable_ready/blob/3f016909de88639c70d5ba6efcacec6767147aa4/javascript/operations.js#L47) operations, which could serve as a starting point.

## Why should this be added

1. People might want to use the [Alpine morph method](https://alpinejs.dev/plugins/morph).
2. People might want to clone the morph method but change it so that `children_only: false`. Note that this will likely interfere with normal SR operations, so consider it a sharp knife.

## Checklist

### DIFF
--- a/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
+++ b/lib/generators/stimulus_reflex/templates/config/initializers/stimulus_reflex.rb
@@ -21,6 +21,11 @@ StimulusReflex.configure do |config|
 
   # config.on_missing_default_urls = :warn
 
+  # Override the CableReady operation used for morphing and replacing content
+
+  # config.morph_operation = :morph
+  # config.replace_operation = :inner_html
+
   # Override the parent class that the StimulusReflex ActionCable channel inherits from
 
   # config.parent_channel = "ApplicationCable::Channel"
--- a/lib/stimulus_reflex/broadcasters/page_broadcaster.rb
+++ b/lib/stimulus_reflex/broadcasters/page_broadcaster.rb
@@ -10,16 +10,16 @@ module StimulusReflex
 
       selectors = selectors.select { |s| fragment.match(s).present? }
       selectors.each do |selector|
-        operations << [selector, :morph]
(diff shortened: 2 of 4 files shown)


marcoroth:
I think this looks good!

We might even be able to automatically register of the operations names if they weren't registered yet to it even easier to use. 

Sort of calling this if the operation is not already present:
```ruby
CableReady.configure do |config|
  config.add_operation_name(StimulusReflex.config.morph_operation)
end
```

leastbad:
Perhaps! I figure that this could backfire because it just creates more *stuff* you have to know for very little real benefit, since you still have to declare all of the other custom operations that you define.

PR #602, opened 2022-08-20
Call initial `setupDeclarativeReflexes()` on `document.readyState === 'complete'`

marcoroth:
## Type of PR

Bug Fix

## Description

This improves the reliability when using Importmaps, so that we can be sure that the DOM is setup properly, so the `setupDeclarativeReflexes()` function can be called in the right order.

Additionally we export the `setupDeclarativeReflexes()` so users can call this function manually if needed, i.e. if they have a specific load order/setup.

## Why should this be added

Improves the reliability when using StimulusReflex with Importmaps.

Thanks to @matheau for reporting this and providing their work-around.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -240,7 +240,7 @@ document.addEventListener('cable-ready:after-morph', afterDOMUpdate)
 window.addEventListener('load', setupDeclarativeReflexes)
 
 document.addEventListener('readystatechange', event => {
-  if (event.target.readyState === 'complete') {
+  if (document.readyState === 'complete') {
     setupDeclarativeReflexes()
   }
 })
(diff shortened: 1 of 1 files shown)


javascript/stimulus_reflex.js
```diff
@@ -239,4 +239,10 @@ document.addEventListener('cable-ready:after-inner-html', afterDOMUpdate)
 document.addEventListener('cable-ready:after-morph', afterDOMUpdate)
 window.addEventListener('load', setupDeclarativeReflexes)
```
marcoroth, line 240: we might even be able to  remove the 
```js
window.addEventListener('load', setupDeclarativeReflexes)
```

since  `readystatechange` could be enough in all cases. But I can't recall why we are using `load` on window here.

leastbad: It could just be that it's always been this way. If you feel confident that the `readystatechange` approach is rock solid, I'd be perfectly fine with you swapping out.

marcoroth: We should be good with just the `readystatechange` EventListener. `window.onload` is almost identical to `document.readyState === 'complete'` and is fired almost immediately after `window.onload`, see: https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event

leastbad:
Thank you for following up with good research!

PR #606, opened 2022-09-01
Declarative Reflexes Resiliency

### LINKED ISSUE #605: Reflex action happens twice when clicking once on a button with a Stimulus controller
# Bug Report

## Describe the bug

Clicking on a Button with a reflex action and a stimulus controller fires the reflex action twice! 

## To Reproduce

1. Clone the repo and setup - bin/setup
2. Run the server - bin/rails s
3. Open localhost:3000 with Chrome (see note below about Safari)
4. Click on the button (or the link)
5. Observe in the DevTools console or in the Server logs that the ExampleReflex#poke is called twice

### Additional notes:
* If you go to the test_controller.js, and uncomment the connect() method, then the double-fire does not happen.
* BUT if you do have connect() uncommented, then the clientside lifecycle callbacks don't fire!  (ie. beforePoke) _**second bug!**_

### Note about Safari:
* Upon refreshing the page, the button doesn't actually work, whereas the link does (after a couple of seconds)
* Interestingly, once the link fires the reflex action, the button works
* Maybe I should create another bug for this?

## Expected behavior

* Should only fire the reflex once!
* Also, the clientside lifecycle callbacks (ie. beforePoke) should fire if connect() is specified.

Link to repo: https://github.com/wheee/double-reflex-fire-importmap
(Repo was originally a clone of https://github.com/marcoroth/stimulus_reflex-importmaps)

image

## Workaround

Add a connect() method to the Stimulus controller.  Doesn't even have to do anything.

### StimulusReflex

- Gem: 3.5.0

### External tools

- Ruby: 3.1.2
- Rails: 7.0.2

### Browser

- Browser: Chrome
- Version: 104.0.5112.101 (Official Build) (arm64)

marcoroth:
# Type of PR

Improvement

## Description

This Pull Requests fixes two problems:
a) the double-fire of reflexes when invoking the reflex once
b) StimulusReflex not executing the client-side Stimulus controller callbacks when invoking a reflex

Because of load order reasons (which are mainly just apparent in Import Maps powered apps) `setupDeclarativeReflexes()` isn't always run at right time. In browsers, where the Import Maps are not natively supported yet, it leads to the fact that the `setupDeclarativeReflexes()` is ran twice. Since there is also no guarantee of load order we can't assume that all Stimulus controller instances are loaded yet.

In some instances this had weird side effects. If you have this element:
```html
<button 
  data-reflex="click->Example#create" 
  data-controller="example"
>Create</button>
```
it would convert the element to this the after calling `setupDeclarativeReflexes()` the first time around, because the `example` controller instance is not setup yet:
```diff
<button 
  data-reflex="click->Example#create" 
- data-controller="example"
+ data-controller="example stimulus-reflex" 
+ data-action="click->stimulus-reflex#__perform"
>Create</button>
```

and after calling `setupDeclarativeReflexes()` the second time with the `example` controller now available:
```diff
<button 
  data-reflex="click->Example#create" 
  data-controller="example stimulus-reflex" 
- data-action="click->stimulus-reflex#__perform"
+ data-action="click->stimulus-reflex#__perform click->example#__perform"
>Create</button>
```

The fact why this issue appeared in the first place is because the `setupDeclarativeReflexes()` is looking if the developer defined a `data-controller` on the element, and if so we are using that controller so we can invoke the callbacks in that controller. 

But at the time when the `setupDeclarativeReflexes()` function is being run the first time the `example` controller instance isn't connected yet and our function then opts for setting the generic `stimulus-reflex` controller instead. At the second time the `example` controller instance is loaded and it will append the second action descriptor, because that one wasn't present already so it causes the action descriptor to be set twice causing the double-fire issue.

Because there are now two action descriptors setup on that element it's also going to invoke the reflex twice. This Pull Requests refactors the `setupDeclarativeReflexes()` function and makes it aware of previously set action descriptors so it can replace them instead of just appending them to prevent the double-fire issue.

I guess depending on the browser used and the browser version it would not add the action descriptor twice, but instead would just use the wrong Stimulus controller and therefore not invoke the Stimulus controller callbacks. 

In order to fix that, we extract a function `setupDeclarativeReflexesForElement(element)` which can be called with a single element to setup the declarative reflexes on it. This function is being called when the Stimulus controller calls `StimulusReflex.register(this)`, which then setups the right action descriptors, guaranteeing that the right controller is actually present.

Because of this approach, we might even be able to remove the global event listener for setting up the declarative reflexes on `document`: 
https://github.com/stimulusreflex/stimulus_reflex/blob/cb62c08686b3725720cfa7cb8fa4510b3aabb616/javascript/stimulus_reflex.js#L240-L244

Fixes #605 
Follow-up of #602 

## Why should this be added

Ensures that the `data-reflex` attribute gets properly converted to it's corresponding `data-controller` and `data-action` attributes by making the conversion independent of the events dispatched on window or document, but rather by hooking into the `StimulusReflex.register(this)` function which anyway just gets called when the Stimulus controller gets connected, meaning that the right Stimulus controller instance will always be available.

## Checklist

### DIFF
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -31,7 +31,13 @@ const collectCheckedOptions = element => {
 //   attributeValue(['', 'one', null, 'two', 'three ']) // 'one two three'
 //
 const attributeValue = (values = []) => {
-  const value = Array.from(new Set(values.filter(v => v && String(v).length).map(v => v.trim()))).join(' ').trim()
+  const value = Array.from(
+    new Set(
+      values
+      .filter(v => v && String(v).length)
+      .map(v => v.trim())
+    )
+  ).join(' ').trim()
 
   return value.length > 0 ? value : null
 }
--- a/javascript/attributes.js
+++ b/javascript/attributes.js
@@ -31,12 +31,9 @@ const collectCheckedOptions = element => {
 //   attributeValue(['', 'one', null, 'two', 'three ']) // 'one two three'
 //
 const attributeValue = (values = []) => {
-  const value = values
-    .filter(v => v && String(v).length)
-    .map(v => v.trim())
-    .join(' ')
-    .trim()
-  return value.length ? value : null
+  const value = Array.from(new Set(values.filter(v => v && String(v).length).map(v => v.trim()))).join(' ').trim()
+
+  return value.length > 0 ? value : null
 }
 
 // Returns an array for the passed string value by splitting on whitespace.
--- a/javascript/reflexes.js
+++ b/javascript/reflexes.js
@@ -156,7 +156,7 @@ const setupDeclarativeReflexesForElement = (element) => {
     const controllerName = controller ? controller.identifier : 'stimulus-reflex'
 
     actions.push(`${reflexName.split('->')[0]}->${controllerName}#__perform`)
-    controllers.push('stimulus-reflex')
+    controllers.push(controllerName)
   })
 
   const controllerValue = attributeValue(controllers)
--- a/javascript/reflexes.js
+++ b/javascript/reflexes.js
@@ -136,44 +136,48 @@ const getReflexRoots = element => {
   return list
 }
 
+// Sets up declarative reflex behavior for a single element
+//
+const setupDeclarativeReflexesForElement = (element) => {
+  const controllerAttribute = element.getAttribute(Schema.controller)
+  const controllers = attributeValues(controllerAttribute)
+
+  const reflexAttribute = element.getAttribute(Schema.reflex)
+  const reflexAttributeNames = attributeValues(reflexAttribute)
+
+  const actionAttribute = element.getAttribute(Schema.action)
+  const actions = attributeValues(actionAttribute).filter(action => !action.includes("#__perform"))
+
+  reflexAttributeNames.forEach(reflexName => {
+    const controller = findControllerByReflexName(
+      reflexName,
+      allReflexControllers(reflexes.app, element)
+    )
+    const controllerName = controller ? controller.identifier : 'stimulus-reflex'
+
+    actions.push(`${reflexName.split('->')[0]}->${controllerName}#__perform`)
+    controllers.push('stimulus-reflex')
+  })
+
+  const controllerValue = attributeValue(controllers)
+  const actionValue = attributeValue(actions)
+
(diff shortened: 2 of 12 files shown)


javascript/attributes.js
```diff
@@ -31,12 +31,9 @@ const collectCheckedOptions = element => {
 //   attributeValue(['', 'one', null, 'two', 'three ']) // 'one two three'
 //
 const attributeValue = (values = []) => {
-  const value = values
-    .filter(v => v && String(v).length)
-    .map(v => v.trim())
-    .join(' ')
-    .trim()
-  return value.length ? value : null
+  const value = Array.from(new Set(values.filter(v => v && String(v).length).map(v => v.trim()))).join(' ').trim()
```
leastbad, line 34: How the heck did you get this line past prettier?

marcoroth: 🤷🏼‍♂️

leastbad: Also, what is the outcome of this change? From what I can tell, it's the same functionality as before. Is converting to a Set and then an Array just removing duplicates?

marcoroth: Exactly

marcoroth:
> Are you confident that filtering controllers on !action.includes("#__perform") will properly handle scenarios where a single element could have Reflex actions declared from multiple classes? For example input->foo#__perform click->bar#__perform.

Yeah, that should be covered, it might be, that we are invoking the function twice for such an element, but that shouldn't matter now since it should be resilient enough.

marcoroth:
We might be able to emit the StimulusReflex ready event after the initialize function finshed or after the ActionCable connection established, since we don't need to "care" about the controllers in that case

leastbad:
> We might be able to emit the StimulusReflex ready event after the initialize function finshed or after the ActionCable connection established, since we don't need to "care" about the controllers in that case

The reason that event gets emitted is to tell the developer that it's ready to process events. If we emit before the controllers have registered themselves, then the event will come before the handlers are installed. This was fine in a syncronous config...

Otherwise, if you're confident about multiple classes working fine (should we add a test for this?) LGTM.

marcoroth:
Okay I kept the global event listener with the `setupDeclarativeReflexes()` so that we still have a `stimulus-reflex:ready` being emitted with the best guess we have available.

I also added some tests for `setupDeclarativeReflexesForElement()`.

PR #610, opened 2022-09-24
StimulusReflex 3.5 installer

leastbad:

Feature

## Description

With sincere and genuine apologies for how long this has been in the oven, here is my best attempt at providing a luxe, single-command install task for SR. It builds directly on the techniques and outputs of preceding foundational efforts by both [Julian](https://github.com/stimulusreflex/cable_ready/pull/179) and [Marco](https://github.com/stimulusreflex/stimulus_reflex/tree/v3.5-installer).

It supports all major bundlers in their default configuration:

Rails 6.1: webpacker 5.4.3, shakapacker, vite
Rails 7: importmap, esbuild, shakapacker, vite

The installer is divided into steps which are designed to be either generic or highly specific to one of the bundlers.

The installer uses temporary files to maintain state; one of the major wins is that it means the setup process can fail, they can fix an issue, and task is smart enough to resume where they left off. It will also survive a reboot, which is something you cannot say about environment variables.

One of the things I'm most excited about is the new `example` the install task offers to generate. Instead of just a Reflex class and Stimulus controller, I realized that there's huge value in creating something that people can just run out of the box. This creates a controller, fully self-contained view and root. Right now, it's super minimalist, but I talk more in the Why section below about my vision for this.

I stash an MD5 hash of `Gemfile` so that we only need to run `bundle` if something changes. Likewise, I actually dynamically rebuild `package.json` based on added dependencies, development dependencies and removed dependencies. This allows us to run the dreaded, slow AF `yarn` build process 0 or 1 times, instead of 3 times since there is no way to combine these three operations on the command line.

I make efforts to create or modify most files in-place, but for files like the `{entrypoint}/controllers/index.js` where it's likely they made changes, it copies the existing file to `.bak`, generates a "perfect" new file, and then gives them a list of such replacements at the end.

```bash
rails stimulus_reflex:install # fully automatic detections, with prompts
rails stimulus_reflex:install esbuild # specify a bundler
rails stimulus_reflex:install entrypoint=app/frontend webpacker # install with webpacker 5.4.3 in app/frontend
rails stimulus_reflex:install:step action_cable yarn bundle # re-run one or more steps
rails stimulus_reflex:install:restart # wipe state and restart install

rails stimulus_reflex:install uncomment=true # uncomment any necessary includes that have been commented
rails stimulus_reflex:install spring=false # do not remove spring (equivalent to answering `n`)
rails stimulus_reflex:install example=true # create an example page and route
rails stimulus_reflex:install trace=true # debug install process with full stack trace

rails stimulus_reflex:install:step mrujs # force install mrujs
rails stimulus_reflex:install:step compression # monkey patch Action Cable to compress WS traffic with deflate

# full fat install with zero interactive prompts, suitable for scripting
rails stimulus_reflex:install esbuild entrypoint=app/javascript uncomment=true example=true
```

## Why should this be added

First, we really need to up our installation and integration game to overcome the emotional barrier people often feel when comparing us against a built-in solution. The more we can make installing SR/CR feel like ordering a Big Mac combo, the less water "golden path" arguments hold.

Second, I am convinced that one of the most powerful things we can do to advance the adoption of this stack is to put as much energy into the onboarding experience as we can, and to that end, I want to develop the built-in `example` to showcase as many core concepts and *differentiating* features as we can on one page. I'm picturing an engaging "long scroll" that alternates between live demos with code examples, short lists of compelling features and maybe even a few testimonials, like a microsite.

---

The CableReady version now lives at https://github.com/stimulusreflex/cable_ready/pull/233.

## Checklist


leastbad:
While I feel lame commenting out `StimulusReflexGeneratorTest`, I cannot figure out why the `assert_file` assertions fail, even if I remove the regex component.
```
Failure:
StimulusReflexGeneratorTest#test_creates_reflex_with_given_reflex_actions [/home/leastbad/stimulus_reflex/test/generators/stimulus_reflex_generator_test.rb:37]:
Expected file "app/javascript/controllers/user_controller.js" to exist, but does not
```
```bash
leastbad:~/foo  $ rails g stimulus_reflex demo
Resolving dependencies...
      create  app/reflexes/demo_reflex.rb
      create  app/javascript/controllers/demo_controller.js
```
So, yeah... I don't really get why this would break. However, I spent my Saturday night trying to make it work, and given the very small scope of the actual test - it tests about ~1% of the install task - I am calling it and moving on.

I very much hope someone can revisit this and show me how to avoid similar issues going forward.

lib/install/development.rb
```diff
@@ -0,0 +1,152 @@
...
+    uncomment_lines gemfile, redis_pattern, verbose: false
+    say "✅ Redis gem uncommented in Gemfile"
+  end
+  
+  # add redis-session-store to Gemfile
+  if !File.read(gemfile).match?(redis_session_store_pattern)
+    append_file gemfile, "\n# Use Redis for session storage", 
+    # rubocop:disable Lint/Syntax
+    gem "redis-session-store", "0.11.4"
+    # rubocop:enable Lint/Syntax
```
mansakondo, line 114:
```suggestion
    append_file gemfile, <<RUBY
# Use Redis for session storage",
gem "redis-session-store", "0.11.4"
RUBY
```

lib/install/development.rb
```diff
@@ -0,0 +1,152 @@
+def write_redis_recommendation(development, lines, index)
+  gemfile = Rails.root.join("Gemfile")
+  redis_pattern = /gem ['"]redis-session-store['"]/
+
+  # add redis-session-store to Gemfile, but comment it out
+  if !File.read(gemfile).match?(redis_pattern)
+    append_file gemfile, "\n# Use Redis for session storage", verbose: false
+    gem "redis-session-store", "0.11.4"
```
mansakondo, line 8:
```suggestion
    append_file gemfile, <<RUBY
# Use Redis for session storage",
gem "redis-session-store", "0.11.4"
RUBY
```

mansakondo (commented):
I'm not familiar with creating custom generators and templates, but when I tried the installer with esbuild, I got this syntax error that prevented `redis-session-store` from being added to the Gemfile:
```
rails aborted!
SyntaxError: /home/mansa/code/rails/stimulus_reflex/lib/install/development.rb:113: syntax error, unexpected string literal, expecting `do' or '{' or '('
    gem "redis-session-store", "0.11.4"
        ^
/home/mansa/code/rails/stimulus_reflex/lib/install/development.rb:113: syntax error, unexpected ',', expecting `end'
    gem "redis-session-store", "0.11.4"
```

lib/install/development.rb
```diff
@@ -0,0 +1,152 @@
...
+
+  # config.session_store :redis_session_store, {
+  #   serializer: :json,
+  #   on_redis_down: ->(*a) { logger.error("Redis down! \#{a.inspect}") },
+  #   redis: {
+  #     expire_after: 120.minutes,
+  #     key_prefix: "session:",
+  #     url: ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" }
+  #   }
+  # }
```
mansakondo, line 29: I'm also getting this error:
```
rails aborted!
ArgumentError: wrong number of arguments (given 2, expected 0..1)
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:5:in `block in <main>'
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:3:in `<main>'
/home/mansa/code/rails/sr_installer_test/config/environment.rb:5:in `<main>'
```
Because `config.session_store` expect `**options`:
```suggestion
  # config.session_store :redis_session_store,
  #   serializer: :json,
  #   on_redis_down: ->(*a) { logger.error("Redis down! \#{a.inspect}") },
  #   redis: {
  #     expire_after: 120.minutes,
  #     key_prefix: "session:",
  #     url: ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" }
  #   }
```

mansakondo, line 29:
```suggestion
  # config.session_store :redis_session_store
  #   serializer: :json,
  #   on_redis_down: ->(*a) { logger.error("Redis down! \#{a.inspect}") },
  #   redis: {
  #     expire_after: 120.minutes,
  #     key_prefix: "session:",
  #     url: ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" }
  #   }
```

lib/install/development.rb
```diff
@@ -0,0 +1,152 @@
...
+
+  config.session_store :redis_session_store, {
+    serializer: :json,
+    on_redis_down: ->(*a) { logger.error("Redis down! \#{a.inspect}") },
+    redis: {
+      expire_after: 120.minutes,
+      key_prefix: "session:",
+      url: ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" }
+    }
+  }
```
mansakondo, line 94:
```suggestion
  config.session_store :redis_session_store,
    serializer: :json,
    on_redis_down: ->(*a) { logger.error("Redis down! \#{a.inspect}") },
    redis: {
      expire_after: 120.minutes,
      key_prefix: "session:",
      url: ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" }
    }
```

mansakondo (commented):
I'm also getting this error:
```
rails aborted!
ArgumentError: wrong number of arguments (given 2, expected 0..1)
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:5:in `block in <main>'
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:3:in `<main>'
/home/mansa/code/rails/sr_installer_test/config/environment.rb:5:in `<main>'
```
Because `config.session_store` expect `**options`.

lib/install/development.rb
leastbad: Sorry about this dumb typo! I actually just pushed a commit that removes the comma. Don't even know how that snuck in there.

lib/install/development.rb
leastbad: Nice catch. What version of Ruby are you using?

lib/install/development.rb
leastbad: I'm on the fence about converting this to a heredoc, though you're probably right. It all comes down to the subtle difference in the output on the console - I like the output that running the `gem` action has.

I know, it's a small thing. 😁

lib/install/development.rb
leastbad: Dropping the (apparently superfluous) extra `{` `}` pair seems fine in 2.7 as well!

lib/install/development.rb
mansakondo: No need to apologize :)

lib/install/development.rb
mansakondo: Yeah, I feel you 🙂

lib/install/development.rb
mansakondo: Sorry for the late answer. I used Ruby 3.1.2.

leastbad: No worries! Lemme know if the latest works for you.

mansakondo: I still get the same errors with the latest changes 🤔

lib/install/development.rb
mansakondo: I still get this error with the latest changes:
```
rails aborted!
ArgumentError: wrong number of arguments (given 2, expected 0..1)
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:5:in `block in <main>'
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:3:in `<main>'
/home/mansa/code/rails/sr_installer_test/config/environment.rb:5:in `<main>'
```

mansakondo: I still get this error with the lastest changes:
```
rails aborted!
ArgumentError: wrong number of arguments (given 2, expected 0..1)
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:5:in `block in <main>'
/home/mansa/code/rails/sr_installer_test/config/environments/development.rb:3:in `<main>'
/home/mansa/code/rails/sr_installer_test/config/environment.rb:5:in `<main>'
```

leastbad: I'm a doofus sometimes. It should work, now with a commit that I just pushed.

This brings me to the `config.cache_store`. I'm wondering if it will have the same problem?

mansakondo: It works like a charm 👍🏾

I don't have any issues with the braces in `config.cache_store`. It was specific to `config.session_store` I guess.

lib/generators/stimulus_reflex/templates/app/javascript/application.js.esbuild.tt
```diff
@@ -0,0 +1,2 @@
+// Entry point for the build script in your package.json
+import "@hotwired/turbo-rails"
```
marcoroth, line 2:
```suggestion
import { Turbo } from "@hotwired/turbo-rails"
```

lib/generators/stimulus_reflex/templates/app/javascript/config/mrujs.js.tt
```diff
@@ -0,0 +1,9 @@
+import CableReady from "cable_ready"
+import mrujs from "mrujs"
+import { CableCar } from "mrujs/plugins"
+
+mrujs.start({
+  plugins: [
+    new CableCar(CableReady)
+  ]
+})
```
marcoroth, line 9: I don't think we should include mrujs in the StimulusReflex installer, since that's not part of SR's scope.

lib/generators/stimulus_reflex/templates/app/javascript/config/cable_ready.js.tt
```diff
@@ -0,0 +1,4 @@
+import consumer from '../channels/consumer'
+import CableReady from 'cable_ready'
+
+CableReady.initialize({ consumer })
```
marcoroth, line 4: This should live in the CableReady installer

lib/generators/stimulus_reflex/templates/app/javascript/config/stimulus_reflex.js.tt
```diff
@@ -0,0 +1,5 @@
+import { application } from "../controllers/application"
+import controller from '../controllers/application_controller'
+import StimulusReflex from 'stimulus_reflex'
```
marcoroth, line 3:
```suggestion
import controller from "../controllers/application_controller"
import StimulusReflex from "stimulus_reflex"
```

lib/generators/stimulus_reflex/templates/app/javascript/controllers/application.js.tt
```diff
@@ -0,0 +1,11 @@
+import { Application } from "@hotwired/stimulus"
+import consumer from '../channels/consumer'
```
marcoroth, line 2:
```suggestion
import consumer from "../channels/consumer"
```

lib/generators/stimulus_reflex/templates/app/javascript/packs/application.js.tt
```diff
@@ -0,0 +1,13 @@
...
+// that code so it'll be compiled.
+
+import Rails from "@rails/ujs"
+import Turbolinks from "turbolinks"
+import * as ActiveStorage from "@rails/activestorage"
+import "channels"
+
+Rails.start()
+Turbolinks.start()
+ActiveStorage.start()
```
marcoroth, line 13:
```suggestion
import { Turbo } from "@hotwired/turbo-rails"
import * as ActiveStorage from "@rails/activestorage"
import "channels"

ActiveStorage.start()
```

lib/generators/stimulus_reflex/templates/config/importmap.rb.tt
```diff
@@ -0,0 +1,17 @@
+# Pin npm packages by running ./bin/importmap
+
+pin "application", preload: true
+pin "@hotwired/turbo-rails", to: "turbo.min.js"
+pin "@hotwired/stimulus", to: "stimulus.min.js"
+pin "@hotwired/stimulus-loading", to: "stimulus-loading.js"
+
+pin "@rails/actioncable", to: "actioncable.esm.js"
+pin "cable_ready", to: "https://devbuilds.herokuapp.com/package/npm/cable_ready/latest"
+pin "stimulus_reflex", to: "https://devbuilds.herokuapp.com/package/npm/stimulus_reflex/latest"
```
marcoroth, line 10: We should default these URLs to the version the Gem ships, otherwise we have version mis-matches again.
```suggestion
pin "cable_ready", to: "cable_ready.min.js"
pin "stimulus_reflex", to: "stimulus_reflex.min.js"
```

lib/generators/stimulus_reflex/templates/config/importmap.rb.tt
```diff
@@ -0,0 +1,17 @@
...
+pin "morphdom", to: "https://ga.jspm.io/npm:morphdom@2.6.1/dist/morphdom.js"
+pin "mrujs", to: "https://ga.jspm.io/npm:mrujs@0.10.1/dist/index.module.js"
+pin "mrujs/plugins", to: "https://ga.jspm.io/npm:mrujs@0.10.1/plugins/dist/plugins.module.js"
```
marcoroth, line 13: This needs to be dynamic based on the user-selection, if we actullay want to include Mrujs in the installer.
```suggestion
```

lib/generators/stimulus_reflex/templates/esbuild.config.js.tt
```diff
@@ -0,0 +1,13 @@
+const path = require('path')
+const rails = require('esbuild-rails')
+
+require('esbuild')
+  .build({
+    entryPoints: ['application.js'],
+    bundle: true,
+    outdir: path.join(process.cwd(), 'app/assets/builds'),
+    absWorkingDir: path.join(process.cwd(), 'app/javascript'),
```
marcoroth, line 9: The path `'app/javascript'` needs to be dynamic based on the selected location

lib/generators/stimulus_reflex/templates/app/javascript/channels/index.js.importmap.tt
```diff
@@ -0,0 +1,22 @@
+const registeredChannels = {}
+
+// Eager load all Action Cable channels registered beneath the `under` path in the import map
+export function eagerLoadChannelsFrom (under) {
```
marcoroth, line 4: You should call the `eagerLoadChannelsFrom()` function in this file

lib/generators/stimulus_reflex/templates/app/javascript/packs/application.js.tt
```diff
@@ -0,0 +1,13 @@
+// This file is automatically compiled by Webpack, along with any other files
+// present in this directory. You're encouraged to place your actual application logic in
+// a relevant structure within app/javascript and only use these pack files to reference
+// that code so it'll be compiled.
+
+import Rails from "@rails/ujs"
+import Turbolinks from "turbolinks"
+import * as ActiveStorage from "@rails/activestorage"
```
marcoroth, line 8: We should either drop `activestorage` here are include it in all installer-environments

lib/install/config.rb
```diff
@@ -0,0 +1,100 @@
...
+pack = File.read(pack_path)
+friendly_pack_path = pack_path.relative_path_from(Rails.root).to_s
+footgun = File.read("tmp/stimulus_reflex_installer/footgun")
+cr_pattern = /import ['"].\/config\/cable_ready['"]/
+cr_commented_pattern = /\s*\/\/\s*#{cr_pattern}/
+cr_import = {
+  "webpacker" => "import \"config\/cable_ready\"\n",
+  "esbuild" => "import \".\/config\/cable_ready\"\n",
+  "importmap" => "import \"config\/cable_ready\"\n"
+}
```
marcoroth, line 35: We should introduce a `config/index.js` which includes all the configs. That way the installer also gets simpler.

I have an example of this here for esbuild:
https://github.com/fractaledmind/hottable/blob/main/app/javascript/config/index.js

That way we don't need to touch the `application.js` but can just append the simple imports, which ends up being just this:
```js
import "./controllers"
import "./config"
import "./channels"
```

lib/install/importmap.rb
```diff
@@ -0,0 +1,74 @@
...
+    copy_file(importmap_path, "#{importmap_path}.bak", verbose: false)
+    remove_file(importmap_path, verbose: false)
+    copy_file(importmap_src, importmap_path, verbose: false)
+    append_file("tmp/stimulus_reflex_installer/backups", "#{friendly_importmap_path}\n", verbose: false)
+    say "#{friendly_importmap_path} has been created"
+    say "❕ original importmap.rb renamed importmap.rb.bak", :green
+  end
+else
+  copy_file(importmap_src, importmap_path)
+end
```
marcoroth, line 20: We don't need to copy the `importmap` file into the app. Rails is smart enough to merge the pinned modules by itself, because of this:
https://github.com/stimulusreflex/stimulus_reflex/blob/master/lib/stimulus_reflex/engine.rb#L22-L26

lib/install/initializers.rb
```diff
@@ -0,0 +1,10 @@
+sr_initializer_path = Rails.root.join("config/initializers/stimulus_reflex.rb")
+cr_initializer_path = Rails.root.join("config/initializers/cable_ready.rb")
+
+generate "stimulus_reflex:initializer" unless sr_initializer_path.exist?
+say "✅ StimulusReflex initializer created at config/initializers/stimulus_reflex.rb"
+
+generate "cable_ready:initializer" unless cr_initializer_path.exist?
+say "✅ CableReady initializer created at config/initializers/cable_ready.rb"
```
marcoroth, line 8: This should live in the CableReady installer

marcoroth (commented):
I left some comments from a first pass-through of the diff

lib/generators/stimulus_reflex/initializer_generator.rb
```diff
@@ -8,7 +8,8 @@ class InitializerGenerator < Rails::Generators::Base
     source_root File.expand_path("templates", __dir__)
 
     def copy_initializer_file
-      copy_file "config/initializers/stimulus_reflex.rb"
+      initializer_path = Rails.root.join("config/initializers/stimulus_reflex.rb")
+      copy_file "config/initializers/stimulus_reflex.rb" unless initializer_path.exist?
```
marcoroth, line 12: The `copy_file` should be smart enough to figure out if it's present, and if it is it prompts the user to override it it's content is not the same

lib/generators/stimulus_reflex/templates/app/javascript/application.js.importmap.tt
```diff
@@ -0,0 +1,3 @@
+// Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
+import "@hotwired/turbo-rails"
```
marcoroth, line 2:
```suggestion
import { Turbo } from "@hotwired/turbo-rails"
```

lib/generators/stimulus_reflex/templates/app/javascript/config/stimulus_reflex.js.tt
```diff
@@ -0,0 +1,5 @@
...
+
+window.reflexes = StimulusReflex.initialize(application, { controller, isolate: true })
```
marcoroth, line 5: I still think this is confusing and redundant if you can access the same thing via `StimulusReflex.reflexes`
```suggestion
StimulusReflex.initialize(application, { controller, isolate: true })
```

lib/generators/stimulus_reflex/templates/app/javascript/controllers/index.js.webpacker.tt
```diff
@@ -0,0 +1,4 @@
+import { application } from "./application"
+import { definitionsFromContext } from '@hotwired/stimulus-webpack-helpers'
+
+application.load(definitionsFromContext(require.context('controllers', true, /_controller\.js$/)))
```
marcoroth, line 4:
```suggestion
import { definitionsFromContext } from "@hotwired/stimulus-webpack-helpers"

const controllers = definitionsFromContext(require.context("controllers", true, /_controller\.js$/))
application.load(controllers)
```

lib/install/npm_packages.rb
```diff
@@ -0,0 +1,22 @@
...
+cr_npm_version = CableReady::VERSION.gsub(".pre", "-pre")
+package_list = Rails.root.join("tmp/stimulus_reflex_installer/npm_package_list")
+package_json = Rails.root.join("package.json")
+lines = File.readlines(package_json)
+
+if !lines.index { |line| line =~ /^\s*["']cable_ready["']: ["'].*#{cr_npm_version}["']/ }
+  FileUtils.touch(package_list)
+  append_file(package_list, "cable_ready@#{cr_npm_version}\n", verbose: false)
+  say "✅ Enqueued cable_ready@#{cr_npm_version} to be added to dependencies"
+end
```
marcoroth, line 14: This should also live in the CableReady installer

lib/install/webpacker.rb
```diff
@@ -0,0 +1,128 @@
...
+  FileUtils.touch(package_list)
+  append_file(package_list, "@hotwired/stimulus-webpack-helpers@^1.0.1\n", verbose: false)
+  say "✅ Enqueued @hotwired/stimulus-webpack-helpers@^1.0.1 to be added to dependencies"
+end
+
+if !lines.index { |line| line =~ /^\s*["']webpack-dev-server["']: ["']\^3.11.3["']/ }
+  FileUtils.touch(dev_package_list)
+  append_file(dev_package_list, "webpack-dev-server@^3.11.3\n", verbose: false)
+  say "✅ Enqueued webpack-dev-server@^3.11.3 to be added to dev dependencies"
+end
```
marcoroth, line 41: Why are you not just running `yarn add [list-of-packages-here]`? Then you can be sure they are installed and that your if checks don't miss because of the version numbers

lib/install/development.rb
leastbad: Thank you **so much** for following this through. I feel genuine gratitude.

lib/generators/stimulus_reflex/templates/app/javascript/config/mrujs.js.tt
leastbad: I proposed this with the intention of starting a discussion! 💯

I speculate that you're, but it's an important point that the `mrujs` step is presented as optional, with an explanation offered. I definitely do not plan to try and sneak it in the back door, but offering folks an opinionated nudge is IMO a very Rails-y thing to do. To put the ball back in your court for a moment... would you feel differently if the prompt to swap was only presented to people who currently have `@rails/ujs` in their deps?

So, here's my pitch: as much as I'm mildly obsessed with Action Cable, there are still things that WS cannot do. `mrujs` has been the tool that has allowed me to do them anyhow. The most significant example is Devise session creation and destruction without the need for a redirect or hard refresh. Another example is setting secure cookies. I propose that `mrujs` can be used as a polyfill to give WS-powered apps the ability to leverage Action Dispatch resources an API that people have been using for a decade.

The icing on top is, of course, the CableCar plugin, which I believe is an excellent marketing flywheel. Keep in mind that the SR installer is also (in my vision, at least) a prototype for a stripped down version that will be a CR installer. I would argue that the connection to SR is still unproven, the opportunity for CR/CC is all upside. Clicking a button and receiving a CR broadcast is still amazing. I honestly believe that if CR/mrujs/CC had been available in 2019, there would have been no need for Turbo. Oh, well?

Anyhow, this is ultimately not my decision to make. People have been bundling software for a long time, and historically it's been an incredibly effective approach. Think about games and GPUs, or modems and fax software. The hanger-on gets distribution and credibility, and the host gets to show that it has an ecosystem.

Bluntly, being able to demonstrate that we have an ecosystem might be one of the most potent signals we can send to the people who refuse to understand why we bother to exist when Turbo is already in their new Rails project. As someone said on Discord, people make technology decisions largely based on their perception that other people are using it, too. We need this! So why not lift the tide for all boats?

lib/generators/stimulus_reflex/templates/app/javascript/config/cable_ready.js.tt
leastbad: My current approach is that while CR will have an installer, it will actually just be the SR installer with a limited subset of steps, and that if you run the CR installer, it will check for the SR gem and if it's present, tell you to just run the SR installer.

So... it's more of an *also*, not an instead of.

lib/generators/stimulus_reflex/templates/app/javascript/packs/application.js.tt
leastbad: I want to jam that "commit" button, but... I feel like if the Rails 6.1 `new` generator creates an app that uses Turbolinks - including the `data-turbolinks` stuff in the layout - then pushing Turbo into their project is probably not the best call. It could have subtle negative outcomes.

lib/install/development.rb
mansakondo: My pleasure. I didn't do much but I'm glad that I could help 🙂

lib/generators/stimulus_reflex/templates/config/importmap.rb.tt
leastbad: As we've established, I'm not even remotely expert with importmap, so have some patience...

I've only got links to your heroku project for the duration of the development - basically while we work on this PR. My assumption is that we'd swap out right before cutting a non-preview release. Does that make sense?

lib/generators/stimulus_reflex/templates/config/importmap.rb.tt
leastbad: Yes! Good catch; it actually *was* [right here](https://github.com/stimulusreflex/stimulus_reflex/pull/610/files#diff-fc21c7099580cece0a61b9d911dbc367a687f4384f621ab6e1512aaae1ae9a11R14) but I still have not been able to figure out how to require the `importmap-rails` actions into these scripts so that we can call `pin` etc inside of the mrujs task.

If you can help crack that code, I'll move the pins back to the mrujs step. (And that still doesn't mean we'll go the mrujs route! I just want to make sure the mrujs path can be evaluated fairly.)

lib/generators/stimulus_reflex/templates/esbuild.config.js.tt
leastbad: Good catch! Seems like this `copy_file` might need to become `template`.

I'm on it. 👍

lib/generators/stimulus_reflex/templates/app/javascript/channels/index.js.importmap.tt
leastbad: Great catch!

lib/generators/stimulus_reflex/templates/app/javascript/packs/application.js.tt
leastbad: Agreed. I'm actually leaning towards dropping it. As I progress, I'm finding myself preferring to copy in "ideal" files while backing up old versions. I'm very proud of how this part turned out, with the list of changed files at the end.

After all, it's the SR installer, not the "AI that understands everything that could be in your code".

So if they are using AS, they can make sure it's there as they merge the suggested file and the `.bak` version.

lib/install/config.rb
leastbad: I like this, and will act on it.

lib/install/importmap.rb
leastbad: Did this because I could not figure out how to access `pin` and `pin_all_from`, as I said previously in this thread and on Discord. Any help is appreciated! I would vastly prefer to just call these methods instead of copying.

lib/install/initializers.rb
leastbad: As I said earlier, it will live in the CableReady installer *also*, but the goal is for people to not run a CR installer if they are using SR. Slightly more work for us, but way less complicated for our developers.

Totally worth it.

lib/generators/stimulus_reflex/initializer_generator.rb
leastbad: Technically correct, but I actually don't want them to even have to think about this if an initializer exists. Actively trying to minimize decisions. People are likely to overwrite their existing settings by accident. (Or maybe that's just me!)

I wish `copy_file` and friends would should a proposed diff so the prompt wasn't so scary. 🤷

lib/install/webpacker.rb
leastbad: Because each run of yarn 1. takes between 10-30 seconds on my machine and 2. creates a ton of visual output, and having worked with my partner as she learns coding, the more shit a script vomits out, the less likely she is to actually read the output, which is a very bad thing.

It's not just that I'm proud of my solution (I am). I'm also confident that it's an order of magnitude faster for the person running the install. Given that we have packages being installed in multiple steps which are not all guaranteed to run, and there's no way to combine dependencies, devDependencies and removed dependencies in one command, it means we'd have to run yarn 3-7 times during the install and depending on what options they select. One yarn == 🥳

lib/generators/stimulus_reflex/initializer_generator.rb
marcoroth: They do show a diff if you press `d`

leastbad: Is the prompt a lower-case `d` with no explanation? 😉

It is an imperfect solution. Better than nothing, but intimidating and less than perfectly intuitive.

marcoroth:
```
    conflict  config/initializers/stimulus_reflex.rb
Overwrite /Users/marcoroth/Development/testapp/config/initializers/stimulus_reflex.rb? (enter "h" for help) [Ynaqdhm] h
        Y - yes, overwrite
        n - no, do not overwrite
        a - all, overwrite this and all others
        q - quit, abort
        d - diff, show the differences between the old and the new
        h - help, show this help
        m - merge, run merge tool
```

lib/install/importmap.rb
marcoroth: If you remove this could it should still just work

lib/generators/stimulus_reflex/templates/app/javascript/packs/application.js.tt
marcoroth: Same here, we should anyway not override the users `application.js` anyway. We just need to append 
```js
import "config"
import "controllers"
import "channels"
```
and then we also shouldn't need to care about `activestorage` or whatever else might be in the `application.js`

lib/generators/stimulus_reflex/templates/app/javascript/packs/application.js.tt
leastbad: Agreed; technically, what I'm doing right now is a fallback if they don't have an `application.js`.

However... there's no good reason for them to not have an `application.js` when this task is run. If it's not there, I should *bail*, not attempt to fix their broken app.

Definitely a rabbit hole. Thanks for pulling me out of it.

lib/generators/stimulus_reflex/templates/app/javascript/config/stimulus_reflex.js.tt
leastbad: I still disagree... and I'm going to go with your perspective, because this is a collective effort. 💙

lib/generators/stimulus_reflex/initializer_generator.rb
leastbad: It's really hard to summarize in words without rambling. Essentially, you are an expert developer, my experience is that you are detail-oriented and actually *read* what's on your screen, you are not intimidated by choices which are hard to reverse, and most significantly, you are not intimidated by the raw diff output.

In my direct experience, many people are genuinely intimidated by prompts like this, and are strongly inclined to accept defaults and get overwhelmed when too many things are on the screen. My partner is one of these people.

Anyhow, in this PR I have a specific behaviour and visual output goal in mind: display a standard rails-style line of output if the copy happened, otherwise just proceed - I don't want them having to answer a dozen y/n questions which ultimately are irrelevent given the goal of installing SR. I believe the end-result delivers an excellent experience with high-level levers for big questions, while making some opinionated decisions (with backups, where appropriate) for all of the small stuff that ultimately just isn't as important. It's a tradeoff, but one I'm making very consciously.

lib/install/importmap.rb
leastbad: Just to re-state: I would be thrilled to be able to add both `pin` and `pin_all_from` lines using Thor actions. I still have not been able to figure out how to call the importmap Thor actions, so I'm going to convert all of the `pin` requirements to `system "bin/importmap pin X"` calls and I guess I'll use `append_file` for the `pin_all_from` directives as there doesn't seem to be a CLI command for these.

If you know how to call `pin` from within these templates, I'd love to switch away from `system` calls.

lib/generators/stimulus_reflex/templates/config/importmap.rb.tt
leastbad: Realizing that usng your `devbuilds` service with `latest` means that if there's multiple PRs in active development, you are fighting over which branch has "control" of `latest`.

Can I suggest that you add the ability to access the latest build for a particular branch? Something like:
```ruby
pin "cable_ready", to: "https://devbuilds.herokuapp.com/package/npm/cable_ready/latest/new_installer"
# or
pin "cable_ready", to: "https://devbuilds.herokuapp.com/package/npm/cable_ready/new_installer/latest"
```

mansakondo:
> While I feel lame commenting out `StimulusReflexGeneratorTest`, I cannot figure out why the `assert_file` assertions fail, even if I remove the regex component.
> 
> ```
> Failure:
> StimulusReflexGeneratorTest#test_creates_reflex_with_given_reflex_actions [/home/leastbad/stimulus_reflex/test/generators/stimulus_reflex_generator_test.rb:37]:
> Expected file "app/javascript/controllers/user_controller.js" to exist, but does not
> ```
> 
> ```shell
> leastbad:~/foo  $ rails g stimulus_reflex demo
> Resolving dependencies...
>       create  app/reflexes/demo_reflex.rb
>       create  app/javascript/controllers/demo_controller.js
> ```
> 
> So, yeah... I don't really get why this would break. However, I spent my Saturday night trying to make it work, and given the very small scope of the actual test - it tests about ~1% of the install task - I am calling it and moving on.
> 
> I very much hope someone can revisit this and show me how to avoid similar issues going forward.

It seems like the assertions fail because of the interactive steps in the installer:
```
pry(#<StimulusReflexGeneratorTest>)> run_generator %w(demo)
=> "Where do JavaScript files live in your app? Our best guess is: app/javascript 🤔\nPress enter to accept this, or type a different path.\n> "
```

test/generators/stimulus_reflex_generator_test.rb
```diff
@@ -5,54 +5,48 @@
...
-  test "creates singular named controller and reflex files" do
-    run_generator %w[demo]
-    assert_file "app/javascript/controllers/application_controller.js"
-    assert_file "app/javascript/controllers/demo_controller.js", /Demo/
-    assert_file "app/reflexes/application_reflex.rb"
-    assert_file "app/reflexes/demo_reflex.rb", /DemoReflex/
-  end
+  # test "creates singular named controller and reflex files" do
+  #   run_generator %w[demo]
+  #   assert_file "app/javascript/controllers/demo_controller.js", /Demo/
```
mansakondo, line 14: I tried to investigate and the reason why the assertions fail is that [`assert_file`](https://api.rubyonrails.org/classes/Rails/Generators/Testing/Assertions.html#method-i-assert_file:~:text=given%20file%20exists.-,You%20need%20to%20supply%20an%20absolute%20path%20or%20a%20path%20relative%20to%20the%20configured%20destination,-%3A) expect an absolute path or a path relative to the configured destination.

So, in this case, we can use `../../app/javascript/controllers/demo_controller.js`.

However, the match assertions will still fail because the controllers are defined as anonymous classes:
```

  connect () {
    super.connect()
  }
}
```
So, we should just assert the presence of the file, I guess:
```suggestion
 assert_file "../../app/javascript/controllers/demo_controller.js"
```

leastbad: With the careful caveat that I am a junior level test comprehender, I think I have the test "working" again. You were right about the anonymous class matcher for the Stimulus classes, for sure.

In order to get this both a) working and b) not generating temporary test files in the gem's root folder, I had to modify the generator to behave differently in the test environment. This suggests that testing tasks which operate on `stdin` is hard, and there's probably a gem or technique which would allow us to obtain the same end result without blocking the thread. So I'll take full responsibility for demanding a UX that relies on such barbaric techniques.

However, part of the reason I subconsciously avoid programmatic tests is precisely that it seems like huge amounts of time and energy seem to get sucked into the toil of making tests pass, with very little accounting for actual coverage. After all, the worst bugs are the ones you cannot anticipate to test for.

Anyhow, I've always understood that you should try not to modify your source classes to make tests work, but I also recognize that having hard-coded paths makes things brittle in general, so I added the ternary to check for `Rails.env.test?` and we move on.

That said, it will take a fair bit of similar ternaries to get the main installer script into a testable state. It's not just the `stdin` prompts, but the entire system is writing files at (often) hard-coded locations. Leaving that challenge for another day.

leastbad:
59069e1 closes https://github.com/stimulusreflex/cable_ready/pull/179 and https://github.com/stimulusreflex/cable_ready/pull/180 🥳

Juanchote:
If you have already redis in your gemfile:

```
✅ redis-session-store gem has been installed

[!] There was an error parsing `Gemfile`: You cannot specify the same gem twice with different version requirements.
You specified: redis (>= 4, < 5) and redis (~> 4.7). Bundler cannot continue.

 #  from /app/Gemfile:103
 #  -------------------------------------------
 #  gem "view_component", "= 2.74"
 >  gem "redis", "~> 4.7"
 #  gem "hiredis", "~> 0.6.3"
 #  -------------------------------------------
 ```

leastbad:
Thank you, @Juanchote! I'll figure this out.

PR #614, opened 2022-12-03
introduce operations option for SR initialize

leastbad:

Bug fix + enhancement

## Description

Introduces a new option to `StimulusReflex.initialize`, `operations`, which is used to override framework default CableReady operations for *morph* and *replace* tasks. The operations will default to `morph` and `inner_html` unless specified. Note that the default *morph* task is likely to become `morphdom` before SR 3.5 ships.

```js
StimulusReflex.initialize(application, {
  controller,
  isolate: true,
  operations: {morph: 'idiomorph'}
})
```

## Why should this be added

This should have been done with #599, which is currently broken. If the default operations are changed via the SR initializer, the Reflex is successfully created and executed on the server, but the event listeners were hard-coded to operate on `morph` and `innerHtml` operations. No client-side updates are possible in this scenario.

I will go back to update the README on the previous PR to reflect the need to pass an `operations` object when this PR is merged.

## Checklist

### DIFF
--- a/javascript/stimulus_reflex.js
+++ b/javascript/stimulus_reflex.js
@@ -10,7 +10,7 @@ import ReflexData from './reflex_data'
 import IsolationMode from './isolation_mode'
 import Transport from './transport'
 import ActionCableTransport from './transports/action_cable'
-
+import { kebabize } from 'cable_ready/javascript/utils'
 import { reflexes } from './reflexes'
 import { dispatchLifecycleEvent } from './lifecycle'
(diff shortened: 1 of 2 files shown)


marcoroth:
Looking at https://github.com/stimulusreflex/stimulus_reflex/pull/599, we came to the conclusion that this is not needed anymore. If people want to use a morph library other than `morphdom` then they can configure it on the server-side via the `morph_operation` option:

```ruby
  config.morph_operation = :idiomorph
end
```

I'm going to close this one for now.

PR #618, opened 2022-12-18
Handle Table-less TR and TD

cmer:

Bug Fix

## Description

Nokogiri dismisses "invalid" HTML, but since we often need to parse incomplete HTML, this causes issues.

Fixes # (issue)

## Why should this be added

When trying to morph a `tr` element, `innerHTML` is replaced instead because the `tr` node is not found in the parsed document (because it gets discarded).

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -124,6 +124,8 @@ GEM
       net-protocol
       timeout
     nio4r (2.5.8)
+    nokogiri (1.13.10-arm64-darwin)
+      racc (~> 1.4)
     nokogiri (1.13.10-x86_64-darwin)
       racc (~> 1.4)
     nokogiri (1.13.10-x86_64-linux)
@@ -204,6 +206,7 @@ GEM
     zeitwerk (2.5.4)
 
 PLATFORMS
+  arm64-darwin-21
   x86_64-darwin-19
   x86_64-linux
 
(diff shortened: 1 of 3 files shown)


leastbad:
We intentionally upgraded to `Nokogiri::HTML5` as it is *dramatically* better. The chances of us reverting are very low.

Can you please provide a markup example of what element you are attempting to morph, as well as the string that you're trying to morph it with?

cmer:
Here's the issue, as discussed on Discord: https://discord.com/channels/629472241427415060/733725826411135107/1053876547628580894 and  https://discord.com/channels/629472241427415060/733725826411135107/1053876547628580894

cmer:
The HTML snippet try to send back is:

```html
<tr id="foo"><td></td><td></td></tr>
```

The problem is that, technically, this is not valid HTML5 by itself. It requires a `table` element. Nokogiri removes `tr` and `td` elements from the document if a `table` is not part of the HTML. This leads to `morph` replacing the `innerHTML` content rather than using MorphDOM.

leastbad:
Still thinking about this.

1. Can you morph a `td`, or is it the same problem? Academic perhaps, just curious.
2. Could you rework your solution to morph the whole table? Obviously not ideal, just thinking of solves.

I'm within spitting distance of getting a pre10 shipped, at which point I can dig into this issue (and the focus gotchas) with gusto.

cmer:
> 1. Can you morph a `td`, or is it the same problem? Academic perhaps, just curious.

No, same thing. A `td` without a `tr` and `table` in invalid.

> 2. Could you rework your solution to morph the whole table? Obviously not ideal, just thinking of solves.

For what I'm doing, it'd be a PITA, but it's definitely possible.

PR #620, opened 2022-12-29
Update manifest file when `stimulus_reflex` generator used for repo's using bundlers other than importmaps

binarygit:
enhancement

## Description
Currently the `stimulus_reflex` generator does not update the ~~`app/javascript/controllers/index.js.bak`~~ `app/javascript/controllers/index.js` file which needs to get updated every time a new stimulus controller is generated _for other build systems expect `importmaps`_.

This PR runs the `stimulus:manifest:update` command to update the ~~`app/javascript/controllers/index.js.bak`~~  `app/javascript/controllers/index.js` file, when the `stimulus_reflex` generator is used.

Fixes # (issue)

## Why should this be added
It should be added because it saves users from running the `manifest:update` command after using the `stimulus_reflex` generator.

## Checklist

### DIFF
--- a/lib/generators/stimulus_reflex/stimulus_reflex_generator.rb
+++ b/lib/generators/stimulus_reflex/stimulus_reflex_generator.rb
@@ -41,6 +41,7 @@ class StimulusReflexGenerator < Rails::Generators::NamedBase
 
     template(reflex_src, reflex_path) unless options[:skip_reflex]
     template(stimulus_controller_src, stimulus_controller_path) unless options[:skip_stimulus]
+    rails_command "stimulus:manifest:update" unless Rails.root.join("config/importmap.rb").exist?
 
     if file_name == "example"
       controller_src = fetch("/app/controllers/examples_controller.rb.tt")


leastbad:
Thank you for this!

Since nobody can know everything, you've identified something I don't know anything about: the relationship between Stimulus controllers and this `index.js.bak` file. I don't know what that file does; can you explain it like this is my first day?

Is this an issue that only impacts esbuild?

When does this issue become important? Is it a production thing?

My experience has been that I can create arbitrary Stimulus controllers and the `yarn watch` process running picks them up immediately. This is why this PR catches me off guard... I've never even heard of the `stimulus:manifest:update` task!

To be 100% clear, I suspect that you're correct. I just don't see the big picture, yet.

binarygit:
Hey @leastbad :wave: :smile: 
(How the turn tables...I teach you now instead you teach me huhaha :laughing:)

Sorry, looks like I was mistaken and it's the `index.js` not the `index.js.bak` file that is updated.

I'm not sure on the _why_ other bundlers except `importmaps` need to have their `index.js` file updated for your stimulus controller to work. I think it's something to do with how those bundlers work.
I actually found a small section on the `stimulus handbook` the tells us that expect `importmaps` other bundlers don't have support for controller auto-loading, and that you must register your controller to `index.js` to make it work. (Also that using the `stimulus generator` command runs `stimulus:manifest:update` for you so that you don't need to do it manually when you use it)
Here's the link: https://stimulus.hotwired.dev/handbook/installing#using-other-build-systems

But like you I do remember when using webpacker I didn't have to do this, and if you scroll just above the section I linked to, it details how to enable live-reloading for webpacker. I suspect it was setup by default before in rails, it doesn't seem to be the case with Rails 7 because I tried it with a new repo with webpacker, and the `index.js` file has to get updated for the 
stimulus_controller to work.
(looks like the `manifest:update` command works with webpacker and other bundlers, and not only with `esbuild` so I'll update the PR description.)

Hope I was of help here. Don't kill me for my `huhaha` above :D

leastbad:
Could you do me a favour and, in a Rails 7+esbuild project where you have installed StimulusReflex with the new installer... could you just create a Stimulus controller? That is, create a file in `app/javascript/controllers` called `test_controller.js` that looks like this:
```js

  connect() {
    super.connect()
    console.log("hello world")
  }
}
```
I have a (strong) hunch that it will just work, and it's because of the StimulusReflex installer creating an `app/javascript/controllers/index.js` that doesn't suck.

binarygit:
I did that and tested in a `Rails 7.0.4` app that uses `esbuild` as bundler.
The "hello" is not logged to the console until I run the `stimulus:manifest:update` command.

leastbad:
Yeah, to be clear: this won't work unless you use the StimulusReflex installer.

Does this mean we should close this ticket?

binarygit:
Yup we should close this ticket

PR #636, opened 2023-02-22
Don't add `data-controller` to reflex element if parent already holds instance

marcoroth:
# Type of PR

Bug fix

## Description

There's was an edge case where we would add another `data-controller` attribute to the reflex element even if a parent element already had a matching controller instance.

This markup: 
```html
  <button 
    data-reflex="click->Example#increment"
  >Click</button>
</div>
```
would get transformed to:
```diff
<div data-controller="example">
  <button 
    data-reflex="click->Example#increment" 
+   data-controller="example" 
+   data-action="click->example#__perform"
  >Click</button>
</div>
```
Which ends up with two controller instances of `example` even if we actually just needed one.

This pull request optimizes this by just conditionally adding the `example` controller to the `data-controller` attribute on the reflex element.

Now with this pull request the end-result is just:
```diff
<div data-controller="example">
  <button 
    data-reflex="click->Example#increment" 
+   data-action="click->example#__perform"
  >Click</button>
</div>
```

## Why should this be added

We shouldn't create more controller instances than necessary. In this case it was apparent because the `connect()` callback was fired twice. To not fire the method twice we add the condition.

## Checklist

### DIFF
--- a/javascript/test/reflexes.setupDeclarativeReflexesForElement.test.js
+++ b/javascript/test/reflexes.setupDeclarativeReflexesForElement.test.js
@@ -170,7 +170,7 @@ describe('scanForReflexesOnElement', () => {
     assert.equal(button.dataset.controller, 'example2')
   })
 
-  it('should', async () => {
+  it('should not add data-controller to reflex element if parent holds', async () => {
     App.app.register('example', ExampleController)
 
     const controllerElement = await fixture(html`
--- a/javascript/test/reflexes.setupDeclarativeReflexesForElement.test.js
+++ b/javascript/test/reflexes.setupDeclarativeReflexesForElement.test.js
@@ -8,17 +8,17 @@ import { initialize } from '../stimulus_reflex'
 import App from '../app'
 import { scanForReflexesOnElement } from '../scanner'
 
+function registeredControllers () {
+  return Array.from(App.app.router.modulesByIdentifier.keys())
+}
+
 describe('scanForReflexesOnElement', () => {
   beforeEach(() => {
     initialize(application)
   })
 
   afterEach(() => {
-    const registeredControllers = Array.from(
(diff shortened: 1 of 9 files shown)


hopsoft (approved):

javascript/test/reflexes.setupDeclarativeReflexesForElement.test.js
```diff
@@ -106,4 +106,151 @@ describe('scanForReflexesOnElement', () => {
...
+    scanForReflexesOnElement(button)
+
+    assert.equal(controllerElement.dataset.controller, 'example1')
+
+    assert.equal(button.dataset.reflex, 'click->Example#click')
+    assert.equal(button.dataset.action, 'click->example2#__perform')
+    assert.equal(button.dataset.controller, 'example2')
+  })
+
+  it('should', async () => {
```
julianrubisch, line 173: misses a description?

marcoroth: Whoops

arambert:
I want to emphasize the importance of this PR, indeed if we want to add some Stimulus controls (on `connect`, targets, actions on elements, etc.) the fact that the component controller is duplicated on each reflex element creates wweird errors ans makes it almost impossible. The only sane solution is be to create a second stimulus controller, which feels wrong to me.

I'd really like to be able to have only one controller!

arambert:
Hi,

Are you planning to finalize & merge this PR?
Do you need any help?

Thanks

marcoroth:
Hey @arambert, yeah I wanted to get this merged soon! I've been exploring some other edge cases, which I wanted to address in the pull request. But either way, I'm going to finish this one as-is! 

Thanks for your patience! 🙏🏼

PR #641, opened 2023-03-05
Improve DX around version mismatch warnings

marcoroth:
# Type of PR 

Enhancement

## Description

This pull request enhances the developer experience when the StimulusReflex client- and server-side versions mismatch. It will show a toast message in the lower right corner of the screen if there's a version mismatch. Previously it would just show it in the Rails server log.

It will only show the toasts if you try to trigger a reflex action, it won't complain earlier, which is also much friendlier, because in previous versions of StimulusReflex it wouldn't even let you start the Rails server.

This is what happens now if you try to trigger a reflex:

* on `:exit` it will show a red error toast message and won't execute the trigged reflex:
Screenshot 2023-03-13 at 15 43 43

* on `:warn` it will show a yellow warning toast message but will still execute the triggered reflex:
Screenshot 2023-03-13 at 15 43 28

* on `:ignore` it won't show anything and also will also regularly execute the triggered reflex. 

Note: the `toastify-js` dependency will be bundled up with the built JavaScript version of the package, so it won't show up as a runtime dependency in applications running the `stimulus_reflex` npm package.

The toasts are currently not disabled in production environments, but I'm also not sure if we should. Any opinions on this?

## Why should this be added

Improves the developer experience and makes it easier to noticed/work with.

## Checklist

### DIFF
--- a/javascript/version_checker.js
+++ b/javascript/version_checker.js
@@ -1,0 +1,75 @@
+import Toastify from "toastify-js/src/toastify-es.js"
+import CableReady from "cable_ready"
+
+CableReady.operations.stimulusReflexVersionMismatch = operation => {
+  const levels = {
+    "info": {},
+    "success": { background: "rgb(0, 176, 155)" },
+    "warn": { background: "rgb(247, 187, 60)" },
+    "error": { background: "rgb(255, 95, 109)" },
+  }
+
+  const defaults = {
+    selector: setupToastify(),
+    close: true,
+    duration: 30 * 1000,
+    gravity: "bottom",
+    position: "right",
+    newWindow: true,
+    style: levels[operation.level || "info"]
+  }
+
+  Toastify({ ...defaults, ...operation }).showToast()
+}
+
+function setupToastify() {
+  const id = "stimulus-reflex-toast-element"
+  let element = document.querySelector(`#${id}`)
+
+  if (!element) {
+    element = document.createElement("div")
+    element.id = id
+    document.documentElement.appendChild(element)
+
+    const styles = document.createElement("style")
+    styles.innerHTML = `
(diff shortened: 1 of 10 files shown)


javascript/version_checker.js
```diff
@@ -0,0 +1,75 @@
+import Toastify from "toastify-js/src/toastify-es.js"
+import CableReady from "cable_ready"
+
+CableReady.operations.stimulusReflexVersionMismatch = operation => {
+  const levels = {
+    "info": {},
+    "success": { background: "rgb(0, 176, 155)" },
+    "warn": { background: "rgb(247, 187, 60)" },
+    "error": { background: "rgb(255, 95, 109)" },
```
andrewmcodes, line 9: Can you make sure this color is accessible with the text? I use a bookmarklet called tota11y.

andrewmcodes (commented):
someone other than me will want to give an approval bc this is the first PR I've looked at in awhile but only one thing came to mind. this is pretty slick 👏 Whish rails had done this with webpacker.

hopsoft:
Still need to review the code, but this is lovely.

hopsoft (approved):
👏🏻 Love it! I'm not the biggest fan of adding more dependencies but agree that it's worth it in this case to enhance DX given that it's so minimal.

javascript/version_checker.js
marcoroth: Good call, will update it, thank you!

marcoroth: I think this should be a lot better!
Screenshot 2023-03-13 at 15 43 43
Screenshot 2023-03-13 at 15 43 28

julianrubisch (approved):

PR #644, opened 2023-03-13
fix: Generator file name references

julianrubisch:
# Bug Fix

## Description

Fixed the generator's references to template files, which were broken.

I haven't currently tested this against the installer, though.

## Checklist

### DIFF
--- a/lib/generators/stimulus_reflex/templates/app/reflexes/%file_name%_reflex.rb.tt
+++ b/lib/generators/stimulus_reflex/templates/app/reflexes/%file_name%_reflex.rb.tt
@@ -66,7 +66,6 @@ class <%= class_name %>Reflex < ApplicationReflex
   <%- actions.each do |action| -%>
   def <%= action %>
   end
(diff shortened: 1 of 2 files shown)


lib/generators/stimulus_reflex/templates/app/reflexes/%file_name%_reflex.rb.tt
```diff
@@ -66,7 +66,6 @@ class <%= class_name %>Reflex < ApplicationReflex
   <%- actions.each do |action| -%>
   def <%= action %>
   end
-  <%= "\n" unless action == actions.last -%>
```
marcoroth, line 69: @julianrubisch did you remove this by accident?
Screenshot 2023-03-13 at 15 19 35

julianrubisch: actually no, because it resulted in a dangling indentation for the last `end`.

might have to have another look at that

PR #663, opened 2023-05-29
Fix exception in ActionCable channel due to reflexes overwriting each other's data

alexander-makarenko:
# Bug fix + refactoring

## Description

Think of a page with multiple StimulusReflex controllers hooked up to the `stimulus-reflex:connected` event and the markup (Haml in this example) looking something like this:

```haml
%div{ data: { controller: "sandwich", action: "stimulus-reflex:connected@document->sandwich#eat" } }

%div{ data: { controller: "tea", action: "stimulus-reflex:connected@document->tea#drink" } }
```

The actions are backed by simple JavaScript controllers and reflex actions looking as follows. The reflex actions have different arity: one take no arguments, the other takes one.

```js
// sandwich.js:

eat() {
  this.stimulate("Sandwich#eat");
}

// tea.js:

drink() {
  this.stimulate("Tea#drink", "green");
}
```

```ruby
# sandwich_reflex.rb:

class SandwichReflex < ApplicationReflex
  def eat
    # ...
  end
end

# tea_reflex.rb:

class TeaReflex < ApplicationReflex
  def drink(flavour)
    # ...
  end
end
```

This kind of setup creates a situation where `StimulusReflex::Channel#receive` is invoked twice in a very short period of time. And since both messages are processed by the **same instance** of the ActionCable channel (as can be verified by printing  out `self.object_id` at the beginning of the method), the use of an instance variable (`@reflex_data`) in the current implementation leads to a situation where it gets overrwritten by the second message before the first one has been processed. As a result, `reflex_data` inside `#delegate_call_to_reflex` ends up containing the same set of data which, after being checked against the method signature of both of the reflex actions, raises an `ArgumentError` due to arity mismatch.

This pull request fixes the issue by eliminating the use of instance variables inside `StimulusReflex::Channel#receive` in favour of bundling the message payload with the respective instance of the reflex. It also simplifies the reflex factory and constructor by delegating most of the calls to the `StimulusReflex::ReflexData` instance.

## Why should this be added

In modern web applications, it is common to load parts of the page asynchronously for a more responsive user experience. With StimulusReflex, it is most easily achieved by hooking up to the `stimulus-reflex:connected` event. This pull request fixes an exception that prevented the use of the specified event with multiple reflexes.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -3,20 +3,6 @@
 require "stimulus_reflex/cable_readiness"
 require "stimulus_reflex/version_checker"
 
-# TODO remove xpath_controller and xpath_element for v4
-ClientAttributes = Struct.new(
-  :id,
-  :tab_id,
-  :reflex_controller,
-  :xpath_controller,
-  :xpath_element,
-  :permanent_attribute_name,
-  :version,
-  :npm_version,
-  :suppress_logging,
-  keyword_init: true
-)
-
 class StimulusReflex::Reflex
   prepend StimulusReflex::CableReadiness
   include StimulusReflex::VersionChecker
@@ -26,26 +12,35 @@ class StimulusReflex::Reflex
   include CableReady::Identifiable
 
   attr_accessor :payload, :headers
-  attr_reader :channel, :url, :element, :selectors, :method_name, :broadcaster, :client_attributes, :logger
-
-  alias_method :action_name, :method_name # for compatibility with controller libraries like Pundit that expect an action name
+  attr_reader :channel, :data, :broadcaster
 
   delegate :connection, :stream_name, to: :channel
   delegate :controller_class, :flash, :session, to: :request
   delegate :broadcast, :broadcast_halt, :broadcast_forbid, :broadcast_error, to: :broadcaster
-  # TODO remove xpath_controller and xpath_element for v4
-  delegate :id, :tab_id, :reflex_controller, :xpath_controller, :xpath_element, :permanent_attribute_name, :version, :npm_version, :suppress_logging, to: :client_attributes
 
-  def initialize(channel, url: nil, element: nil, selectors: [], method_name: nil, params: {}, client_attributes: {})
+  # TODO remove xpath_controller and xpath_element for v4
+  delegate :url,
+           :element,
+           :selectors,
+           :method_name,
+           :form_params,
+           :id,
+           :tab_id,
+           :reflex_controller,
+           :xpath_controller,
+           :xpath_element,
+           :permanent_attribute_name,
+           :version,
+           :suppress_logging,
+           to: :data
+  # END TODO: remove
+
+  alias_method :action_name, :method_name # for compatibility with controller libraries like Pundit that expect an action name
+
+  def initialize(channel, data:)
     @channel = channel
-    @url = url
-    @element = element
-    @selectors = selectors
-    @method_name = method_name
-    @params = params
-    @client_attributes = ClientAttributes.new(client_attributes)
+    @data = data
     @broadcaster = StimulusReflex::PageBroadcaster.new(self)
-    @logger = suppress_logging ? nil : StimulusReflex::Logger.new(self)
     @payload = {}
     @headers = {}
 
@@ -60,6 +55,10 @@ class StimulusReflex::Reflex
     id
   end
   # END TODO: remove
+
+  def logger
+    StimulusReflex::Logger.new(self) unless suppress_logging
+  end
 
   def request
     @request ||= begin
@@ -91,7 +90,7 @@ class StimulusReflex::Reflex
       req = ActionDispatch::Request.new(env)
 
       # fetch path params (controller, action, ...) and apply them
-      request_params = StimulusReflex::RequestParameters.new(params: @params, req: req, url: url)
+      request_params = StimulusReflex::RequestParameters.new(params: form_params, req: req, url: url)
       req = request_params.apply!
 
       req
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -12,21 +12,21 @@ class StimulusReflex::Reflex
   include CableReady::Identifiable
 
   attr_accessor :payload, :headers
-  attr_reader :channel, :data, :broadcaster
+  attr_reader :channel, :reflex_data, :broadcaster
 
   delegate :connection, :stream_name, to: :channel
   delegate :controller_class, :flash, :session, to: :request
   delegate :broadcast, :broadcast_halt, :broadcast_forbid, :broadcast_error, to: :broadcaster
 
   # TODO remove xpath_controller and xpath_element for v4
-  delegate :url, :element, :selectors, :method_name, :form_params, :id, :tab_id, :reflex_controller, :xpath_controller, :xpath_element, :permanent_attribute_name, :version, :suppress_logging, to: :data
+  delegate :url, :element, :selectors, :method_name, :form_params, :id, :tab_id, :reflex_controller, :xpath_controller, :xpath_element, :permanent_attribute_name, :version, :suppress_logging, to: :reflex_data
   # END TODO: remove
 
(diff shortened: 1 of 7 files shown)


lib/stimulus_reflex/reflex.rb
```diff
@@ -61,6 +56,10 @@ def reflex_id
   end
   # END TODO: remove
 
+  def logger
+    StimulusReflex::Logger.new(self) unless suppress_logging
```
Matt-Yorkley, line 60: Isn't it preferable to keep the logger instance memoized on the reflex like it was before, instead of re-initializing it on each log call? Is there a specific issue this change is addressing?

Matt-Yorkley:
Nice refactor :heart:

lib/stimulus_reflex/reflex.rb
```diff
@@ -26,26 +12,35 @@ class StimulusReflex::Reflex
...
-  alias_method :action_name, :method_name # for compatibility with controller libraries like Pundit that expect an action name
+  attr_reader :channel, :data, :broadcaster
 
   delegate :connection, :stream_name, to: :channel
   delegate :controller_class, :flash, :session, to: :request
   delegate :broadcast, :broadcast_halt, :broadcast_forbid, :broadcast_error, to: :broadcaster
+
   # TODO remove xpath_controller and xpath_element for v4
-  delegate :id, :tab_id, :reflex_controller, :xpath_controller, :xpath_element, :permanent_attribute_name, :version, :npm_version, :suppress_logging, to: :client_attributes
+  delegate :url,
```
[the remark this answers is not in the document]
alexander-makarenko: I ended up putting them on a single line, as was suggested by the StandardRB linter used in this project.

lib/stimulus_reflex/reflex.rb
alexander-makarenko: Thanks, I have added memoization. The reason for moving it out of `#initialize` is complexity reduction.

alexander-makarenko:
@Matt-Yorkley Thanks for your review. I have addressed your comments and also fixed the tests and linter warnings. Would appreciate if you could give it another look when you have time.

lib/stimulus_reflex/reflex.rb
```diff
@@ -91,7 +77,7 @@ def request
       req = ActionDispatch::Request.new(env)
 
       # fetch path params (controller, action, ...) and apply them
-      request_params = StimulusReflex::RequestParameters.new(params: @params, req: req, url: url)
+      request_params = StimulusReflex::RequestParameters.new(params: form_params, req: req, url: url)
```
Matt-Yorkley, line 80: I'm pretty sure this should be `params: data.params` now instead of `params: form_params` and I think the delegation to `form_params` in the class definition is not needed.

alexander-makarenko:
@Matt-Yorkley Thanks for your suggestions! I have addressed all feedback.

lib/stimulus_reflex/reflex.rb
```diff
@@ -32,8 +32,6 @@ def initialize(channel, reflex_data:)
     @headers = {}
 
     check_version!
-
-    params
```
Matt-Yorkley: I think this line actually needs to stay in rather than being dropped, it's calling the `params` _method_ of this class (around line 145) which is (confusingly) doing something completely unrelated to the `params` property of the `reflex_data` object.

alexander-makarenko: Well, the method is still there and is called on L117. But what was the point of calling it inside `#initialize`?

I agree that generally speaking, it is confusing to see two sets of params referenced inside this class. However, that is how it also worked before my changes, and I did not dare refactoring that part due not having enough understanding of the code around the request. If you have ideas how to refactor it safely, please share your thoughts and I'll try to fix it.

alexander-makarenko:
@Matt-Yorkley How can we proceed? Is there anything preventing this PR from getting merged?

marcoroth:
Hey @alexander-makarenko, thanks so much for working on this!

Do you have an app or a test case which we could use to reproduce the issue itself? Just to make sure we are not re-introducing this by accident again in the future, even if it's a manual test.

I just want to make sure we are not breaking other things with this change, as there are quite a few changes here.

alexander-makarenko:
@marcoroth Since it's a race condition, it's actually quite difficult to reproduce, since it depends on how fast the StimulusReflex code runs in a specific application on a specific system. In a larger production application where I originally discovered the issue, it failed about once every 50 page loads. In a scenario that I outlined in the description of this PR (with no logic in the JS controllers or reflex), it seems to fail even rarer on my computer.

In order to debug the issue, I remember that I cloned the repo of StimulusReflex 3.5.0.pre9 locally and pointed my application to use it. Then I added a sleep somewhere down `StimulusReflex::Channel#receive` to emulate slow processing of messages, until I was able to reproduce the issue consistently. I then fixed the implementation of said method to eliminate the issue and submitted my changes in this pull request.

Thus, due to the nature of the issue, I unfortunately won't be able to provide a test application or add an automated test for it.

lib/stimulus_reflex/reflex.rb
```diff
@@ -26,32 +12,26 @@ class StimulusReflex::Reflex
...
-    @client_attributes = ClientAttributes.new(client_attributes)
+    @reflex_data = reflex_data
     @broadcaster = StimulusReflex::PageBroadcaster.new(self)
-    @logger = suppress_logging ? nil : StimulusReflex::Logger.new(self)
     @payload = {}
     @headers = {}
 
     check_version!
-
-    self.params
```
RolandStuder, line 54: I don't have time to make a reproduction, but I believe that removing the `self.params` call will bring back a bug, where the params, may not be set once the `request` reaches the action controller. 

Calling the self.params ensures that the params are properly set in the request object / headers when passing the request to the ActionController to rerender the current action. see https://github.com/stimulusreflex/stimulus_reflex/pull/249

lib/stimulus_reflex/reflex_data.rb
```diff
@@ -4,7 +4,7 @@ class StimulusReflex::ReflexData
   attr_reader :data
 
   def initialize(data)
-    @data = data
+    @data = data.deep_transform_keys { |k| k.to_s.underscore }.with_indifferent_access
```
RolandStuder, line 7: Hm, what is the motivation behind making this with indifferent access?

RolandStuder (commented):
Absolutely great catch, I really think think this should be merged, the bug you uncovered is the kind that can haunt people for weeks if not months, we should prevent such pains!

I did not make a thorough review, but noticed one line removal, that may reintroduce an existing but. I don't think it's covered by a test. See my comment

lib/stimulus_reflex/reflex_data.rb
alexander-makarenko: I think it's generally nicer and more idiomatic to work with snake case symbols rather than camel case strings in Ruby.

https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#snake-case-symbols-methods-vars
https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#symbols-as-keys

lib/stimulus_reflex/reflex.rb
alexander-makarenko: I appreciate your feedback, but I could not find any confirmation of your theory in the code.

I have [updated a test](https://github.com/stimulusreflex/stimulus_reflex/pull/663/files#diff-7e418500360dd99316b785f57d0ac7ef7d79c17d855fcefe2da575cd0d8d2492R36-R44) to remove unnecessary stubbing and better exercise the code around request parameters. If I were to print out `reflex.controller.request.parameters` there right before the first assertion, it would output `{"controller"=>"test", "action"=>"index", "a"=>"1", "b"=>"2", "c"=>"3"}`, which confirms that all parameters do get passed along to the controller instance without the `self.params` in the constructor.

RolandStuder: @alexander-makarenko ah you are right, I made a quick test app, and it indeed worked. 

Request params are a bit weird, if you set the at the wrong moment they won't persist as the request memoizes them. But apprently this new implementation works around that.

So all is good! Thank you!

marcoroth (approved):
Thank you so much for this @alexander-makarenko!

Issue #659, opened 2023-04-27
Controller over reflex attribute is ignore and reflex is always called with default StimulusReflexController


Laykou:
# Bug Report

## Describe the bug

Using the custom `data-controller` above `data-reflex` is ignored and reflex is always executed with default `stimulus-reflex` controller.

This breaks behavior and callbacks described in https://docs.stimulusreflex.com/guide/lifecycle.html#callback-methods

It may be related to: https://github.com/stimulusreflex/stimulus_reflex/pull/636

Discussed in Discord: https://discord.com/channels/629472241427415060/733725826411135107/1100724714760044606

## To Reproduce

```html
<div data-controller="example">
  <a href="#" data-reflex="Example#masticate">Eat</a>
</div>
```

```js
import ApplicationController from './application_controller.js'

export default class extends ApplicationController {
  beforeReflex(element) {
    console.log('Called here');
  }
}
```

In the end the element has `<a href="#" data-reflex="Example#masticate" data-action="stimulus-reflex#__perform">Eat</a>`  tag. Console log is not shown

## Expected behavior

Element should have `<a href="#" data-reflex="Example#masticate" data-action="example#__perform">Eat</a>` tags

## Screenshots or reproduction

- pre10 - where it worked
  ![image](https://user-images.githubusercontent.com/2313005/234763525-b0ca16bf-cefe-4d03-b94a-8e4c527e801a.png)

- rc1 - where it's not working
  ![image](https://user-images.githubusercontent.com/2313005/234763547-9fe13dd9-965c-4992-8b4b-040cd8f3e6f7.png)

## Versions

- pre9 - worked
- pre10 - worked
- rc1 - not working
- rc2 - not working

Laykou:
In our case the hotfix is creating a method in JS `data-action="my-controller#myMethod"` and call `this.stimulate('Reflex#action', event.target)` within it manually.

Laykou:
@julianrubisch Is this still an open topic or was this issue somehow addressed/fixed in the latest changes in the past month?

I'd like to know if it's worth upgrading the lib or if we still have to use workaround via custom actions `data-action` + `this.stimulate`

marcoroth:
I need to double check, it might be related #636, but I think it's more related to an unhandled edge case with setting up the declarative reflex attributes.

Laykou:
Is there some new version/release I could try this on?

marcoroth:
Looks like I can't reproduce this on rc2. 

[...]

PR #670, opened 2023-07-14
Don't choose first controller if no matching StimulusReflex-enabled controller was found

marcoroth:
# Type of PR

Bug Fix

## Description

This pull request solves an edge-case for scenarios when StimulusReflex tries to detect the right controller to use for performing the StimulusReflex `__perform` controller action for executing reflexes.

For example, if an element had a `data-controller` and a `data-reflex` attribute, and the controller in `data-controller` was a StimulusReflex-enabled controller, but didn't match the Reflex specified in the `data-reflex` there were some cases where it wouldn't add the default `stimulus-reflex` controller to the element, but would still reference the `stimulus-reflex` controller in the `data-action` attribute. This basically lead to a no-op for the reflex action.

Like:
```html
<a
  data-controller="
    some-sr-controller
  "
  data-reflex="click->Example#doSomething"
></a>
```

was expanded to this in some cases:
```diff
<a
  data-controller="
    some-sr-controller
  "
  data-reflex="click->Example#doSomething"
+ data-action="click->stimulus-reflex#__perform"
></a>
```

Whereas it should have been:
```diff
<a
  data-controller="
    some-sr-controller
+   stimulus-reflex
  "
  data-reflex="click->Example#doSomething"
+ data-action="click->stimulus-reflex#__perform"
></a>
```
With this, the `data-action` is not referencing an undefined Stimulus controller anymore and the reflex should execute again.

## Why should this be added

For consitancy reasons, so that the declarative reflex syntax still continues to make sense and work as expected.

Maybe this is even fixing @Laykou's issue in #659 which I couldn't reproduce.

Thanks to @foliosus for helping to debug this! 

## Checklist

### DIFF
--- a/javascript/scanner.js
+++ b/javascript/scanner.js
@@ -28,7 +28,7 @@ const scanForReflexesOnElement = (element, controller = null) => {
   )
 
   reflexAttributeNames.forEach(reflexName => {
-    const potentialControllers = [controller].concat(allReflexControllers(element))
+    const potentialControllers = [controller].concat(allReflexControllers(element)).filter(c => c)
 
     controller = findControllerByReflexName(
       reflexName,
@@ -42,7 +42,7 @@ const scanForReflexesOnElement = (element, controller = null) => {
 
     const parentControllerElement = element.closest(`[data-controller~=${controllerName}]`)
 
-    if (!parentControllerElement) {
+    if (!parentControllerElement || (parentControllerElement && controllerName === 'stimulus-reflex')) {
       controllers.push(controllerName)
     }
   })
--- a/javascript/scanner.js
+++ b/javascript/scanner.js
@@ -28,7 +28,7 @@ const scanForReflexesOnElement = (element, controller = null) => {
   )
 
   reflexAttributeNames.forEach(reflexName => {
-    const potentialControllers = [controller].concat(allReflexControllers(element)).filter(c => c)
+    const potentialControllers = [controller].concat(allReflexControllers(element))
 
     controller = findControllerByReflexName(
       reflexName,
@@ -42,7 +42,9 @@ const scanForReflexesOnElement = (element, controller = null) => {
 
     const parentControllerElement = element.closest(`[data-controller~=${controllerName}]`)
 
-    if (!parentControllerElement || (parentControllerElement && controllerName === 'stimulus-reflex')) {
+    const elementPrevisoulyHadStimulusReflexController = (element === parentControllerElement && controllerName === 'stimulus-reflex')
+
+    if (!parentControllerElement || elementPrevisoulyHadStimulusReflexController) {
       controllers.push(controllerName)
     }
   })
--- a/javascript/controllers.js
+++ b/javascript/controllers.js
@@ -43,7 +43,7 @@ const findControllerByReflexName = (reflexName, controllers) => {
     return identifier === controller.identifier
   })
 
-  return controller || controllers[0]
+  return controller
 }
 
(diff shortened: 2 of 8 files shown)


Laykou:
@marcoroth Shouldn't it be expanding into `data-action="click->some-sr-controller#__perform"`?

marcoroth:
@Laykou it could, but I feel like it's weird if it "leaks" the lifecycle callbacks into a controller which potentially is not related to the reflex itself.

Maybe maybe it should? 

The issue could also be resolved if we do this instead, yeah:
```html
<a
  data-controller="
    some-sr-controller
  "
  data-reflex="click->Example#doSomething"
></a>
```

was expanded to this in some cases:
```diff
<a
  data-controller="
    some-sr-controller
  "
  data-reflex="click->Example#doSomething"
+ data-action="click->some-sr-controller#__perform"
></a>
```

Laykou:
How would this work then? https://docs.stimulusreflex.com/guide/lifecycle.html#callback-methods

having 

```html
  <a href="#" data-reflex="Example#masticate">Eat</a>
</div>
```

I believe it **must** expand to

```haml
  <a href="#" data-reflex="Example#masticate" data-action="click->example#__perform">Eat</a>
</div>
```

in order to calll callbacks in `example_controller.js`, right?

```js
import ApplicationController from './application_controller.js'

  beforeReflex(anchorElement) {
    const { reflex } = anchorElement.dataset
    if (reflex.match(/masticate$/)) anchorElement.innerText = 'Eating...'
    if (reflex.match(/defecate$/)) anchorElement.innerText = 'Pooping...'
  }
}
```

Otherwise how would this `beforeReflex` from `example_controlller.js` be called if it expands only to generic `data-action="click->stimulus-reflex#__perform"` ?

marcoroth:
Yeah, this is currently how it works. It will also look on any ancestor to find a matching StimulusReflex-enabled controller. 
```html
  <a href="#" data-reflex="click->Example#masticate" data-action="click->example#__perform">Eat</a>
</div>
```

Only if the reflex specified in `data-reflex` doesn't match a StimulusReflex-enabled controller it will add the `stimulus-reflex` controller.

mbajur:
Just for the record - this PR also solved a case in which i had a stimulus_reflex button with plain stimulus controller attached to it and thanks to that fact - adding `stimulus-reflex` actions were not triggered because `stimulus-reflex` controller was not being attached to the element alongside with the original controller.

javascript/scanner.js
```diff
@@ -42,7 +42,9 @@ const scanForReflexesOnElement = (element, controller = null) => {
 
     const parentControllerElement = element.closest(`[data-controller~=${controllerName}]`)
 
-    if (!parentControllerElement) {
+    const elementPrevisoulyHadStimulusReflexController = (element === parentControllerElement && controllerName === 'stimulus-reflex')
```
cmer, line 45: You have a typo here!

marcoroth: Thanks, fixed in: https://github.com/stimulusreflex/stimulus_reflex/commit/6348953fb75299f696eab5dd4590e20699bf8c78

Laykou:
@marcoroth May I get back to this - is it possible to override the default `stimulus-reflex` e.g. when I want to use `my-custom-base-stimulus-reflex` controller instead?

I'd like to setup som **default** `beforeReflex` and `afterReflex` actions (like always adding spinner, if element is a button, etc...) and it would help a lot to be able to define this globally

PR #673, opened 2023-10-08
Fix `reflex_data` keyword argument

brunoenten:
bug fix

## Description
Rename named parameter, and change error message that was causing an exception in a rescue block.

## Why should this be added

Stimulus reflex doesn't work at all without these fixes.

## Checklist

- [X ] My code follows the style guidelines of this project
- [ X] Checks (StandardRB & Prettier-Standard) are passing

### DIFF
--- a/lib/stimulus_reflex/reflex.rb
+++ b/lib/stimulus_reflex/reflex.rb
@@ -12,21 +12,22 @@ class StimulusReflex::Reflex
   include CableReady::Identifiable
 
   attr_accessor :payload, :headers
-  attr_reader :channel, :data, :broadcaster
+  attr_reader :channel, :reflex_data, :broadcaster
 
   delegate :connection, :stream_name, to: :channel
   delegate :controller_class, :flash, :session, to: :request
   delegate :broadcast, :broadcast_halt, :broadcast_forbid, :broadcast_error, to: :broadcaster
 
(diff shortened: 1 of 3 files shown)


lib/stimulus_reflex/reflex.rb
```diff
@@ -12,21 +12,21 @@ class StimulusReflex::Reflex
   include CableReady::Identifiable
 
   attr_accessor :payload, :headers
-  attr_reader :channel, :reflex_data, :broadcaster
+  attr_reader :channel, :data, :broadcaster
```
marcoroth, line 15: Thanks for taking this on @brunoenten!

Can we instead rename the keyword argument from `data:`  to `reflex_data:` here instead of renaming it throughout the `Reflex` class:

https://github.com/stimulusreflex/stimulus_reflex/blob/3137a0f36a27dc26ee271eee5c17c3a96dce280d/lib/stimulus_reflex/reflex_factory.rb#L13-L15

This way the internal API doesn't change we also don't have to update the currently failing tests.

Thank you!

brunoenten: Done! I added an alias for data, with the same goal of impacting as little existing code as possible.

marcoroth (approved):
Thank you so much @brunoenten!

PR #674, opened 2023-11-08
Properly handle HTML fragments with self closing tags

### LINKED ISSUE #652: Document fragment now do not handle 2 siblings inputs
# Bug Report

with the latest version if I have two sibling inputs (in the erb file):
```
<div>
  <div id="label-container">
    <label>
      <input type="file" accept="image/*" />
      <input type="hidden" value="" />
    </label>
  </div>
  <div id="after-label"></div>
</div>
```

## Describe the bug

after parsing with Nokogiri.parse instead of Nokogiri::HTML5 it drops the second input and "move"  the closing label tag outside the original div.

## To Reproduce

you can add this test in `document_fragment_test`

```
test "should properly parse two siblings input" do
    raw_html = "<div><div id=\"label-container\"><label><input type=\"file\" accept=\"image/*\"><input type=\"hidden\" value=\"\"></label></div><div id=\"after-label\"></div></div>"
    fragment = StimulusReflex::HTML::DocumentFragment.new(raw_html)
    assert_equal 2, fragment.document_element.at_css('input').children.size
    assert_equal "<div><div id=\"label-container\"><label><input type=\"file\" accept=\"image/*\"><input type=\"hidden\" value=\"\"></label></div><div id=\"after-label\"></div></div>", fragment.to_html.squish
    assert_equal "<div><div id=\"label-container\"><label><input type=\"file\" accept=\"image/*\"><input type=\"hidden\" value=\"\"></label></div><div id=\"after-label\"></div></div>", fragment.outer_html.squish
end
```

## Expected behavior

parse correctly the input tags

### StimulusReflex

- Gem: [3.5.0.rc2]
- Node package: [N/A]

### External tools

- Ruby: [3.1.2]

arambert:

Bug Fix

## Description

Properly handle HTML fragments including self closing tags like `img`, `br`, `input` that Nokogiri XML parser would normally 
try to fix by adding a closing tag. For example, `<div><span>test</span></div>` would be changed internally to `<div><span>test</span></img></div>` by Nokogiri and then converted to `<div></div>` by Nokogiri through the `to_html` method. 

This is due to the use of `Nokogiri` parser for the selector morphs (as opposed to `Nokogiri::HTML5::Document` parser), but we cannot use the `Nokogiri::HTML5::Document` when handling incomplete HTML documents (see https://github.com/stimulusreflex/stimulus_reflex/commit/69cb070ec961dd52d34ec9a66a88f895686100c1 for more info)

One could say that we should produce valid html with `/` at the end of self-closing tags (ie: `<br/>` instead of `<br>`) but Nokogiri itself strips the trailling `/` when producing html string from XML nodes...

This PR detects these instances and handle them gracefully.

Fixes #652 (issue)

## Why should this be added

The main issue for the user before this PR is that html tags are sometimes removed from the reflex response because of this behaviour.

This PR allows Stimulus Reflex users to apply selector morphs with HTML containing self-closing tags with siblings.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -123,6 +123,8 @@ GEM
     net-smtp (0.3.3)
       net-protocol
     nio4r (2.5.8)
+    nokogiri (1.15.3-arm64-darwin)
+      racc (~> 1.4)
     nokogiri (1.15.3-x86_64-darwin)
       racc (~> 1.4)
     nokogiri (1.15.3-x86_64-linux)
@@ -202,6 +204,7 @@ GEM
     zeitwerk (2.6.8)
 
 PLATFORMS
+  arm64-darwin-23
   x86_64-darwin-19
   x86_64-darwin-22
   x86_64-linux
--- a/lib/stimulus_reflex/html/document.rb
+++ b/lib/stimulus_reflex/html/document.rb
@@ -28,15 +28,11 @@ module StimulusReflex
       end
 
       def parse_html(html)
-        parsing_class.parse(html.to_s)
+        Nokogiri::HTML5::Document.parse(html.to_s)
       end
 
       def empty?
         @document.content.empty?
-      end
-
-      def parsing_class
-        Nokogiri::HTML5::Document
       end
 
       def match(selector)
--- a/lib/stimulus_reflex/html/document.rb
+++ b/lib/stimulus_reflex/html/document.rb
@@ -21,10 +21,14 @@ module StimulusReflex
       end
 
       def initialize(html)
-        @document = parsing_class.parse(html.to_s)
+        @document = parse_html(html)
         @matches = {
           "body" => Match.new(@document.at_css("body"))
         }
+      end
+
+      def parse_html(html)
+        parsing_class.parse(html.to_s)
       end
 
       def empty?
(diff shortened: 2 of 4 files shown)


marcoroth:
Hey @arambert, thanks for opening this pull request!

We've been experimenting on how to properly handle the issue you outlined here. I have some pending bugs I want to report to Nokogiri which should make most of the workarounds unnecessary.

One of them is here already: https://github.com/sparklemotion/nokogiri/issues/3023

Before merging this, please allow me make sure we can make this work in upstream Nokogiri, so we don't need to work around this in StimulusReflex.

arambert:
Hi @marcoroth,
Thank you for your review!

I agree with you on the fact that this issue is originally a Nokogiri issue. Of course ideally it would be fixed on Nokogiri's side but in the meantime it's also a StimulusReflex issue ;) 
For my use case, I've monkeypatched StimulusReflex on our app and it works fine, so I'm not in a hurry.

Thanks again!

marcoroth:
@arambert just wanted to tag you and mention that https://github.com/stimulusreflex/stimulus_reflex/pull/696 should be solving this issue! If you get a chance, I'd love to hear how it works for you! Thank you!

PR #688, opened 2024-02-25
Fix ReflexData parsing

Matt-Yorkley:
# Bug Fix

## Description

The data in `StimulusReflex::ReflexData` was recently changed to transform all keys to underscored keys without preserving the original kebab-case keys (like `data-controller`), but this causes breaking changes elsewhere in the codebase. Both sets of keys need to be preserved but this was missed by the relevant tests because the test input was using a plain hash instead of instantiating a ReflexData object (see updated test setup).

## Why should this be added

The previous change caused issues like missing attributes in `element.dataset` in Reflexes. It's in `main`, but it's not released yet.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/reflex_data.rb
+++ b/lib/stimulus_reflex/reflex_data.rb
@@ -4,7 +4,7 @@ class StimulusReflex::ReflexData
   attr_reader :data
 
   def initialize(data)
-    @data = data.deep_transform_keys { |k| k.to_s.underscore }.with_indifferent_access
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Thank you @Matt-Yorkley! I'm wondering if there's an easy way how we can test this in a full-stack setup so test could catch another regression!

Matt-Yorkley:
I had the same thought, a couple of full-stack tests would be great.

PR #703, opened 2024-05-16
Add page which describes the "redisless" setup

### LINKED ISSUE #684: `solid_cache` seems to work as well as Rails cache store
# Feature Request

The setup guide suggest that you need to use Redis for caching and ideally also for `ActionCable`, plus the gem has a dependency on Redis. We deploy our applications on Windows machines and simply do not have Redis available.

I never used `stimulus_reflex` before, but worked on a first PoC this afternoon in said app, and it seems everything works fine when using `solid_cache` as Rails cache store. Is this observation correct? If so, the setup guide could be updated and maybe even the dependency of `stimulus_reflex` on `redis` could be dropped.

(I also used the `async` adapter for `ActionCable` instead of Redis - I know, not recommended in production, but again, constraints. App runs on Windows and against MSSQL, so PostgreSQL adapter is also no option).

No.

Happy if I can circumvent my way around Redis.

## Additional context

No additional context needed.

## PR link

No PR so far.

andyundso:

Documentation

## Description

This PR adds a new page which describes how to setup StimulusReflex without Redis. I extracted it into a separate page because I felt it was too much to rewrite all the existing documentation, which relies on Redis. I assume most people will use Redis, and the "redisless" configuration likely is a niche thingy.

Fixes #684 

## Why should this be added

Good to know for people that StimulusReflex can work without Redis.

## Checklist

### DIFF
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -74,6 +74,7 @@ export default {
         items: [
           { text: "Working with Events", link: "/appendices/events" },
           { text: "Testing (WIP)", link: "/appendices/testing" },
+          { text: "StimulusReflex without Redis", link: "/appendices/no-redis" },
           { text: "Deployment", link: "/appendices/deployment" },
           { text: "Troubleshooting", link: "/appendices/troubleshooting" },
           { text: "Release History", link: "/appendices/release-history" },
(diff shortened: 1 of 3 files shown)


marcoroth (approved):
Thank you @andyundso! 🙏🏼

PR #705, opened 2024-05-16
Add Thor require for `StimulusReflex::Installer`

juna-nb:
## Description

Rake tasks were failing in projects because the required Thor dependency was not always required, depending on the way it was called.

The fix also needed to explicitly require the thor gem in the installer file that used it as a mixin.

Fixes https://github.com/stimulusreflex/stimulus_reflex/issues/701

## Why should this be added

Rake tasks in Rails applications were no longer working with the latest version of this Gem.

## Checklist

### DIFF
--- a/lib/stimulus_reflex/installer.rb
+++ b/lib/stimulus_reflex/installer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "thor"
+
 module StimulusReflex
   class Installer
     include Thor::Base


juna-nb:
Just to note this change actually broke my app for some reason so this PR obviously still needs work. I'm not sure why it broke yet and this still blocks my deploys.

julianrubisch (changes requested):
Hm, I'm not sure if an explicit dependency on `Thor` is necessary and sensible, since `railties` itself depends on it. I think just a `require "thor"` should do it?

marcoroth (approved):
Thank you, @juna-nb! 🙌🏼

PR #726, opened 2024-11-23
work on rails 8 and future versions

sara-g:
Installing stimulus reflex on Rails 8.0 was failing, this change will get it to install.


marcoroth (commented):
Hey @sara-g, thanks for this!

Would you mind commit the `Gemfile.lock` too, thank you!

marcoroth (approved):
Thanks @sara-g

Issue #698, opened 2024-04-26
Refactor away instances of OpenStruct


marcoroth:
StimulusReflex relies on OpenStruct in both the [Dataset](https://github.com/stimulusreflex/stimulus_reflex/blob/3a5111e085b21d1e628813e2cee72e978acb257a/lib/stimulus_reflex/dataset.rb#L5) and the [Element](https://github.com/stimulusreflex/stimulus_reflex/blob/3a5111e085b21d1e628813e2cee72e978acb257a/lib/stimulus_reflex/element.rb#L6) models. We should stop using OpenStruct altogether as its [use is discouraged](https://github.com/ruby/ostruct/blob/master/lib/ostruct.rb#L136).

This was initially raised in #695  and the workaround was addressed by #697, but we should implement the suggested longterm solution proposed in #695 which is to remove all uses of OpenStruct in the gem.

Issue #695, opened 2024-04-11
OpenStruct no longer supported by json


TanookiMario:
# Bug Report

## Describe the bug

[Rubocop v1.63.0](https://github.com/rubocop/rubocop/releases/tag/v1.63.0) updates [json to v2.7.2](https://github.com/flori/json/releases/tag/v2.7.2). `json` v2.7.2 [makes OpenStruct optional](https://github.com/flori/json/pull/565). `StimulusReflex` relies on `OpenStruct` in both the [Dataset](https://github.com/stimulusreflex/stimulus_reflex/blob/3a5111e085b21d1e628813e2cee72e978acb257a/lib/stimulus_reflex/dataset.rb#L5) and the [Element](https://github.com/stimulusreflex/stimulus_reflex/blob/3a5111e085b21d1e628813e2cee72e978acb257a/lib/stimulus_reflex/element.rb#L6) models. Attempting to run rails, ie a `rails console`, results in the error

```
/.rvm/gems/ruby-3.2.2/gems/stimulus_reflex-3.5.0.rc4/lib/stimulus_reflex/dataset.rb:5:in `<main>': uninitialized constant OpenStruct (NameError)
```

## To Reproduce

using `rubocop v1.63.0` or above, attempt to run rails console or start a rails server

## Expected behavior

the rails action can be performed without an error that causes it to exit

it seems a quick fix would be to explicitly requiring `ostruct`, like [this suggestion](https://github.com/flori/json/issues/579)

a longer term fix would be to stop using `OpenStruct` altogether as its [use is discouraged](https://github.com/ruby/ostruct/blob/master/lib/ostruct.rb#L136)

### StimulusReflex

- Gem: `3.5.0.rc4`
- Node package: `yarn 1.22.19`

### External tools

- Ruby: `3.2.2p53`
- Rails: `7.0.8.1`
- Node: `18.19.0`
- Rubocop: `1.63.0`
- Json: `2.7.2`

marcoroth:
Thanks for reporting @TanookiMario, I put up #697 to address this!

PR #735, opened 2025-04-02
Declare dependency on ostruct in gemspec

Zajn:
Warning fix

## Description
The absence of this declaration in `stimulus_reflex.gemspec` was emitting this warning in projects using this library:

> warning: .rubies/ruby-3.4.2/lib/ruby/3.4.0/ostruct.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0. You can add ostruct to your Gemfile or gemspec to silence this warning. Also please contact the author of stimulus_reflex-3.5.3 to request adding ostruct into its gemspec.

Related to #695, #698 

## Why should this be added
I know that the long-term goal is to move away from using ostruct entirely, but since it's still used in the meantime, this feels like a reasonable thing to do.

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,6 +9,7 @@ PATH
       cable_ready (~> 5.0)
       nokogiri (~> 1.0)
       nokogiri-html5-inference (~> 0.3)
+      ostruct (~> 0.6)
       rack (>= 2, < 4)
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Thank you @Zajn! 🙌🏼

